PR #12, opened 2018-11-15
Support bean qualifier

cescoffier:


cescoffier:
Wow! Didn't know about Weld Unit! Something to consider, I will open an issue right now.

PR #32, opened 2018-12-18
CloudEvents Serializer

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

Adding a CloudEventSerializer, for posting "structured mode" CloudEvents.

Possible demo code:
```java
  @Incoming("source")
  @Outgoing("to-http")
  public HttpMessage<CloudEventMessage>  process(CloudEventMessage<String> message) {
    return HttpMessage.HttpMessageBuilder.<CloudEventMessage>create()
      .withMethod("PUT")
      .withPayload(message)
      .withHeader("Content-Type", "application/cloudevents+json")
      .build();
  }
```

and the server would receive it like:

```
Host: localhost:8080
Content-Length: 236
Content-Type: application/cloudevents+json
User-Agent: Vert.x-WebClient/3.6.0

{"schemaURL":null,"specVersion":"0.2","time":"2018-12-18T14:51:03.791784+01:00","data":"11","extensions":[],"contentType":"text/plain","id":"244741bd-a8c4-4af1-89b4-87809581541c","type":"counter","payload":"11","source":"local://timer"}

```

### DIFF
--- a/smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/converters/CloudEventSerializer.java
+++ b/smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/converters/CloudEventSerializer.java
@@ -1,0 +1,22 @@
+package io.smallrye.reactive.messaging.http.converters;
+
+import io.cloudevents.json.Json;
+import io.smallrye.reactive.messaging.cloudevents.CloudEventMessage;
+import io.vertx.reactivex.core.buffer.Buffer;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public class CloudEventSerializer extends Serializer<CloudEventMessage> {
+
+ @Override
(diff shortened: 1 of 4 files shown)


smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/converters/CloudEventSerializer.java
```diff
@@ -0,0 +1,22 @@
...
+import io.vertx.reactivex.core.buffer.Buffer;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public class CloudEventSerializer extends Serializer<CloudEventMessage> {
+
+ @Override
+  public CompletionStage<Buffer> convert(CloudEventMessage payload) {
+    return CompletableFuture.completedFuture(Buffer.buffer(Json.encode(payload)));
```
matzew: this uses CloudEvents Json class, which adds `Jdk8Module` and some (De)Serializers for `ZonedDateTime` - all with Jackson.

PR #48, opened 2019-01-25
Introduce StreamProducer and simplify ReactiveMessagingExtension

mkouba:
- also mark Stream.value() as Nonbinding


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/StreamProducer.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @SuppressWarnings("rawtypes")
+    private Publisher<? extends Message> getPublisher(InjectionPoint injectionPoint) {
+        String name = getStreamName(injectionPoint);
+        List<Publisher<? extends Message>> list = streamRegistry.getPublishers(name);
+        if (list.isEmpty()) {
+            throw new IllegalStateException("Unable to find a stream with the name " + name + ", available streams are: " + streamRegistry.getPublisherNames());
+        }
+        // TODO Manage merge.
```
mkouba: I'm not quite sure what this `TODO` means. It's just copied from `PublisherInjectionManager` ;-)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java
```diff
@@ -79,6 +88,18 @@ void afterDeploymentValidation(@Observes AfterDeploymentValidation done, BeanMan
         CompletableFuture<Void> future = mediatorManager.initializeAndRun();
         try {
             future.get();
+
+            // NOTE: We do not validate @Stream annotations added by portable extensions
+            Set<String> names = registry.getPublisherNames();
+            for (InjectionPoint ip : streamInjectionPoints) {
+                String name = StreamProducer.getStreamName(ip);
+                if (!names.contains(name)) {
+                    done.addDeploymentProblem(new DeploymentException("No stream found for name: " + name + ", injection point: " + ip));
```
mkouba: We should probably add a test for this...

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/StreamProducer.java
```diff
@@ -0,0 +1,95 @@
...
+import io.smallrye.reactive.messaging.annotations.Stream;
+
+@ApplicationScoped
+public class StreamProducer {
+
+    @Inject
+    StreamRegistry streamRegistry;
+
+    @Produces
+    @Stream("") // Stream name is ignored during type-safe resolution
```
cescoffier: I would say it should fails. Stream name cannot be empty.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/StreamProducer.java
cescoffier: Yes, that's for me. Keep it for now. 

Merge means that if you have several stream with the same name, we may want to introduce a merge strategy:
* one (current implementation)
* merge (that might be a better default) would create a stream with the items for every matching streams. The items are emitted as they arrive
* concat - like merge but streams are concatenated, it only works for bounded streams.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java
cescoffier:
````
@return the stream name, mandatory, non null and non-blank. It must matches one of the available streams.
```

It should be enforced. (working on it)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/StreamProducer.java
cescoffier:
```
@return the stream name, mandatory, non null and non-blank. It must matches one of the available streams.
```

It should be enforced. (working on it)

cescoffier: ah got it, it's in the implementation.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java
mkouba: Yes. What I'm saying is that we throw `DeploymentException` if there is a `@Stream` with unknown stream name but we don't have a test for this right now. I can add the test today.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/StreamProducer.java
mkouba: Ok.

mkouba:
> However, thanks to the reactive type converters it is possible to transform the internal stream into a (Spring) Flux for instance ...

@cescoffier Hm, this wasn't supported by the previous version either, or? We would have to add another producer for Spring Flux etc.

cescoffier:
Actually, I'm not even sure that the previous approach I used would have worked with this without creating many many beans....

mkouba:
Test added we should be ready to merge?

cescoffier:
Thanks!

Issue #71, opened 2019-03-19
Invalid CloudEvent payload


matzew:
Using 0.0.5-SNAPSHOT... and when creating - and sending a Cloudevent, like:

```
    final CloudEventMessage<String> kafkaMessageCloudEventMessage = new CloudEventMessageBuilder()
    .id(UUID.randomUUID().toString())
    .source(URI.create(message.getTopic()))
    .type("kafka-event")
    .data(message.getPayload())
    .build();

    return HttpMessage.HttpMessageBuilder.<CloudEventMessage<String>>create()
      .withMethod("POST")
      .withPayload(kafkaMessageCloudEventMessage)
      .withHeader("Content-Type", "application/cloudevents+json")
      .build();

```

the entire object to the wire/http server is:

```
[...]

PR #72, opened 2019-03-19
Adding JsonIgnore for payload property

matzew:
@cescoffier not sure if that's the best... but the solution works, because the CloudEvent should not write out the `payload` field (see #71)

Also included a delegate to the `toString()`

### DIFF
--- a/smallrye-reactive-messaging-cloud-events/src/main/java/io/smallrye/reactive/messaging/cloudevents/DefaultCloudEventMessage.java
+++ b/smallrye-reactive-messaging-cloud-events/src/main/java/io/smallrye/reactive/messaging/cloudevents/DefaultCloudEventMessage.java
@@ -1,5 +1,6 @@
 package io.smallrye.reactive.messaging.cloudevents;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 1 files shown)


smallrye-reactive-messaging-cloud-events/src/main/java/io/smallrye/reactive/messaging/cloudevents/DefaultCloudEventMessage.java
```diff
@@ -61,7 +62,13 @@ public String getId() {
...
 
   @Override
+  @JsonIgnore
   public T getPayload() {
     return delegate.getData().orElseThrow(() -> new IllegalArgumentException("Invalid message - no payload"));
   }
+
+  @Override
+  public String toString() {
+    return delegate.toString();
```
cescoffier: Any reason for this?

matzew: hrm... I personally like "proper" toString :-)   I can remove it if you dont want it

cescoffier:
Thanks!

Issue #114, opened 2019-05-20
Add support for generics on MqttMessage:payload (sending side)


misl:
When using dynamic topics the only supported payload type is `byte[]`. This enforces the user to do the type conversion. This even though MqttSink internally already supports a number of conversions automatically.

It would be really nice if this automatic conversion would also be available when using MqttMessage directly. So instead of forcing MqttMessage<byte[]> it would be nice to use MqttMessage<T>.

misl:
PR is nearly ready.

PR #115, opened 2019-05-20
#114 Added support for generics, MqttMessage<T>

misl:
- Added support for generics: MqttMessage --> MqttMessage<T>
- Added unittest to check MqttMessage<Integer>
- Updated mqtt-quickstart

Fixes: #114


misl:
In my opinion the documentation requires no changes for this PR. 

The documentation could however be extended with sample methods just like present for Kafka messaging. But this is, in my opinion, a different issue.

PR #117, opened 2019-05-20
Update to new artifactId and version scheme

kenfinnigan:


misl:
Sorry for asking, but what issue is this PR solving?

kenfinnigan:
This is inline with other PRs to all SmallRye repos I've been working on over the last few days.

In essence, it means we can innovate with the implementation without breaking alignment with a particular MP specification version. Full details of the discussion is here: https://groups.google.com/forum/#!topic/smallrye/xs5HFiTq6gk

misl:
Thanks for explaining it, I will read the discussion.

rsvoboda:
I see one concern from end-user perspective.

Let's assume smallrye-reactive-messaging version is defined in BOM file - quarkus, thorntail, eap one
User application has dependency specified like this:
```
<dependency>
   <groupId>io.smallrye.reactive</groupId>
   <artifactId>smallrye-reactive-messaging</artifactId>
</dependency>
```

So when updating the app user just changes BOM version, no other changes.

With this proposal user will have to change not just BOM version but also several artifactIds
```
<artifactId>foo-1.0</artifactId> ==> <artifactId>foo-1.1</artifactId>
<artifactId>bar-1.2</artifactId> ==> <artifactId>bar-2.0</artifactId>
<artifactId>baz-1.1</artifactId> ==> <artifactId>baz-1.2</artifactId>
```
This might work for projects and new products, but it can be trouble for products with various backward compatibility requirements - @msve @maeste

I assume similar changes will be done to all SmallRye projects.

kenfinnigan:
These changes have been done to all SmallRye projects.

SmallRye pieces are going to be on a different support lifecycle compared to traditional EAP, and thus a different backward compatibility requirement.

Sure, it's not ideal to have users update the `artifactId` when updating a BOM, but we need to keep in mind that the `artifactId` is only updated when it's aligning with a different MP spec version, which isn't for every release.

Also, as per the discussion on Google Groups, without this change there's absolutely no way to support innovation in the implementations without breaking semantic versioning.

rsvoboda:
The shouldn't be much pain if users depend just on MP API artifacts.

It will be a bit clumsy if users depend on SmallRye artifacts directly (present in main EAP BOM too)

kenfinnigan:
I don't believe the SmallRye artifacts will be in the main EAP BOM

rsvoboda:
They are, I checked

kenfinnigan:
Those are internal use only artifacts. Once MP is properly added to EAP, it will be a completely separate piece

rsvoboda:
https://maven.repository.redhat.com/ga/org/jboss/bom/eap-runtime-artifacts/7.2.0.GA/eap-runtime-artifacts-7.2.0.GA.pom

rsvoboda:
MP stuff in EAP 7.2 is delivered in technical preview mode, so I thing there is a way to deffer my concerns :)

Ladicek:
I don't think this is the best place for this discussion, but if it's already ongoing, I'll add: as long as SmallRye is mainly used by integrators (Quarkus, Thorntail, WildFly, etc.), I think we're good, as it limits the number of people who need to track and apply these `artifactId` changes. When we start adding more public APIs, SmallRye will be depended on by end users, and that will be more painful and will require more care.

cescoffier:
@Ladicek that's exactly why it's still not merged. This implementation of reactive messaging is not only implementing a spec but provide extension, so end users may depend on it directly.

kenfinnigan:
I thought there was agreement on the way forward to handle innovation, etc in naming/versioning in SmallRye?

If that's not the case @cescoffier, what alternative do you propose?

cescoffier:
There is an agreement once the spec is finalized, but right now it’s a bit difficult to explain it’s not released. How would you refer to RCs? 

Hopefully the spec is going to be stable enough and this part of the version won’t change every month.

kenfinnigan:
Ah ok, I understand the concern now.

In that case, would it not make sense to make it "-1.0" as the intention is to have the implementation align with 1.0 of the specification?

Whether that means it's 1.0.RC1 now, and 1.0.RC2 in a few weeks, and then 1.0.Final, it's all targeting 1.0 of the spec

kenfinnigan:
Not sure how I closed the PR without it being merged, oops

cescoffier:
I was wondering. It’s going to be
Merged before the next release.

Issue #118, opened 2019-05-20
Why does the root pom has a jackson-databind dependency?


misl:
The project's root pom include the jackson-databind dependency, even though it does not seem to be used anywhere. Master branch builds just fine without it.
```
   <dependency>
     <groupId>com.fasterxml.jackson.core</groupId>
     <artifactId>jackson-databind</artifactId>
     <version>${jackson.version}</version>
   </dependency>
```
In general, why are there so many dependencies in the root pom? Should these not be  in dependency management instead? Then every module should include its own dependencies.

PR #119, opened 2019-05-20
Removed jackson-databind dependency

misl:
- Removed dependency on jackson-databind
- Removed jackson maven property

Fixes: #118


misl:
Hmm, odd that the build failed. Even stranger that it was tried 2 times and failed in different parts.

cescoffier:
This is a Semaphore CI issue, the launched JVM crashed sometimes... Very annoying.

PR #130, opened 2019-06-12
Update to use new SmallRye Config artifacts

kenfinnigan:


kenfinnigan:
forked jvm crash, argh!

I swear it passed locally

PR #131, opened 2019-06-14
Putting topic names in quotes consistently

gunnarmorling:
It's given in quotes at another places, so let's do it consistently everywhere.


gunnarmorling:
@cescoffier, there's another minor glitch, messages are logged with an un-implemented toString() method:

    producer_1    | 2019-06-14 14:56:29,740 INFO  [io.sma.rea.mes.kaf.KafkaSink] (RxComputationThreadPool-1) Sending message io.smallrye.reactive.messaging.kafka.SendingKafkaMessage@5c1f914a to Kafka topic 'temperature-values'

I thought about logging the key, but I'm wondering whether there's some security concern attached to that. WDYT?

cescoffier:
Use used `toString` to let the user sub-class the `Message` class print a better message. For security reasons we cannot really provide something more explicit, both the key and the value can contain sensitive data.

Would you prefer dropping the display of the message? I agree that it's not really useful like this.

PR #138, opened 2019-06-25
Spotted another typo in outgoing config

misl:
One of the key in outgoing was using incoming instead.


misl:
Grrr, interference with other change. I should have made a proper branch. Withdrawing this PR

PR #147, opened 2019-06-26
Fix #146: Added amqp-quickstart

### LINKED ISSUE #146: Add AMQP quickstart
Add a quickstart for AMQP similar to the quickstarts for Kafka and MQTT.

misl:
Fixes #146 

New amqp-quickstart examples is very similar to the existing Kafka and MQTT examples.


cescoffier:
Thanks!

PR #148, opened 2019-06-27
Fixes another copy-paste error in documentation

misl:


cescoffier:
Yet another great catch, eagle eye!

PR #149, opened 2019-06-30
Use 'amqp-host' as stated in documentation and not 'amqp-server'

lulf:
The documentation states that amqp-host can be used to set the hostname of the AMQP server, but the code actually uses amqp-server. This change makes the actual setting be amqp-host.


cescoffier:
Thanks!

PR #175, opened 2019-08-19
Add an @OnOverflow annotation to configure the back pressure strategy

cescoffier:


cescoffier:
@mswiderski Here is a proposal. Can you have a look?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
```diff
@@ -7,29 +7,54 @@
...
+                            BackpressureOverflowStrategy.ERROR);
+        } else {
+            switch (onOverflow.value()) {
+                case BUFFER:
+                    Flowable<Message<? extends T>> p = Flowable.create(deferred, BackpressureStrategy.BUFFER);
+                    if (onOverflow.bufferSize() > 0) {
+                        publisher = p.onBackpressureBuffer(onOverflow.bufferSize(),
+                                () -> LOGGER.error("Buffer full for emitter {}", name), BackpressureOverflowStrategy.ERROR);
+                    } else {
+                        publisher = p;
```
mswiderski: just a question here, is the buffer size set here to default value as described in the doc -127? It looks to me that the onBackpressureBuffer is not set at all if the bufferSize in annotation is not set which the defaults to 0

https://github.com/smallrye/smallrye-reactive-messaging/pull/175/files#diff-d47d3adc7e916519aa6946ebf2957af0R67

or am I missing something?

cescoffier: By default, if `OnOverflow` is not used, we use a buffer of 127 elements (current behavior). We may want to add a way to configure the `127`magic number.

If `OnOverflow`is used with the buffer strategy, it would:

* use a buffer of the size passed in the `bufferSize` attribute
* use an unbounded buffer is `bufferSize` is not set

So, we cannot use a _default_ value for these cases, as it depends on the stream.

cescoffier:
About adding a configuration property to configure the default buffer size:

I agree and I will add this. However, keep in mind that it would only be used when the `OnOverflow` annotation is not used. See the rationale in my previous comment.

cescoffier:
I've added the configuration support and at the same time replace the _broken_ way to inject `Config` objects in tests.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
mswiderski: I was referring to this line

```
if (onOverflow.bufferSize() > 0) {
```
where I assume `onOverflow` is annotation and the `bufferSize` is optional with default value 0. So if someone creates

```
@Inject
	@Stream("hello")
	@OnOverflow(value = OnOverflow.Strategy.BUFFER)
	Emitter<String> emitter;
```

then the back pressure is not set at all, isn't it? because it goes into the else statement and `p` is only 

```
Flowable<Message<? extends T>> p = Flowable.create(deferred, BackpressureStrategy.BUFFER);
```

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MediatorManager.java
```diff
@@ -42,6 +44,10 @@
 
     private final List<AbstractMediator> mediators = new ArrayList<>();
 
+    @Inject
+    @ConfigProperty(name = "smallrye.messaging.emitter.default-buffer-size", defaultValue = "127")
```
mswiderski: great to see this :) will it need extra property in quarkus as well? or can this be somehow reused/aliased?

cescoffier: No, you will be able to use it directly. So you can configure it from the `application.properties` file.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
cescoffier: In this case, the backpressure is handled using an unbounded buffer. I'm going to extend the documentation to clarify this case.

Issue #176, opened 2019-08-22
Allow SSL connections for AMQP


nderwin:
In `AmqpConnector` where `AmqpClientOptions` is created, it would be helpful if there was a way to configure whether SSL was enabled or not.

I ran into this trying to use Quarkus 0.21.1 (which uses version 0.0.10 of this project) to connect to an Amazon MQ instance, but it failed because the connection was not created using SSL.

cescoffier:
Yes, the SSL configuration is missing. Fancy a PR? 

The options that can be passed are the following:
https://vertx.io/docs/apidocs/io/vertx/amqp/AmqpClientOptions.html.

nderwin:
So, I created https://github.com/smallrye/smallrye-reactive-messaging/pull/179, though it just covers the SSL option.  If you want, I can add the other documented options.

cescoffier:
Fixed in #179.

PR #179, opened 2019-08-23
Fix for #176

nderwin:
* added new configuration item for using SSL for the AMQP connection
* added the config value to the AMQP client options


cescoffier:
Did you test it with your use case?

nderwin:
Effectively, yes.  Using [this](https://medium.com/@yazidaqel/quarkus-vertx-a-powerfull-combination-part-1-introduction-b039b911686) as an example of how to work around not being able to use the Quarkus extension, I've been able to inject the boolean value from the config and use it to set the value in the AMQP options object passed to the client, and been able to successfully connect to Amazon MQ.  I've also been able to tell it to not use SSL for connecting to a local Docker container running ActiveMQ.

PR #185, opened 2019-09-09
Add an MQTT server connector

francois-travais:
Based on https://github.com/Agaetis-IT/agaetis-reactive-messaging-mqtt-server

It only implements publish requests of any QoS.

### DIFF
--- a/doc/mqtt.adoc
+++ b/doc/mqtt.adoc
@@ -1,11 +1,20 @@
 == Interacting using MQTT
 
-You can connect to a MQTT broker or server as a source or sink. The MQTT support is based on the
-https://vertx.io/docs/vertx-mqtt/java/#_vert_x_mqtt_client[Vert.x MQTT Client].
+You have two options here:
 
-=== Dependency
+1. you can connect to an MQTT broker or server as a source or sink. The MQTT support is based on
+the https://vertx.io/docs/vertx-mqtt/java/#_vert_x_mqtt_client[Vert.x MQTT Client].
+2. you can receive MQTT messages as an MQTT server. The MQTT support is based on the
+https://vertx.io/docs/vertx-mqtt/java/#_vert_x_mqtt_server[Vert.x MQTT Server]
 
-To enable the MQTT support, you need the following dependency:
+=== MQTT client (incoming or outgoing)
+
+For if you want to produce or consumer MQTT messages coming or going to an MQTT broker
+or server.
+
+==== Dependency
+
+To enable the MQTT client support, you need the following dependency:
 
 [source,xml,subs=attributes+]
 ----
@@ -16,7 +25,7 @@ To enable the MQTT support, you need the following dependency:
 </dependency>
 ----
 
-=== Retrieving messages from MQTT
+==== Retrieving messages from an MQTT broker or server
 
 [source]
 ----
@@ -34,7 +43,7 @@ Message coming from MQTT are `Message<byte[]>`. The payload is a `byte[]`. You c
 
 NOTE: If the `topic` attribute is not set the channel name is used instead.
 
-=== Forwarding messages to MQTT
+==== Forwarding messages to an MQTT broker or server
 
 [source]
 ----
@@ -46,7 +55,7 @@ mp.messaging.outgoing.data.username=username
 mp.messaging.outgoing.data.password=secret
 ----
 
-The payload of message forwarded to MQTT must be:
+The payload of message forwarded to an MQTT broker or server must be:
 
 * a `io.vertx.core.json.JsonObject` or a `io.vertx.core.json.JsonArray`
 * a `String` or a primitive types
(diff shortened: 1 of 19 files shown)


smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
```diff
@@ -0,0 +1,159 @@
...
+        options.setHost(
+                config.getOptionalValue("host", String.class).orElse(DEFAULT_HOST));
+        return options;
+    }
+
+    MqttServerSource(Vertx vertx, Config config) {
+        this.broadcast = config.getOptionalValue("broadcast", Boolean.class).orElse(false);
+        final MqttServerOptions options = mqttServerOptions(config);
+        this.mqttServer = MqttServer.create(vertx, options);
+        final BehaviorProcessor<MqttMessage> processor = BehaviorProcessor.create();
```
cescoffier: Are you sure about the `BehaviorProcessor`. Is it what we need here? 
The `BehaviorProcessor` cached the last item emitted by the MQTT Server and replay it as well as all the subsequent item. So while there are no subscribers, the items are dropped. It may be what you want, just checking.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
```diff
@@ -0,0 +1,159 @@
...
+
+        mqttServer.exceptionHandler(error -> {
+            logger.error("Exception thrown", error);
+            processor.onError(error);
```
cescoffier: Do we want to retry? Might be something we can add later.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
```diff
@@ -0,0 +1,159 @@
...
+            // this implementation doesn't keep track of sessions
+            endpoint.accept(false);
+        });
+
+        this.source = ReactiveStreams.fromPublisher(processor
+                .delaySubscription(mqttServer.rxListen()
+                        .doOnSuccess(ignored -> logger
+                                .info("MQTT server listening on {}:{}", options.getHost(), mqttServer.actualPort()))
+                        .doOnError(throwable -> logger.error("Failed to start MQTT server", throwable))
+                        .toFlowable()
```
cescoffier: I believe the `)` is required here. 
This is the publisher used for the subscription delay.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
francois-travais: I was counting on the fact that for QoS 0 messages it's not guarantee that any messages arrives and for QoS 1 and 2, if the acknowledgment is well handled by the subscriber the publisher should try to send it again if he receives no response in return.
I guess I could have used a `MulticastProcessor` right? it would have buffered a bit. I'm not sure be it feels like the other processors fitting there. What do you think would be the most suitable? I could also add a configuration flag to switch to `MulticastProcessor`.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
francois-travais: In my case I was going to count on Kubernetes to restart it. But it makes sense, I'll look at it to see if I can quickly implement a retry mechanism.

cescoffier: No hurry, what you did it what it generally intended.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
cescoffier: Ok, So I agree with you, the BehaviorProcessor is the best candidate here.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
francois-travais: I'm not sure to understand what you mean: you want to apply the `compose()` on the result of the `delaySubscription()` call right?

cescoffier: My bad, I missed a `))` at the end of the `doOnError` line.

cescoffier:
All good! Just need some documentation and a quickstart (optional).

francois-travais:
@cescoffier I'm wandering if it makes sense to have both the current MQTT client connectors and the MQTT server connectors in the same maven module like it is in Vert.x project? If not does it make sense to share the `MqttMessage` definition among the two modules?

I've also another question: right now when `getPublisherBuilder(Config config)` is called for the first time it creates an MQTT server with this config but if  `getPublisherBuilder(Config config)` is called again with a different config it will return the previously instantiated MQTT server and ignore the new configuration. Is it a problem?

cescoffier:
> I'm wondering if it makes sense to have both the current MQTT client connectors and the MQTT server connectors in the same maven module like it is in Vert.x project? If not does it make sense to share the MqttMessage definition among the two modules?

That's a good question. We can merge them and add an "attribute" selecting the mode.

> I've also another question: right now when getPublisherBuilder(Config config) is called for the first time it creates an MQTT server with this config but if getPublisherBuilder(Config config) is called again with a different config it will return the previously instantiated MQTT server and ignore the new configuration. Is it a problem?

The method is called for every channel using your connector. so if you have 2 channels with different config it can be an issue. However, we can clearly state that only one server would be created, and the first one is reused.

francois-travais:
While writing the quickstart I trigger the following exception from the `MediatorConfiguration`: https://github.com/smallrye/smallrye-reactive-messaging/blob/5ae736d9cfea68f09731c7d7aa9a436e78da1fef/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/MediatorConfiguration.java#L497-L501

The culprit being:
```java
@Incoming("my-server")
public void source(MqttMessage message) {
  LOGGER.info("MQTT Message received {}", message);
}
```
The easy fix would be to change the method signature but I'm wandering why the code I wrote wouldn't work?

francois-travais:
I refactored the MQTT documentation to include both the client and server connectors documentations.

> That's a good question. We can merge them and add an "attribute" selecting the mode.
I think it would be better to make the difference in the connector name: `smallrye-mqtt-client` and `smallrye-mqtt-server`. I don't think I'll merge them just yet, I can add an issue for later if you want.

cescoffier:
> While writing the quickstart I trigger the following exception from the `MediatorConfiguration`:
> 
> https://github.com/smallrye/smallrye-reactive-messaging/blob/5ae736d9cfea68f09731c7d7aa9a436e78da1fef/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/MediatorConfiguration.java#L497-L501
> 
> The culprit being:
> 
> ```java
> @Incoming("my-server")
> public void source(MqttMessage message) {
>   LOGGER.info("MQTT Message received {}", message);
> }
> ```
> 
> The easy fix would be to change the method signature but I'm wandering why the code I wrote wouldn't work?

the signature is not listed in the set of signatures in the spec.  The reason is the following:
Your method receives a `Message` (actually an implementation of `Message`) and is synchronous (return a `void`). So it cannot handle acknowledgment, as acknowledgement is an async process.

I'm quite sure we could add support for it somehow.

For now, you can return a `CompletionStage<Void>` and return `message.ack()`

francois-travais:
It's all good to me, I've noting more to add/change

doc/mqtt.adoc
```diff
@@ -56,8 +65,55 @@ Other objects are mapped to JSON.
...
+
+==== Dependency
+
+To enable the MQTT server support, you need the following dependency:
+
+[source,xml,subs=attributes+]
+----
+<dependency>
+  <groupId>io.smallrye.reactive</groupId>
+  <artifactId>smallrye-reactive-messaging-mqtt-server-1.0</artifactId>
```
cescoffier: So, the artifact id has changed (again...)
You can drop the 1.0 suffix.

cescoffier:
Thanks! Great job!

Issue #231, opened 2019-10-01
Add configuration for connection / reconnection options


nderwin:
The configuration for reconnect attempts, reconnect interval, and connect timeout are commented as needing to be configurable, and with the upgrade to using Quarkus 0.23.1, I will need to change the default values of the reconnect attempts and interval.  I will submit a PR for this issue.

nderwin:
Looks like this was included in the 1.0.7 release.   :tada:

PR #232, opened 2019-10-01
Fix for #231

nderwin:
* added injectable configuration for the reconnect attempts, reconnect interval, and connection timeout


cescoffier:
Thanks!

Issue #284, opened 2019-10-30
Clean up Emitter interface


xstefank:
Points collected on the MP reactive hangout on which needs to be done before Emitter can be pushed up to spec: 

- [Emitter<T> send(T msg)](https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/Emitter.java#L34) should return `CompletionStage<Void>`

- change javadoc with regards to `@OnOverflow` - https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/Emitter.java#L31

- rename `smallrye.messaging.emitter.default-buffer-size` to `mp.messaging.emitter.defult-buffer-size`

Emily-Jiang:
Emtter L3, remove  "if the stream does not have any pending requests, "

cescoffier:
The new emitter has been merged

PR #297, opened 2019-11-06
#284 improve Emitter 

Emily-Jiang:
Signed-off-by: Emily Jiang <emijiang6@googlemail.com>

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
@@ -22,7 +22,7 @@ public class EmitterImpl<T> implements Emitter<T> {
 
     EmitterImpl(String name, String overFlowStrategy, long bufferSize, long defaultBufferSize) {
         FlowableOnSubscribe<Message<? extends T>> deferred = fe -> {
-            if (!internal.compareAndSet(null, fe)) {
+            if (!internal.compareAndSet(null, fe.serialize())) {
                 fe.onError(new Exception("Emitter already created"));
             }
         };
@@ -88,7 +88,7 @@ public class EmitterImpl<T> implements Emitter<T> {
     }
 
     @Override
-    public synchronized void send(Message<T> msg) {
+    public synchronized <M extends Message<? extends T>> void send(M msg) {
         if (msg == null) {
             throw new IllegalArgumentException("`null` is not a valid value");
         }
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
@@ -119,7 +119,7 @@ public class EmitterImpl<T> implements Emitter<T> {
             throw new IllegalArgumentException("`null` is not a valid exception");
         }
         verify().onError(e);
-        
+
     }
 
     @Override
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
@@ -1,5 +1,7 @@
 package io.smallrye.reactive.messaging.extension;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.microprofile.reactive.messaging.Message;
@@ -25,9 +27,8 @@ public class EmitterImpl<T> implements Emitter<T> {
             }
         };
         if (overFlowStrategy == null) {
-            publisher = Flowable.create(deferred, BackpressureStrategy.BUFFER)
-                    .onBackpressureBuffer(defaultBufferSize, () -> LOGGER.error("Buffer full for emitter {}", name),
-                            BackpressureOverflowStrategy.ERROR);
+            publisher = Flowable.create(deferred, BackpressureStrategy.BUFFER).onBackpressureBuffer(defaultBufferSize,
+                    () -> LOGGER.error("Buffer full for emitter {}", name), BackpressureOverflowStrategy.ERROR);
         } else {
             OnOverflow.Strategy strategy = OnOverflow.Strategy.valueOf(overFlowStrategy);
             switch (strategy) {
@@ -37,8 +38,13 @@ public class EmitterImpl<T> implements Emitter<T> {
                         publisher = p.onBackpressureBuffer(bufferSize,
                                 () -> LOGGER.error("Buffer full for emitter {}", name), BackpressureOverflowStrategy.ERROR);
                     } else {
-                        publisher = p;
+                        publisher = p.onBackpressureBuffer(defaultBufferSize,
+                                () -> LOGGER.error("Buffer full for emitter {}", name), BackpressureOverflowStrategy.ERROR);
+
                     }
+                    break;
+                case UNBOUNDED_BUFFER:
+                    publisher = Flowable.create(deferred, BackpressureStrategy.BUFFER);
                     break;
                 case DROP:
                     publisher = Flowable.create(deferred, BackpressureStrategy.DROP);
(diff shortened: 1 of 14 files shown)


Emily-Jiang:
@cescoffier can you take a look to see why the tests are failing? What is the completion criteria for the CompletionStage to complete?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
```diff
@@ -67,18 +73,18 @@ boolean isConnected() {
...
-    public synchronized Emitter<T> send(T msg) {
+    public synchronized CompletionStage<Void> send(T msg) {
         if (msg == null) {
             throw new IllegalArgumentException("`null` is not a valid value");
         }
         FlowableEmitter<Message<? extends T>> emitter = verify();
         if (msg instanceof Message) {
             //noinspection unchecked
-            emitter.onNext((Message) msg);
+            return CompletableFuture.runAsync(() -> emitter.onNext((Message) msg));
```
cescoffier: That's not correct.

You need something like (be aware this is pseudo-code):

```
CompletableFuture future = new CompletableFuture();
if (msg instanceof Message) {
    // here things need to be discussed
    // if the user passes a Message, he can attach the acknowledgment logic
   // however we cannot really do anything. We would need a way to customize the Supplier
} else {
   emitter.onNext(Message.of(msg, () -> future.complete(null));
}
```

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
```diff
@@ -268,13 +274,15 @@ public boolean isCaught() {
         }
 
         public void run() {
-            emitter.send("a").send("b");
+            emitter.send("a");
```
cescoffier: The tests need to check that the returned CompletionStage is completed when the message is processed.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/Emitter.java
```diff
@@ -11,27 +13,28 @@
...
 public interface Emitter<T> {
 
     /**
      * Sends a payload or a message to the stream.
      *
      * @param msg the <em>thing</em> to send, must not be {@code null}
-     * @return the current emitter
-     * @throws IllegalStateException if the stream does not have any pending requests, or if the stream has been
-     *         cancelled or terminated.
+     * @return the {@code CompletionStage}
```
cescoffier: it needs to explain when the CS is completed.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
```diff
@@ -35,10 +35,17 @@
...
-         * buffer. If the buffer is full, an error is propagated.
+         * This creates a buffer with the size specified by {@link #bufferSize()} if present.
+         * Otherwise, the size will be the value of the config property
+         * <strong>mp.messaging.emitter.defult-buffer-size</strong>.
+         * If the buffer is full, an error is propagated.
          */
         BUFFER,
+        /**
+         * Buffers <strong>all</strong> values until the downstream consumes it.
+         * This creates an unbound buffer. If the buffer is full, an error is propagated.
```
cescoffier: _If the buffer is full_... in this case if the buffer is full, you are running out of memory :-)

cescoffier:
There is a design issue when the user passes a `Message`, as we cannot return a `CompletionStage` completed when the message is acknowledged.

One way would be to have 2 methods:

* `void send(Message<? extends T> message);` // can also return `this`
* `CompletionStage<Void> send(T payload);` 

We need to check how practical is it.

cescoffier:
For the record, the attached patch should fix the test failures and fix the generic signature.
[emitter.patch.txt](https://github.com/smallrye/smallrye-reactive-messaging/files/3868391/emitter.patch.txt)

cescoffier:
Remaining tasks:

- [ ] Test with a custom message implementation
- [ ] Test that the ack is correctly propagated to the returned CompletionStage.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/Emitter.java
Emily-Jiang: done

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
Emily-Jiang: done

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
Emily-Jiang: updated

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
Emily-Jiang: added more tests

Emily-Jiang:
@cescoffier please check whether you have more comments.

cescoffier:
The Kafka tests are failing because they use an emitter.
Apply the following diff:

```
         @Inject
         @Channel("temperature-values")
-        private Emitter<KafkaMessage<String, String>> emitter;
+        private Emitter<String> emitter;
```

Also, it would be great to rebase and cleanup the history a bit.

cescoffier:
We need more tests:
1. Test emitter with messages
2. Test that the acknowledgment is propagated correctly. This set of tests is very important because it's the core of this PR: the completion stage returned by the emitter but be completed when the message is acknowledged.

cescoffier:
Also, this PR is a breaking change on the API, so, after discussing with @kenfinnigan we would need to:

* keep the previous behavior for the Emitter class contained in the provider module (the old one, already deprecated)
* use the new API for the Emitter class from the `api` package.

Emily-Jiang:
I have added the tests to check the acknowledgment was propagated correctly though.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
```diff
@@ -137,21 +167,24 @@ public void testWithTwoEmittersConnectedToOneProcessor() {
...
+        public void consume(final String s) {
             list.add(s);
         }
     }
 
     @ApplicationScoped
     public static class MyBeanEmittingPayloadsUsingStream {
         @Inject
-        @Stream("foo")
+        @Channel("foo")
```
cescoffier: This test is about testing the legacy Stream annotation, do don't update to @Channel.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
```diff
@@ -220,16 +380,16 @@ public void run() {
...
-        public void consume(String s) {
+        public void consume(final String s) {
             list.add(s);
         }
     }
 
     public static class BeanWithMissingStream {
         @Inject
-        @Stream("missing")
+        @Channel("missing")
```
cescoffier: This test is testing the behavior of the legacy @Stream annotation, so don't replace it.

Emily-Jiang: done

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
Emily-Jiang: done

Emily-Jiang:
@cescoffier I tried to add tests to test the deprecated Emitter. However, all plugging was removed by the package refractoring and the new ground needs to be laid out. I think if you want to rewire the old api, all of the scaffolding needs to be done again. This should be done under a different issue. By the way, I've addressed all of your comments except this one.

cescoffier:
Starting working on it to include the PR _soonish_.

cescoffier:
This PR is superseded by #339

Emily-Jiang:
Close this one because #339 superseded this.

PR #303, opened 2019-11-13
Rework Kafka Message API

### LINKED ISSUE #268: How to get the offset of a KafkaMessage
Hi,

I'd like to know how to get the **offset** of a **KafkaMessage**? The offset information is there (as we can see in the picture)... Why is it not public as partition and topic name?

Screen Shot 2019-10-19 at 17 32 22

Best regards,

cescoffier:
* avoid using objects to primitive type
* allow accessing offset on received message (Fix #268)
* add Kafka message checks in the tests

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/SendingKafkaMessage.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/SendingKafkaMessage.java
@@ -9,12 +9,12 @@ public class SendingKafkaMessage<K, T> implements KafkaMessage<K, T> {
     private final String topic;
     private final K key;
     private final T value;
-    private final Integer partition;
-    private final Long timestamp;
+    private final int partition;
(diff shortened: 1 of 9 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/SendingKafkaMessage.java
```diff
@@ -65,10 +65,15 @@ public MessageHeaders getHeaders() {
...
+    public int getPartition() {
+        if (partition < 0) {
+            return -1;
         }
         return partition;
     }
+
+    @Override
+    public long getOffset() {
+        return -1;
```
kenfinnigan: Is the offset meaningless when sending, that's why we return -1?

cescoffier: yes, the offset is the position of the record in the partition. It only makes sense when receiving a record.

PR #305, opened 2019-11-13
Update the Mosquitto version used in test to 1.6.7

cescoffier:


kenfinnigan:
What's the "Build" job? Where did it come from?

cescoffier:
Trying these github actions thingy...

kenfinnigan:
Let me know how it goes, considering it for all SRye.

Depends on whether it can handle release type jobs, emails, etc

cescoffier:
Right now, it does not work well. Some failure when we try to connect to the port 32769 (and always this one...)

PR #312, opened 2019-11-15
Update parent to version version 13

cescoffier:
Supersedes #260

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,10 +231,17 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
+          <version>3.1.0</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -231,10 +231,17 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
+          <version>3.1.0</version>
```
kenfinnigan: Do these versions not come from the parent?

If not, we should fix that

cescoffier: I pinned it because I want to have the control of it. It's mostly to support new version of java with the annoying linter.

kenfinnigan: so you don't think we should be managing these versions in a parent, for any SmallRye project?

cescoffier: Both :-)

The parent should give a recent enough version.
Projects can decide to override to tweak their build.

kenfinnigan: fair enough

PR #315, opened 2019-11-15
Provide an in-memory connector

cescoffier:
Provides an _in-memory_ connector mainly designed for testing applications without having to use an external broker.

The idea is to substitute the `connector` of a specific channel to use `smallrye-in-memory`.
Then your test can send messages and checked the received messages:

```
InMemoryConnector bean = container.getBeanManager().createInstance()
                .select(InMemoryConnector.class, ConnectorLiteral.of(InMemoryConnector.CONNECTOR)).get();

        InMemorySink<String> bar = bean.sink("bar");
        InMemorySource<String> foo = bean.source("foo");
        foo.send("hello");
        assertThat(bar.received())
              .hasSize(1).extracting(Message::getPayload).containsExactly("HELLO");
```


cescoffier:
The connector qualifier is not yet able to target non-class. This has been fixed in the upcoming Reactive Messaging spec.

doc/testing.adoc
```diff
@@ -0,0 +1,60 @@
+== Testing without external broker
+
+It's not rare to have to test your application but deploying the infrastructure can be cumbersome.
+While Docker or Test Containers have improved the the testing experience, you may want to mock this infrastructure.
+
+SmallRye Reactive Messaging proposes an _in-memory_ connector for this exact purpose.
+It allows switching the connector used for a channel with an _in-memory_ connector.
+This connector can then provide a way to send messages for incoming channels, or checked the received messages for outgoing
```
kenfinnigan: s/checked/check

doc/testing.adoc
```diff
@@ -0,0 +1,60 @@
...
+import io.smallrye.reactive.messaging.connector.InMemorySink;
+// ...
+
+@Test
+class MyTest {
+
+    // 1. Switch the channels to the in-memory connector:
+    @BeforeAll
+    public static void switchMyChannels() {
+        InMemoryConnector.switchIncomingChannelToInMemory("prices");
```
kenfinnigan: Does it make sense for this to be a more generic `switchChannelToInMemory()`? Or it needs to specifically refer to incoming and outgoing?

For instance, does it make sense to set it for both?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connector/InMemoryConnector.java
```diff
@@ -0,0 +1,266 @@
...
+     * </pre></code>
+     * <p>
+     * With the {@link InMemorySink}, you can checked the messages received by the channel, to verify that the expected
+     * messages have been received.
+     *
+     * @param channels the channels to switch, must not be {@code null}, must not contain {@code null}, must not contain
+     *        a blank value
+     */
+    public static void switchOutgoingChannelToInMemory(String... channels) {
+        for (String channel : channels) {
```
kenfinnigan: Is this missing the same null check as the incoming?

doc/testing.adoc
cescoffier: unfortunately no as it would configure for both incoming and outgoing which may lead to conflicts with the rest of the application.

kenfinnigan: Would there be conflicts? If you're mocking a set of channels, you're not likely using real channels for the others are you?

So if "prices" is mocked for incoming, but not outgoing, that would prevent a method being used to create the values for the incoming channel, unless you explicitly mocked the outgoing as well?

Wondering if we can just presume everything is being mocked, and therefore incoming or outgoing is irrelevant, we're just switching a channel name to "in memory"?

cescoffier: Actually, I need to try, it may work by accident.

cescoffier: It does work :-) Forgot one small details about the spec: channel names are unique.
Updated.

cescoffier:
About the location of the in-memory connector, I'm not sure. I would prefer keeping it in the main provider to avoid adding another dependency,

Ladicek:
Stupid question (because I don't know anything about MP Reactive Messaging): could the in-memory connector be used outside of tests, to provide an in-app event bus? (Similarly to Vert.x event bus, or CDI async events.)

cescoffier:
It should not be used as an event bus because it’s not an event bus, but it can be used as a _connector_ connecting sources to sinks. So it has use cases outside of tests.

PR #316, opened 2019-11-16
Extract API classes into their own artifact

kenfinnigan:
- Makes it easier to identify possible spec proposals in the future
 - Deprecated Emitter in annotations package, and created copy in main package as replacement


pom.xml
```diff
@@ -56,21 +56,26 @@
...
-    <cloudevent.version>0.2.1</cloudevent.version>
-    <axle.version>0.0.10</axle.version>
 
     <asciidoctorj.version>1.5.8.1</asciidoctorj.version>
 
     <sonar.coverage.jacoco.xmlReportPaths>target/site/jacoco/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
   </properties>
 
   <modules>
+    <module>reactive-messaging-api</module>
```
cescoffier: should it be named smallrye-reactive-messaging-api?

smallrye-reactive-messaging-provider/pom.xml
```diff
@@ -13,13 +13,18 @@
...
-      <artifactId>jandex</artifactId>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-api</artifactId>
     </dependency>
 
     <dependency>
-      <groupId>javax.annotation</groupId>
-      <artifactId>javax.annotation-api</artifactId>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
```
cescoffier: Any reason for this dependency?

kenfinnigan:
probably, let me check docs

pom.xml
kenfinnigan: the artifact is named that, I just cheated on the directory name to have it at the top.

Can add "smallrye" for consistency if desired

smallrye-reactive-messaging-provider/pom.xml
kenfinnigan: `MediatorConfigurationSupport` uses `ClassUtils` from it.

cescoffier: We should consider removing it at some point. Depending on the lib for a single class, does not make a lot of sense.

kenfinnigan:
Totally agree. I can look at that this week

cescoffier:
Can you resolve the conflict?

PR #320, opened 2019-11-18
Initial version of the JMS connector

cescoffier:
This PR implements a connector for JMS.
It requires having a bean exposing the `ConnectionFactory` (platform-specific).

### DIFF
--- a/doc/jms.adoc
+++ b/doc/jms.adoc
@@ -1,0 +1,315 @@
+== Interacting using JMS (Java Message Service)
+
+You can use SmallRye Reactive Messaging to receive and send JMS messages.
+
+=== Dependency
+
+To enable the Camel support, you need the following dependency:
+
+[source,xml,subs=attributes+]
+----
+<dependency>
+  <groupId>io.smallrye.reactive</groupId>
+  <artifactId>smallrye-reactive-messaging-jms</artifactId>
+  <version>{version}</version>
+</dependency>
+----
+
+The connector is named `smallrye-jms`.
+
+=== Connection Factory Requirement
+
+To interact with JMS, you need to _expose_ a `javax.jms.ConnectionFactory` as CDI bean.
+For example, you need a bean producing the `ConnectionFactory`:
+
+[source, java]
+----
+@ApplicationScoped
+public class ConnectionFactoryBean {
+
+    @Inject @ConfigProperty(name="jms-username")
+    String username;
+
+    @Inject @ConfigProperty(name="jms-password")
+    String password;
+
+    @Produces
+    ConnectionFactory factory() {
+        ActiveMQJMSConnectionFactory factory = new ActiveMQJMSConnectionFactory(
+                "tcp://localhost:61616",
+                username, password);
+        return factory;
+    }
+
+}
+----
+
+The JMS connector is consuming the produced _factory_ to connect to the broker.
+
+==== Selecting ConnectionFactory
+
+In case you have several connection factories, you can use the `@Named` qualified on your factory to specify the name.
+Then, in the channel configuration, configure the name as follows:
+
+[source,properties]
+----
+# Configure the connector globally
+mp.messaging.connector.smallrye-jms.connection-factory-name=my-factory-name
+# Configure a specific incoming channel
+mp.messaging.incoming.my-channel.connection-factory-name=my-factory-name
+# Configure a specific outgoing channel
+mp.messaging.outgoing.my-channel.connection-factory-name=my-factory-name
+----
+
+==== Configure credentials
+
+To configure the _username_ and _password_ to connect to the broker and create a session, use the following properties:
+
+[source,properties]
+----
+# Configure the connector globally
+mp.messaging.connector.smallrye-jms.username=my-username
+mp.messaging.connector.smallrye-jms.password=secret
+# Configure a specific incoming channel
+mp.messaging.incoming.my-channel.username=my-username
+mp.messaging.incoming.my-channel.password=secret
+# Configure a specific outgoing channel
+mp.messaging.outgoing.my-channel.username=my-username
+mp.messaging.outgoing.my-channel.password=secret
+----
+
+==== Session mode
+
+By default, the connector uses the `AUTO_ACKNOWLEDGE` session mode.
+You can configure it using:
+
+[source,properties]
+----
+# Configure the connector globally
+mp.messaging.connector.smallrye-jms.session-mode=AUTO_ACKNOWLEDGE
+# Configure a specific incoming channel
+mp.messaging.incoming.my-channel.session-mode=AUTO_ACKNOWLEDGE
+# Configure a specific outgoing channel
+mp.messaging.outgoing.my-channel.session-mode=AUTO_ACKNOWLEDGE
+----
+
+Accepted values are:
+
+* `AUTO_ACKNOWLEDGE`
+* `SESSION_TRANSACTED`
+* `CLIENT_ACKNOWLEDGE`
+* `DUPS_OK_ACKNOWLEDGE`
+
+
+==== Configure the client id
+When dealing with _topics_, your subscription must need to be identified with a _client-id_.
+Configure the client id as follows:
+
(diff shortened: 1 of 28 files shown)


cescoffier:
@kenfinnigan if you want to have a look

cescoffier:
Remaining tasks:

- [ ] Documentation
- [ ] Tests with topics and durability
- [ ] Tests the different parameters

cescoffier:
The back pressure on the source side can be managed as follow:

* enqueue 1 task per request
* if request is Long.MAX, use the current loop approach

doc/jms.adoc
```diff
@@ -0,0 +1,315 @@
+== Interacting using JMS (Java Message Service)
+
+You can use SmallRye Reactive Messaging to receive and send JMS messages.
+
+=== Dependency
+
+To enable the Camel support, you need the following dependency:
```
kenfinnigan: "Camel"??

doc/jms.adoc
```diff
@@ -0,0 +1,315 @@
...
+==== Receiving Message
+
+You can decide to receive `io.smallrye.reactive.messaging.jms.ReceivedJmsMessage` which contains all the details of the underlying JMS Message.
+So you can access properties and metadata.
+You can also use the `unwrap` method to retrieve the underlying JMS message:
+
+[source,java]
+----
+@Incoming("my-channel")
+public CompletionStage<Void> consume(ReceiveJmsMessage<String> msg) {
```
kenfinnigan: "ReceivedJmsMessage"

smallrye-reactive-messaging-jms/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.eclipse</groupId>
+      <artifactId>yasson</artifactId>
+      <version>1.0.5</version>
+      <scope>test</scope>
+    </dependency>
+    
+    <dependency>
```
kenfinnigan: Are these spec APIs provided scope?

If not, probably should be

doc/jms.adoc
cescoffier: copy and paste of course...

cescoffier:
@kenfinnigan can you do another pass?

smallrye-reactive-messaging-jms/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+      <groupId>javax.enterprise</groupId>
```
kenfinnigan: didn't notice this earlier, but should we be mixing jakarta and java ee APIs?

Should these javax artifacts be updated to the jakarta ones?

cescoffier: Do we have a cdi in jakarta? It does not seem to be included in SamllRye-Parent.

cescoffier: I've updated to use jakarta common annotation everywhere.

doc/jms.adoc
```diff
@@ -0,0 +1,315 @@
...
+# Configure a specific outgoing channel
+mp.messaging.outgoing.my-channel.session-mode=AUTO_ACKNOWLEDGE
+----
+
+Accepted values are:
+
+* `AUTO_ACKNOWLEDGE`
+* `SESSION_TRANSACTED`
+* `CLIENT_ACKNOWLEDGE`
+* `DUPS_OK_ACKNOWLEDGE`
```
Emily-Jiang: The spec needs to state an `IllegalArgumentException` occurs if the value provided is not among the accepted values.

doc/jms.adoc
```diff
@@ -0,0 +1,315 @@
...
+[source, text]
+----
+# Queue:
+mp.messaging.outgoing.my-channel.connector=smallrye-jms
+mp.messaging.outgoing.my-channel.destination=my-queue
+
+# Topic:
+mp.messaging.outgoing.my-channel.connector=smallrye-jms
+mp.messaging.outgoing.my-channel.destination=my-topic
+mp.messaging.outgoing.my-channel.destination-type=topic
```
Emily-Jiang: The spec needs to state an `IllegalArgumentException` occurs if the destination-type is neither `queue` nor `topic`.

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsConnector.java
```diff
@@ -0,0 +1,150 @@
...
+    static final String CONNECTOR_NAME = "smallrye-jms";
+
+    @Inject
+    Instance<ConnectionFactory> factories;
+
+    @Inject
+    Instance<Jsonb> jsonb;
+
+    @Inject
+    @ConfigProperty(name = "smallrye.jms.threads.max-pool-size", defaultValue = "10")
```
Emily-Jiang: Do you think it makes sense to put the constants in an API (such as Interface) so that it can be viewed and changed easier in the future?

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSource.java
```diff
@@ -0,0 +1,182 @@
...
+        String selector = config.getOptionalValue("selector", String.class).orElse(null);
+        boolean nolocal = config.getOptionalValue("no-local", Boolean.TYPE).orElse(false);
+        boolean broadcast = config.getOptionalValue("broadcast", Boolean.TYPE).orElse(false);
+
+        Destination destination = getDestination(context, name, config);
+
+        JMSConsumer consumer;
+        if (config.getOptionalValue("durable", Boolean.TYPE).orElse(false)) {
+            if (!(destination instanceof Topic)) {
+                throw new IllegalStateException("Invalid destination, only topic can be durable");
```
Emily-Jiang: Previously, you throw `IllegalArgumentException` for this condition though. I suggest to stick with `IllegalArgumentException` for the sake of consistence.

doc/jms.adoc
cescoffier: By spec, do you mean the documentation? 

Also, this would result in a DeploymentException (caused by this IllegalArgumentException). 
I'm going to add a sentence in the documentation and a test (because we never know...)

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsConnector.java
cescoffier: Not necessary in an interface, but I've extracted the values into constant in the JMSConnector class.

doc/jms.adoc
cescoffier: I'm going to add a sentence in the documentation and a test (because we never know...)

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSource.java
cescoffier: Good catch, definitely wrong here.

doc/jms.adoc
cescoffier: Done.

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsConnector.java
cescoffier: Done.

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSource.java
cescoffier: Fixed.

doc/jms.adoc
cescoffier: Done.

PR #321, opened 2019-11-19
Update camel.adoc to fix a compile error

davsclaus:
Fixed issue in doc as reported on stackoverflow here https://stackoverflow.com/questions/58925759/cant-to-determine-why-smallrye-camel-code-snippet-is-not-compiling-no-subscrib


cescoffier:
Thanks @davsclaus !

PR #338, opened 2019-11-27
Add support for multiple incoming channels

cescoffier:
Like this:

```
@Incoming("a")
@Incoming("b")
@Outgoing("out")
public String process(String s) {
    // ...
}
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
 trim_trailing_whitespace = true
 end_of_line = lf
 insert_final_newline = true
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -1,41 +1,46 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
 
-  <parent>
-    <groupId>io.smallrye.reactive</groupId>
-    <artifactId>smallrye-reactive-messaging</artifactId>
-    <version>1.0.9-SNAPSHOT</version>
-  </parent>
+    <parent>
+        <groupId>io.smallrye.reactive</groupId>
+        <artifactId>smallrye-reactive-messaging</artifactId>
+        <version>1.0.9-SNAPSHOT</version>
+    </parent>
 
-  <artifactId>smallrye-reactive-messaging-api</artifactId>
-  <name>SmallRye Reactive Messaging API</name>
+    <artifactId>smallrye-reactive-messaging-api</artifactId>
+    <name>SmallRye Reactive Messaging API</name>
 
-  <dependencies>
-    <!--
-    This API does not depends on the upstream specification API as it's embedded.
-    The reason is because we need to modify the API to experiment and get feedback.
-    -->
-      <dependency>
-          <groupId>org.eclipse.microprofile.config</groupId>
-          <artifactId>microprofile-config-api</artifactId>
-      </dependency>
-  </dependencies>
+    <dependencies>
+        <!--
+        This API does not depend on the upstream specification API as it's embedded.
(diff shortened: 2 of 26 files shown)


.editorconfig
```diff
@@ -3,7 +3,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
```
kenfinnigan: Does this conform to the definition of the IDE config we have for SmallRye? ;-)

api/pom.xml
```diff
@@ -12,10 +12,14 @@
   <name>SmallRye Reactive Messaging API</name>
 
   <dependencies>
-    <dependency>
-      <groupId>org.eclipse.microprofile.reactive.messaging</groupId>
-      <artifactId>microprofile-reactive-messaging-api</artifactId>
-    </dependency>
+    <!--
```
kenfinnigan: Maybe leave the dependency there but commented out?

doc/advanced.adoc
```diff
@@ -124,6 +124,22 @@ Emitter<String> emitter;
 The default buffer size (`127`) can be configured using the `smallrye.messaging.emitter.default-buffer-size` property set using MicroProfile Config.
 Note that this value is only applied when `OnOverflow` is not used.
 
+=== Repeatable `@Incoming`
+
+You can use several `@Incoming on a single method:
+
+[source,java]
```
kenfinnigan: Do we need to document `@Incomings` as well?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/incomings/IncomingsTest.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+    }
+
+    @ApplicationScoped
+    public static class MyBeanUsingMultipleIncomingsAndMerge {
+
+        private List<String> list = new CopyOnWriteArrayList<>();
+
+        @Incoming("a")
```
kenfinnigan: Does one of these need to test `@Incomings`?

api/pom.xml
cescoffier: Yes, I was wondering about that, probably a good idea.

.editorconfig
cescoffier: nope, but the maven build does the job :-D

kenfinnigan: consistency!! ;-)

doc/advanced.adoc
cescoffier: Good question, I don't expect users to use it directly, but use multiple `@Incoming`. We should maybe say so in the javadoc. 

Unfortunately, it's how the repeatable annotations work, you create a container that would be "used" automatically.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/incomings/IncomingsTest.java
cescoffier: See my previous comment, do we want @Incomings to be a public API?
What's the other MP spec do?

doc/advanced.adoc
kenfinnigan: Ok, if it's purely for an implementation thing, then should `@Incomings` be in the api artifact or in the impl? And agree we need to doc it to indicate it's not for use

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/incomings/IncomingsTest.java
cescoffier: Ok, as discussed in the call, I've added javadoc and explain how it can be used.

api/src/main/java/io/smallrye/reactive/messaging/annotations/Incomings.java
```diff
@@ -8,13 +8,37 @@
...
+ * This class is used to allow multiple {@link Incoming} declarations.
+ * You can either use:
+ *
+ * <pre><code>
+ * &#64;Incoming("a")
+ * &#64;Incoming("b")
+ * public void consume(T t);
+ * </code></pre>
+ *
+ * Or use the {@link Incomings} annotation as container:
```
kenfinnigan: so we're saying it's ok to use `@Incomings` now?

For some reason I thought the resolution was to javadoc not to use it? But I may have been distracted ;-)

api/pom.xml
cescoffier: Done.

api/src/main/java/io/smallrye/reactive/messaging/annotations/Incomings.java
cescoffier: :-D

I just follow what other specs have done, and let the user decide which flavor he prefers. 
See https://javaee.github.io/javaee-spec/javadocs/javax/annotation/Resources.html, too succinct for my taste, and does not say anything about the usage. 

So as we can't prevent the user to use it, let's explain how.

kenfinnigan: fair enough

PR #339, opened 2019-11-27
New emitter API

cescoffier:
Fix #297 
Supersedes #293 

@Emily-Jiang I have taken your PR and made sure we support both the old and new Emitter interface.  I've also added tests for the old emitter and updated the documentation.

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
@@ -37,10 +37,17 @@ public @interface OnOverflow {
     enum Strategy {
         /**
          * Buffers <strong>all</strong> values until the downstream consumes it.
-         * This creates an unbound buffer if not used in conjunction with {@link #bufferSize()} to limit the size of the
-         * buffer. If the buffer is full, an error is propagated.
+         * This creates a buffer with the size specified by {@link #bufferSize()} if present.
+         * Otherwise, the size will be the value of the config property
+         * <strong>mp.messaging.emitter.defult-buffer-size</strong>.
+         * If the buffer is full, an error is propagated.
          */
         BUFFER,
+        /**
+         * Buffers <strong>all</strong> values until the downstream consumes it.
+         * This creates an unbound buffer. If the buffer is full, an error is propagated.
+         */
+        UNBOUNDED_BUFFER,
         /**
          * Drops the most recent value if the downstream can't keep up. It means that new value emitted by the upstream
          * are ignored.
@@ -71,7 +78,8 @@ public @interface OnOverflow {
 
     /**
      * @return the size of the buffer when {@link Strategy#BUFFER} is used. If not set and if the {@link Strategy#BUFFER}
-     *         strategy is used, an unbound buffer is used.
+     *         strategy is used, the buffer size will be defaulted to the value of the config property
+     *         mp.messaging.emitter.defult-buffer-size.
      */
     long bufferSize() default 0;
 
--- a/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
@@ -40,12 +40,12 @@ public @interface OnOverflow {
          * This creates a buffer with the size specified by {@link #bufferSize()} if present.
          * Otherwise, the size will be the value of the config property
          * <strong>mp.messaging.emitter.defult-buffer-size</strong>.
-         * If the buffer is full, an error is propagated.
+         * If the buffer is full, an error will be propagated.
          */
         BUFFER,
         /**
          * Buffers <strong>all</strong> values until the downstream consumes it.
-         * This creates an unbound buffer. If the buffer is full, an error is propagated.
+         * This creates an unbound buffer. If the buffer is full, the application will die of {@code OutOfMemory}.
          */
         UNBOUNDED_BUFFER,
         /**
--- a/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
@@ -39,17 +39,19 @@ public @interface OnOverflow {
          * Buffers <strong>all</strong> values until the downstream consumes it.
          * This creates a buffer with the size specified by {@link #bufferSize()} if present.
          * Otherwise, the size will be the value of the config property
-         * <strong>mp.messaging.emitter.defult-buffer-size</strong>.
+         * <strong>mp.messaging.emitter.default-buffer-size</strong>.
          * If the buffer is full, an error will be propagated.
          */
         BUFFER,
+
         /**
          * Buffers <strong>all</strong> values until the downstream consumes it.
-         * This creates an unbound buffer. If the buffer is full, the application will die of {@code OutOfMemory}.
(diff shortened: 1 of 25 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ChannelProducer.java
```diff
@@ -72,19 +72,51 @@
...
+    @Deprecated
+    <T> io.smallrye.reactive.messaging.annotations.Emitter<T> produceEmitterLegacy(InjectionPoint injectionPoint) {
+        LegacyEmitterImpl emitter = new LegacyEmitterImpl(getEmitter(injectionPoint));
+        return cast(emitter);
+    }
+
     @Produces
     @Stream("")
-    <T> Emitter<T> produceEmitterLegacy(InjectionPoint injectionPoint) {
+    @Deprecated
```
kenfinnigan: Should this have `@Deprecated`?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MediatorManager.java
```diff
@@ -63,7 +63,7 @@
     private final List<AbstractMediator> mediators = new ArrayList<>();
 
     @Inject
-    @ConfigProperty(name = "smallrye.messaging.emitter.default-buffer-size", defaultValue = "127")
+    @ConfigProperty(name = "mp.messaging.emitter.default-buffer-size", defaultValue = "127")
```
kenfinnigan: is this because we intend to move it to the spec soon and don't want to break people once it is?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ChannelProducer.java
cescoffier: I was wondering if Deprecated made sense here as it's very internal. I believe it would help me removing the old emitter once we decide to remove it completely. But no strong opinion.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MediatorManager.java
cescoffier: Yes, as it's a matter of weeks I believe it's ok.
We could consider this as a breaking change, but I don't believe anyone uses this property.

cescoffier:
@kenfinnigan are we good about the property? If so let's merge this one.

We could also support both property for a while.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ChannelProducer.java
kenfinnigan: Sorry, was referring to `produceEmiiterStream()` as having `@Deprecated`.

That's the new method right? So was confused as to why it's marked deprecated

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MediatorManager.java
kenfinnigan: ok, that's fine

cescoffier: actually no, it's not fine, the documentation mentions the property, we need to support both.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ChannelProducer.java
cescoffier: I've added the javadoc and annotation to all the injection method from ChannelPRoducer explaining what are they injecting and why some of them are deprecated.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MediatorManager.java
cescoffier: I've re-introduced the management of both properties and updated the doc.

cescoffier:
@kenfinnigan sorry to bother you again, I've fixed the property management issue and added javadoc to `ChannelProducer` explaining the deprecation.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/LegacyEmitterInjectionTest.java
```diff
@@ -0,0 +1,442 @@
...
+            return list;
+        }
+
+        public void run() {
+            emitter.send("a");
+            emitter.send("b");
+            /*
+             * try { emitter.send(null); } catch (IllegalArgumentException e) { caught =
+             * true; }
+             */
```
xstefank: this is intentionally commented? On line https://github.com/smallrye/smallrye-reactive-messaging/pull/339/files#diff-4a279cdc72ce161238860cec0e680dcdR108 it seems to be checked whether caught == true.

api/src/main/java/io/smallrye/reactive/messaging/annotations/OnOverflow.java
```diff
@@ -53,7 +62,7 @@
         FAIL,
 
         /**
-         * Keeps only the latest value, dropping any previous value if the downstream can't keep up.
+         * Keeps only the latest value, dropping any previous values if the downstream can't keep up.
```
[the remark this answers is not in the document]
cescoffier: All of them, only keeping the very last received item.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ChannelProducer.java
```diff
@@ -41,12 +53,28 @@
...
     @Stream("")
+    @Deprecated
     <T> Flowable<T> producePublisherLegacy(InjectionPoint injectionPoint) {
         return producePublisher(injectionPoint);
     }
 
+    /**
+     * Injects {@code PublisherBuilder<Message<X>>} and {@code PublisherBuilder<X>}
+     *
+     * @param injectionPoint the injection poirt
```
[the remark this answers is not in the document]
cescoffier: Ok... did I copy and paste a sentence with a typo ;-)

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/LegacyEmitterInjectionTest.java
Emily-Jiang: Yes, it is. With the new change, you cannot pass 'null' any more, because compiler cannot figure out which method to match null because of the method overloading.
@cescoffier maybe it is better to delete this block. My bad for leaving them.

cescoffier: Great catch, fixing...

cescoffier: Fixed.

PR #347, opened 2019-12-04
Support for message headers

cescoffier:


cescoffier:
Just a head's up.

I've started adding header support to connectors. I did Kafka and HTTP. I will continue shortly.

pgrabarsky:
Hi Clement,

1. The way headers are stored in a `Message` is exposed in the API as Headers extends `LinkedHashMap`. (Or, a new instance will have to be created each time `Message.getHeaders()` is called but I guess it is not the idea...)

2. The way it is used in `KafkaHeaders` is to have a lot of keys in this Map. And these keys are used for each piece of data (e.g. TOPIC, PARTITION...). Then the `KafkaHeaders` is fully static and can be used to facilitate access to the data, e.g. `KafkaHeaders.getKafkaTopic(message, "")`.

The possible improvements that I can see are:

1. Store the full structure of a `KafkaHeader` as a header. This would look like:
```
public interface IncomingKafkaHeader {
    public static final String KEY = "kafka.incoming";
    String getTopic();
    ...
}
IncomingKafkaHeader ikh = x.get(IncomingKafkaHeader.KEY);
String topic = ikh.getTopic()
```

This is an enabler for a possible future injections of parameters in the processor method or using CDI.

2. Hide the internal structure of the `Headers` storage inside the `Message`. To do so, a method `Message.getHeader(String key)` could be used to retrieve an entry. , e.g.:

```
IncomingKafkaHeader ikf = message.getHeader(IncomingKafkaHeader.KEY);
```

 (e.g. With header being a `MessageContext`:
   https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/reactive-messaging/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/core/Message.java#L12
   https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/business-app/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/business/app/MyOutgoingProcessor.java#L36
   ). 


3. It is also possible to require the headers to implement an interface with a method providing their KEY (e.g. https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/reactive-messaging/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/core/MessageContext.java#L11).

Doing so enable this kind of API (shorter and less error-prone) to specify a Kafka target for the message:

```
KafkaTargetHeader kth = new KafkaTargetHeader("topic_x");
message.addHeader(kth); // instead of message.addHeader(KafkaTargetHeader.KEY, kth);
```

Also, it enable to use a non-associative container (e.g. a simple array) to store Headers in the message. This could be useful to reduce the memory impact when there is a of of messages in flight.

4. Then some suggestions for the name instead of Header:

* Context like in `MessageContext` ;-)

```
IncomingKafkaContext ikc = message.getContext(IncomingKafkaContext.KEY);
```

* Meta

```
IncomingKafkaMeta ikn = message.getMeta(IncomingKafkaMeta.KEY);
```

cescoffier:
Thanks @pgrabarsky! Great feedback.

Let's start with the last proposal (4) - the name. I discussed it a bit last week, and I'm going toward `metadata`. It avoids the confusion with `headers` and conveys the right meaning. 

So it would be something like:

```
MessageMetadata meta = message.getMetadata();
```

Also, you mentioned it, and I agree, I will add (this is your point (2)):

```
Whatever m = message.getMetadata(key);
Whatever m = message.getMetadata(key, new DefaultWhatever());
```

About (1) - I'm a bit puzzled. It may improve usability but make the immutable assumption much harder to verify and the stored object can be updated. This would create visibility issues. I was thinking of doing something like:

```
IncomingKafkaMessage<String, String> m = IncomingKafkaMessage.unwrap(message);
String topic = m.getTopic();
```

The `unwrap` method would create a high-level and read-only object to access the metadata. It would also manage the key and payload type. 

About (3) - I like the idea. I'd like the hear the opinion from @kenfinnigan

cescoffier:
I have renamed headers into metadata and provided Message method to access the metadata directly.

pgrabarsky:
Hi Clement,

I'm a bit puzzled by the `IncomingKafkaMessage` stuff ;-).

I probably miss something, but it looks to me that there are two assumptions behind this piece of code:

* The actual message dynamic type is `IncomingKafkaMessage`.

However, the actual message type might not always be `IncomingKafkaMessage`: Besides the discussion of whether a connector providor should extends/implements a `Message`, the (developer) end-user still could want to create a new child message. And it is then likely to have `Message` using the framework implementation propagating inside the stream. However the Kafka metadata should still be accessible.

* There is only one kind of metadata, matching the actual message type.

However, it is a great feature to support several kind of metadata into a message. One example of it is https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/corporate-framework/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/corporate/framework/EventContext.java. This type of metadata is created from the initial message payload.
In this example, there is a unique id which is similar to a trace-id but is more functional. It can be used to connect the several distributed components logs participating in a single flow. More generally, it's an example of having messaging system dependent and messaging system non-depend metadata. Other use cases can be considered, e.g. `Message` coming from RabbitMQ could be send to Kafka by adding a `KafkaTargetMetadata` containing the topic name.

On top of these, the `unwrap` approach would lead to something like:

```
IncomingKafkaMessage<String, String> m = IncomingKafkaMessage.unwrap(message);
String topic = m.getTopic();
X m.getPayload();

EventMetadataMessage em = EventMetadataMessage.unwrap(message);
String em = em.getUniqueMessageId();
X em.getPayload();
```

Then each type of metadata would need to re-implement the `Message` interfaces and delegate the payload & acknowlegment method calls.

Hence, I feel more like having metadata seen purely as a message aggregate.

Philippe

cescoffier:
The `unwrap` parameter does not have to be a `Message` implementation. You could imagine something like:

```
Person person = Person.unwrap(message);
```

_unwrap_ might not be the right name actually. Basically this is a factory method that would create an object from a message object. In general, it would read the metadata of the given message. 

So you could imagine something like:

```
Message<String> message = incoming;

Person person = Person.unwrap(message); // read the `firstName` and `lastName` metadata
SpanContext sc = Tracing.unwrap(message).getSpanContext(); // Retrieves the Span Context (open tracing)

// Create a new message
X newPayload = ...;
Message<X> anotherMessage = message.withPayload(newPayload); // metadata from message are copied

// HAnd-made addition and propagation
Message<X> yetAnotherMessage = message.withPayload(newPayload).withMetadata(MetadataBuilder.from(message.getMetadata()).with("foo", "bar");
```
The last line could be improved with an "add" method. 
Also, not that each `with` is returning a new instance of Message (because of immutability).

pgrabarsky:
:thumbsup:

Ok then it looks better to me!

If I understand  correctly:

```
Person person = Person.unwrap(message);
// Is an equivalent of:
Person person = message.getMetadata(Person.KEY);
```

Just a side note as it is another topic: To enable a message to be a child of several ones, a `MessageBuilder` is likely to be needed (e.g. https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/reactive-messaging/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/core/MessageBuilder.java#L6).

cescoffier:
About the `MessageBuilder` I agree, we start seeing a need for this.

cescoffier:
I Implemented a new metadata model using objects:

```
Message<String> message = Message.of("hello", Metadata.of(new Foo(...), new Bar(...));
Optional<Foo> foo = message.getMetadata(Foo.class);
Optional<Bar> bar = message.getMetadata(Bar.class);
```

AMQP, Kafka, JMS and HTTP connectors have been updated, as well as the documentation.

CC @pgrabarsky  @kenfinnigan

kenfinnigan:
I've taken a look through the changes and I have a few comments based on my limited understanding.

Though I love things being type-safe, as we get nice errors when compiling, given reactive messaging is working across many different types of connectors I prefer the flexible approach of a single bucket, a map, that captures all the headers. It's still possible for developers to create typed metadata classes for themselves that interrogate the map to provide a more type-safe access pattern if they so choose. But we don't lose the flexibility of having just a bucket of things.

I also think it's important that we don't require downstream consumers of the messages to know the types of values in the header to be able to process them. We don't always know who is developing a service that will receive a message we sent.

I may be wrong in some of this understanding, apologies if that's the case.

PR #348, opened 2019-12-09
Export a metric for the count of messages on each channel

Azquelt:
Add support for the spec proposal here: eclipse/microprofile-reactive-messaging#31
Spec PR: eclipse/microprofile-reactive-messaging#84

Currently, this introduces a dependency from the provider component on the MP Metrics API, though the provider will still work (without metrics) if there's no MP Metrics implementation present at runtime.

If that's not desirable, then the `MetricDecorator` could be moved out to its own component.

I also fixed a couple of issues I encountered along the way as separate commits, though they could be moved out to separate PRs instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,9 +64,11 @@
     <microprofile-reactive-messaging.version>1.0</microprofile-reactive-messaging.version>
     <microprofile-reactive-streams.version>1.0.1</microprofile-reactive-streams.version>
     <microprofile-config.version>1.3</microprofile-config.version>
+    <microprofile-metrics-api.version>2.0</microprofile-metrics-api.version>
 
     <smallrye-reactive-streams-ops.version>1.0.10</smallrye-reactive-streams-ops.version>
     <smallrye-config.version>1.5.1</smallrye-config.version>
+    <smallrye-metrics.version>2.1.5</smallrye-metrics.version>
 
     <asciidoctorj.version>1.5.8.1</asciidoctorj.version>
 
@@ -146,11 +148,23 @@
         <version>${microprofile-config.version}</version>
         <scope>provided</scope>
       </dependency>
+      <dependency>
+          <groupId>org.eclipse.microprofile.metrics</groupId>
+          <artifactId>microprofile-metrics-api</artifactId>
+          <version>${microprofile-metrics-api.version}</version>
+        <scope>provided</scope>
+      </dependency>
 
       <dependency>
         <groupId>io.smallrye.config</groupId>
         <artifactId>smallrye-config</artifactId>
         <version>${smallrye-config.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-metrics</artifactId>
+        <version>${smallrye-metrics.version}</version>
       </dependency>
 
       <dependency>
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
@@ -232,7 +232,7 @@ public class ProcessorMediator extends AbstractMediator {
         if (configuration.consumption() == MediatorConfiguration.Consumption.PAYLOAD) {
             this.processor = ReactiveStreams.<Message> builder()
                     .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> {
+                    .<Message> map(input -> {
                         Object result = invoke(input.getPayload());
                         if (configuration.getAcknowledgment() == Acknowledgment.Strategy.POST_PROCESSING) {
                             return Message.of(result, () -> input.ack());
@@ -244,7 +244,7 @@ public class ProcessorMediator extends AbstractMediator {
         } else {
             this.processor = ReactiveStreams.<Message> builder()
                     .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> {
+                    .<Message> map(input -> {
                         Object result = invoke(input);
                         if (configuration.getAcknowledgment() == Acknowledgment.Strategy.POST_PROCESSING) {
                             return Message.of(result, () -> input.ack());
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
@@ -1,0 +1,20 @@
+package io.smallrye.reactive.messaging;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
+
+/**
+ * SPI to allow extension of publishers included in the final graph
+ */
+public interface PublisherDecorator {
+
+    /**
+     * Decorate a publisher
+     * 
+     * @param publisher the publisher to decorate
+     * @param the name of the channel to which this publisher publishes
+     * @return the extended publisher
+     */
+    public PublisherBuilder<? extends Message> decoratePublisher(PublisherBuilder<? extends Message> publisher, String name);
+
+}
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
@@ -12,9 +12,9 @@ public interface PublisherDecorator {
      * Decorate a publisher
      * 
      * @param publisher the publisher to decorate
-     * @param the name of the channel to which this publisher publishes
+     * @param channelName the name of the channel to which this publisher publishes
      * @return the extended publisher
      */
-    public PublisherBuilder<? extends Message> decoratePublisher(PublisherBuilder<? extends Message> publisher, String name);
+    public PublisherBuilder<? extends Message> decorate(PublisherBuilder<? extends Message> publisher, String channelName);
 
 }
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
@@ -76,7 +76,7 @@ public class ConfiguredChannelFactory implements ChannelRegistar {
     }
 
     private List<String> getConnectors(BeanManager beanManager, Class<?> clazz) {
-        return beanManager.getBeans(clazz).stream()
+        return beanManager.getBeans(clazz, Any.Literal.INSTANCE).stream()
                 .map(BeanAttributes::getQualifiers)
                 .flatMap(set -> set.stream().filter(a -> a.annotationType().equals(Connector.class)))
                 .map(annotation -> ((Connector) annotation).value())
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
@@ -10,6 +10,7 @@ import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.BeanAttributes;
 import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.CDI;
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.config.Config;
@@ -23,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import io.smallrye.reactive.messaging.ChannelRegistar;
 import io.smallrye.reactive.messaging.ChannelRegistry;
+import io.smallrye.reactive.messaging.PublisherDecorator;
 
 /**
  * Look for stream factories and get instances.
@@ -142,7 +144,13 @@ public class ConfiguredChannelFactory implements ChannelRegistar {
         IncomingConnectorFactory mySourceFactory = incomingConnectorFactories.select(ConnectorLiteral.of(connector))
                 .stream().findFirst().orElseThrow(() -> new IllegalArgumentException("Unknown connector for " + name + "."));
 
-        return mySourceFactory.getPublisherBuilder(config);
+        PublisherBuilder<? extends Message> publisher = mySourceFactory.getPublisherBuilder(config);
+
+        for (PublisherDecorator decorator : CDI.current().select(PublisherDecorator.class)) {
+            publisher = decorator.decoratePublisher(publisher, name);
+        }
+
+        return publisher;
     }
 
     private SubscriberBuilder<? extends Message, Void> createSubscriberBuilder(String name, Config config) {
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
@@ -147,7 +147,7 @@ public class ConfiguredChannelFactory implements ChannelRegistar {
         PublisherBuilder<? extends Message> publisher = mySourceFactory.getPublisherBuilder(config);
 
         for (PublisherDecorator decorator : CDI.current().select(PublisherDecorator.class)) {
-            publisher = decorator.decoratePublisher(publisher, name);
+            publisher = decorator.decorate(publisher, name);
         }
 
         return publisher;
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
@@ -10,7 +10,6 @@ import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.BeanAttributes;
 import javax.enterprise.inject.spi.BeanManager;
-import javax.enterprise.inject.spi.CDI;
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.config.Config;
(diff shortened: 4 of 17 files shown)


cescoffier:
The CI issue is due to a network issue, unrelated to the PR.

pom.xml
```diff
@@ -146,13 +148,25 @@
...
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-config</artifactId>
         <version>${smallrye-config.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-metrics</artifactId>
+        <version>${smallrye-metrics.version}</version>
+      </dependency>
```
cescoffier: should it be in "test" scope?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -244,7 +244,7 @@ private void processMethodReturningIndividualPayloadAndConsumingIndividualItem()
         } else {
             this.processor = ReactiveStreams.<Message> builder()
                     .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> {
+                    .<Message> map(input -> {
```
cescoffier: Any reason why the explicit type is required now?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
```diff
@@ -0,0 +1,20 @@
...
+public interface PublisherDecorator {
+
+    /**
+     * Decorate a publisher
+     * 
+     * @param publisher the publisher to decorate
+     * @param the name of the channel to which this publisher publishes
+     * @return the extended publisher
+     */
+    public PublisherBuilder<? extends Message> decoratePublisher(PublisherBuilder<? extends Message> publisher, String name);
```
cescoffier: `decorate` is shorter and good enough.
Actually the current name is misleading because it gets a PublisherBuilder.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
```diff
@@ -142,7 +144,13 @@ private static String getConnectorAttribute(Config config) {
         IncomingConnectorFactory mySourceFactory = incomingConnectorFactories.select(ConnectorLiteral.of(connector))
                 .stream().findFirst().orElseThrow(() -> new IllegalArgumentException("Unknown connector for " + name + "."));
 
-        return mySourceFactory.getPublisherBuilder(config);
+        PublisherBuilder<? extends Message> publisher = mySourceFactory.getPublisherBuilder(config);
+
+        for (PublisherDecorator decorator : CDI.current().select(PublisherDecorator.class)) {
+            publisher = decorator.decoratePublisher(publisher, name);
```
cescoffier: What about the order? Should we introduce some kind of `ordinal`?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+
+@ApplicationScoped
+public class MetricDecorator implements PublisherDecorator {
+
+    @Inject
+    private Instance<MetricRegistry> registryInstance;
+
+    @Override
+    public PublisherBuilder<? extends Message> decoratePublisher(PublisherBuilder<? extends Message> publisher,
+            String channelName) {
```
cescoffier: the interface use `name` as parameter name, here you use `channelName`, should it be consistent?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+        if (registryInstance.isResolvable()) {
+            return publisher.peek(incrementCount(channelName));
+        } else {
+            return publisher;
+        }
+    }
+
+    private <T extends Message> Consumer<T> incrementCount(String channelName) {
+        MetricRegistry registry = registryInstance.get();
```
cescoffier: can we save the retrieval and cache the registry.
Same question for the counter?

Should it be a `Consumer<? extends Message<P>>` ?

cescoffier:
About the dependency, I would be interested in @kenfinnigan POV.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
Azquelt: No, it showed up as an error in eclipse before I started working.

> Type mismatch: cannot convert from Processor<Message,Message<Object>> to Processor<Message,Message>

As it builds correctly on the command line, I suspect it's an eclipse issue, though it's possible I have something misconfigured in my workspace. I can take the change out again if you want.

pom.xml
Azquelt: Err, possibly?

Currently I'd followed the same pattern as for `smallrye-config` where it's declared without a scope in the dependency management block but put into "test" scope by projects which use it as a test dependency.

For config, it looks like your quickstarts include it as a compile dependency but your modules include it as a test dependency. I'd assume that metrics would be used in a similar way.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: Yeah, it should, I'll change the interface since I think I don't think it's particularly obvious that it's the name of the channel

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: Not actually sure. As it's a private method and we're passing the return value to `peek()` which is expecting `Consumer<? super Message>`, it could be `Consumer<Message>` or maybe `Consumer<Object>` would make more sense as it actually doesn't care at all about what the object it consumes is.

Azquelt: As a side note, it looks like you use the raw `Message` type everywhere, so I've continued that.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
Azquelt: We could do. At the moment there's only one. Would you like me to add an ordinal now or wait until we have another decorator use case?

pom.xml
cescoffier: Yes, that should be ok, we just need to be sure it's in the `test` scope in the implementation module.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
cescoffier: oh ok, no problem.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
cescoffier: We can wait I believe.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
cescoffier: ok.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
kenfinnigan: +1

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
```diff
@@ -0,0 +1,20 @@
+package io.smallrye.reactive.messaging;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
+
+/**
+ * SPI to allow extension of publishers included in the final graph
+ */
+public interface PublisherDecorator {
```
kenfinnigan: Should this be in an SPI package?

Is this class intended for possible spec inclusion in the future? If so, it should probably be in the API artifact instead

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
+
+import io.smallrye.reactive.messaging.PublisherDecorator;
+
+@ApplicationScoped
+public class MetricDecorator implements PublisherDecorator {
+
+    @Inject
+    private Instance<MetricRegistry> registryInstance;
```
kenfinnigan: I presume we're using `Instance` here so we can lazy load?

Any particular advantage over just grabbing it from the start?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+        if (registryInstance.isResolvable()) {
+            return publisher.peek(incrementCount(channelName));
+        } else {
+            return publisher;
+        }
+    }
+
+    private <T extends Message> Consumer<T> incrementCount(String channelName) {
+        MetricRegistry registry = registryInstance.get();
+        Counter counter = registry.counter("mp.messaging.message.count", new Tag("channel", channelName));
```
kenfinnigan: Should we be creating a new `Tag` instance every time, or should we be caching them?

Could we even cache the `Counter` instance by channel name? `registry.counter()` internally calls a `synchronized` method to retrieve the counter, so we probably want to minimize calls to it?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: I was using an `Instance` here so that it would be able to run whether or not the `MetricRegistry` is present, on the basis that someone might want to use reactive messaging without metrics at all.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: This code only gets called once for each channel, the last line returns the lambda that gets called for each message.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
kenfinnigan: Would you get a CNFE when `MetricsDecorator` is instantiated if metrics were not present?

Though an instance of it may not be there, it still needs the Type to be able to create the `Instance`, or am I wrong?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
kenfinnigan: So `publisher.peek()` is a continual operation?

The javadoc reads like it only operates on the current state of the publisher, which to me implies any items that are added to the stream after `peek()` is called wouldn't have this applied.

I'm still grokking reactive so I could be way off base with my thinking

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: You're right, `MetricDecorator` needs the `MetricRegistry` class to be loadable, so it will always require that the Metrics API is present, and I've added a compile dependency on the Metrics API to the pom for the provider.

However, as it's written, it doesn't require an MP metrics implementation to be present to provide the `MetricsRegistry` implementation and injectable bean.

Note: If we didn't want the provider to have a dependency on the metrics API, we could move the `MetricDecorator` to a separate module.

kenfinnigan: Sorry, missed the dependency when looking at this.

Thanks

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
Azquelt: I'm not sure, I wasn't really thinking about it being used outside of this project. Would you like it moved to an SPI project?

I don't think there are any plans to put something like this in the specification. If it turns out to be useful for more than just the metrics integration then we could add it, though it would be good to have a few more use cases so we know if there's more information that needs to be provided to the decorator.

kenfinnigan: I wasn't sure if this could be something that's standardized at some point, even if only for how metrics are handled with Publishers.

Not wedded to any particular approach, was curious as to future thoughts

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: Yes, the consumer passed to `peek()` is called for each element emitted from the publisher. It's like a `map()` which doesn't alter the input.

pom.xml
Azquelt: Cool, I've now added tests and I've used `smallrye-metrics` as a test dependency.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherDecorator.java
Azquelt: Done

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java
Azquelt: Interface updated

cescoffier:
Damned... sorry, my fault, could you rebase your PR?

Azquelt:
Yep, will do that tomorrow.

>

Azquelt:
Rebase done.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
```diff
@@ -142,7 +144,13 @@ private static String getConnectorAttribute(Config config) {
         IncomingConnectorFactory mySourceFactory = incomingConnectorFactories.select(ConnectorLiteral.of(connector))
                 .stream().findFirst().orElseThrow(() -> new IllegalArgumentException("Unknown connector for " + name + "."));
 
-        return mySourceFactory.getPublisherBuilder(config);
+        PublisherBuilder<? extends Message> publisher = mySourceFactory.getPublisherBuilder(config);
+
+        for (PublisherDecorator decorator : CDI.current().select(PublisherDecorator.class)) {
```
cescoffier: For my own information (remember, I'm a CDI noob), what's the difference between this and injecting an `Instance` and iterating on the set of providers?

Azquelt: There's no difference, except that `CDI.current()` has to look up the CDI container first. Injecting an `Instance` would be better. I'll change that here.

The other place I used `CDI.current()` was in the `AbstractMediator` which isn't a CDI bean, so we'd have to pass the instance in there instead, or change the place that the decoration is done.

I'll have a poke around and see if I can get rid of the calls to `CDI.current()` as I don't really like them.

Azquelt: I've pushed a new commit to inject an instance instead of using `CDI.current()`

Azquelt: There were a couple of things I wasn't sure about:
1. I've used a setter on AbstractMediator to pass in the decorators, which must be called. At the moment it would NPE when applying the decorators if it wasn't set. We could add a null check first or pass it through the constructor instead (although then you need to pass it in through the `MediatorFactory` instead and pass it down through the constructor of each subclass).
2. Your `AbstractMediator.setInvoker()` method is `synchronized` and I couldn't work out why that was necessary but it made be concerned that it might also be necessary for `setDecorators()`.

cescoffier: Not sure why it's synchronizer. I need to check.
I believe the usage of a setter method is good, we can always revisit this later if we found an issue.

cescoffier:
Sorry, yet another rebase required. 
Then I believe we can merge.

Azquelt:
Rebase done

cescoffier:
Thanks, tracking the CI progress.

Azquelt:
Could squash up all the review changes I guess?

cescoffier:
Merged! Thanks!

Azquelt:
Cool :+1:

cescoffier:
no, no need for squashing.

Issue #349, opened 2019-12-10
More SSL configuration options for AMQP


famartinrh:
AmqpConnector currently adds options to configure the amqp client via config properties and for ssl configuration there are still a few missing settings. 

The question is, would be considered to add the posibility to inject an optional `AmqpClientOptions` ?, this way there won't be a need to add more and more config properties and the AmqpConnector could be configured producing by CDI an AmqpClientOptions with the appropiate settings needed. I understand there are missing details like how to deal with the two config ways, allow both and merge properties, only allow one,... this is something could be discussed if this proposal is accepted.

cescoffier:
I like this idea! 

Should we inject a AmqpClientOptions or directly an AmqpClient (already configured) ? 

We also need to check how we will handle multiple channels using different clients.

famartinrh:
nice!

I think AmqpClientOptions is the thing, it will ease the moving to build time in quarkus, and the AmqpClient used here is the axle one and people maybe won't use it and we will have to convert it.... I think AmqpClientOptions is just fine.

I'll have to think about dealing with multiple channels

famartinrh:
We have an example in JmsConnector!!

We can add a config property named `client-options-name` or `client-name`, if we go with the AmqpClient approach, ie:
```
mp.messaging.incoming.foo.client-options-name=myfooclientoptions
```
and then we will have to produce a named instance like 
```
@Produces @Named("myfooclientoptions")
AmqpClientOptions produceOptions() {...
```
we could create our own qualifier, but named is enough.

Then I would implement that if there isn't any `client-options-name` configured but there is an instance of AmqpClientOptions then the amqp client would be created from that options instance.

And finally, to keep backwards compatibility, I would implement it to behave as always if there isn't any instance of AmqpClientOptions

cescoffier:
I would use `@Named` (even if it's a bit of an abuse of the annotation). I already use it in the JMS connector, let's continue with it.

However, you should, as in the JMS connector I believe, use `Instance<AmqpClientOptions`, and manage the cases:
1. no match
2. multiple matches

cescoffier:
@famartinrh, should we close this one as #352 has been merged?

PR #352, opened 2019-12-12
AMQP connector - configure amqp client via CDI producing an AmqpClientOptions bean

famartinrh:
Related to #349 
This PR adds a new config property `client-options-name` to allow configure the amqp connector via CDI ie:
```properties
mp.messaging.incoming.data.client-options-name=my-topic-config
```
```Java
    @Produces
    @Named("my-topic-config")
    public AmqpClientOptions options() {
        return new AmqpClientOptions()
                .setHost("localhost")
                .setPort(5672)
                .setUsername("smallrye")
                .setPassword("smallrye");
    }
```

This eases the configuration of features like SSL to connect with remote AMQP brokers.
Also note that this functionality is backwards compatible.

### DIFF
--- a/examples/amqp-quickstart/src/main/resources/META-INF/microprofile-config.properties
+++ b/examples/amqp-quickstart/src/main/resources/META-INF/microprofile-config.properties
@@ -13,8 +13,10 @@ smallrye.messaging.sink.data.durable=true
 smallrye.messaging.sink.my-channel.connector=smallrye-amqp
 smallrye.messaging.sink.my-channel.address=hello
 smallrye.messaging.sink.my-channel.durable=true
+smallrye.messaging.sink.my-channel.client-options-name=my-topic-config
 
 # AMQP source (we read from)
 smallrye.messaging.source.my-topic.connector=smallrye-amqp
 smallrye.messaging.source.my-topic.address=hello
 smallrye.messaging.source.my-topic.durable=true
+smallrye.messaging.source.my-topic.client-options-name=my-topic-config2
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -1,10 +1,12 @@
 package io.smallrye.reactive.messaging.amqp;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.annotation.PostConstruct;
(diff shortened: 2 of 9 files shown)


famartinrh:
The failures in java11 and java12 pipelines are related to mqtt which presumably isn't related with this PR :)

examples/amqp-quickstart/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -13,8 +13,10 @@ smallrye.messaging.sink.data.durable=true
 smallrye.messaging.sink.my-channel.connector=smallrye-amqp
 smallrye.messaging.sink.my-channel.address=hello
 smallrye.messaging.sink.my-channel.durable=true
+smallrye.messaging.sink.my-channel.client-options-name=my-topic-config
 
 # AMQP source (we read from)
 smallrye.messaging.source.my-topic.connector=smallrye-amqp
 smallrye.messaging.source.my-topic.address=hello
 smallrye.messaging.source.my-topic.durable=true
+smallrye.messaging.source.my-topic.client-options-name=my-topic-config
```
cescoffier: It is on purpose you don't use the second one?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -106,11 +111,28 @@ void init() {
...
+            Instance<AmqpClientOptions> options = clientOptions.select(NamedLiteral.of(optionsName));
+            if (options.isUnsatisfied()) {
+                throw new IllegalStateException(
+                        "Cannot find a " + AmqpClientOptions.class.getName() + " bean named " + optionsName);
+            }
+            LOGGER.debug("Creating amqp client from bean named " + optionsName);
+            client = AmqpClient.create(new io.vertx.axle.core.Vertx(vertx.getDelegate()), options.get());
+        } else {
+            //TODO pick a default client options if there is one and use it, using a specific qualifier? like @DefaultAmqpConfiguration ?
+            //backwards compatibility
```
cescoffier: Good question, I would fall back to the current behavior, meaning ignore the options.
Maybe issue a warning.

examples/amqp-quickstart/src/main/resources/META-INF/microprofile-config.properties
famartinrh: oh, missed this, I was playing with it and just leaved it like this, I'll change it

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
famartinrh: do you mean to create an issue to discuss it?

cescoffier: Yes please do.

famartinrh: done, I also removed those comments so this PR could be ready to be merged

cescoffier:
Thanks!

PR #374, opened 2020-01-14
Allow delayed subscription to channel handled by emitters

cescoffier:
This is useful when the user wants to inject the `Publisher` attached to the emitter and managed the subscription directly.
So, the emitter must not checked whether it got a subscription during the initialization, but should thrown an IllegalStateException when an interaction is attempted (send, complete, fail...).

Also this PR improves the very bad error messages (like the unfamous 'stream is not connected')

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
@@ -133,7 +133,7 @@ public class EmitterImpl<T> implements Emitter<T> {
             String name) {
         FlowableEmitter<Message<? extends T>> emitter = reference.get();
         if (emitter == null) {
-            throw new IllegalStateException("No one subscribed to channel " + name);
+            throw new IllegalStateException("No subscriber found for the channel " + name);
         }
         if (emitter.isCancelled()) {
             throw new IllegalStateException("The subscription to " + name + " has been cancelled");
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
@@ -15,7 +15,7 @@ import io.smallrye.reactive.messaging.annotations.OnOverflow;
 
 /**
  * Implementation of the emitter pattern.
- * 
+ *
  * @param <T> the type of payload sent by the emitter.
  */
 public class EmitterImpl<T> implements Emitter<T> {
@@ -24,9 +24,10 @@ public class EmitterImpl<T> implements Emitter<T> {
     private final Flowable<Message<? extends T>> publisher;
 
     private static final Logger LOGGER = LoggerFactory.getLogger(EmitterImpl.class);
+    private final String name;
 
     EmitterImpl(String name, String overFlowStrategy, long bufferSize, long defaultBufferSize) {
-
+        this.name = name;
         if (defaultBufferSize <= 0) {
             throw new IllegalArgumentException("The default buffer size must be strictly positive");
         }
@@ -81,7 +82,7 @@ public class EmitterImpl<T> implements Emitter<T> {
 
     /**
      * Creates the stream when using the default buffer size.
-     * 
+     *
      * @param name the name of the emitter
      * @param defaultBufferSize the default buffer size
      * @param stream the upstream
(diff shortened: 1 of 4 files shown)


cescoffier:
@Emily-Jiang this may have an impact on https://github.com/eclipse/microprofile-reactive-messaging/pull/87

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
```diff
@@ -123,33 +124,34 @@ boolean isConnected() {
...
 
     }
 
-    static <T> FlowableEmitter<Message<? extends T>> verify(AtomicReference<FlowableEmitter<Message<? extends T>>> reference) {
+    static <T> FlowableEmitter<Message<? extends T>> verify(AtomicReference<FlowableEmitter<Message<? extends T>>> reference,
+            String name) {
         FlowableEmitter<Message<? extends T>> emitter = reference.get();
         if (emitter == null) {
-            throw new IllegalStateException("Stream not yet connected");
+            throw new IllegalStateException("No one subscribed to channel " + name);
```
kenfinnigan: "No one" or "Nothing"?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java
```diff
@@ -110,9 +110,11 @@ void afterDeploymentValidation(@Observes AfterDeploymentValidation done, BeanMan
             for (InjectionPoint ip : emitterInjectionPoints) {
                 String name = ChannelProducer.getChannelName(ip);
                 EmitterImpl<?> emitter = (EmitterImpl<?>) registry.getEmitter(name);
-                if (!emitter.isConnected()) {
-                    done.addDeploymentProblem(
-                            new DeploymentException("No channel found for name: " + name + ", injection point: " + ip));
+                if (!emitter.isSubscribed()) {
+                    // Subscription may happen later, just print a warning.
+                    // Attempting an emission without being subscribed would result in an error.
+                    LOGGER.warn("No subscriber for channel {}  attached to the emitter {}.{}", name,
```
kenfinnigan: Warning or just a trace?

How important is it for developers/debugging to know nothing is subscribed on creation?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
Emily-Jiang: How about "No subscriber for the channel "?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
```diff
@@ -137,14 +141,16 @@ public void testWithNull() {
...
-    @Test(expected = DeploymentException.class)
+    @Test(expected = IllegalStateException.class)
     public void testWithMissingStream() {
-        installInitializeAndGet(BeanWithMissingStream.class);
+        // The error is only thrown when a message is emitted as the subscription can be delayed.
+        installInitializeAndGet(BeanWithMissingStream.class).emitter().send(Message.of("foo"));
     }
 
-    @Test(expected = DeploymentException.class)
+    @Test(expected = IllegalStateException.class)
```
Emily-Jiang: I see what you meant. This will impact spec and tck change because the validation is at runtime instead of startup time.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterImpl.java
cescoffier: Much better, let me change this!

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java
cescoffier: It can be important. In the previous version, it was failing the _deployment_, that's why I went for warning.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
cescoffier: Yes, That's what I meant. This is actually required for https://github.com/eclipse/microprofile-reactive-messaging/issues/3

PR #378, opened 2020-01-17
Improve logging and error handling in the kafka connector

### LINKED ISSUE #372: How to handle errors outside of @Incoming and @Outgoing methods?
**Set-up:**

Kafka custom deserializer
Consumer method annotated with @Incoming
Messages in Kafka topic that cannot be deserialized (e.g. due to error or schema change).

**Expected behavior**

A deserializing exception should be at least logged.
An exception should trigger onErrorResume.

**Actual behavior**

An exception is silently ignored; further message processing stops.

**To reproduce**

1. Create an outgoing channel that writes non-string byte[]. 

       @Incoming("asc-processing-status-v1-in")
       Subscriber<KafkaMessage<String, ProcessingStatus>> gotUpdate() {

            return ReactiveStreams.<KafkaMessage<String, ProcessingStatus>>builder()
                    .onErrorResume(throwable -> {
                        log.warn("Cannot get value" + throwable);
                        return KafkaMessage.of(
                                "unknown",
                                new ProcessingStatus("unknown", Status.FAILURE, "unknown", throwable.getMessage())
                        );
                    })
                    .filter(message -> !message.getKey().equals("unknown"))
                    .forEach(message -> {
                        System.out.println("ShiftSocketApi.gotUpdate");
                        send(message.getPayload());
                    })
                    .build();
        }

1. Create an incoming channel that tries to read String.
1. Send a message.

**Environment**

* Quarkus 1.1.1.Final with quarkus-smallrye-reactive-messaging-kafka plugin
* java version "1.8.0_231"

cescoffier:
Fix #372

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaSource.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaSource.java
@@ -60,7 +60,8 @@ public class KafkaSource<K, V> {
 
         Objects.requireNonNull(topic, "The topic must be set, or the name must be set");
 
-        Flowable<KafkaConsumerRecord<K, V>> flowable = consumer.toFlowable();
+        Flowable<KafkaConsumerRecord<K, V>> flowable = consumer.toFlowable()
+                .doOnError(t -> LOGGER.error("Unable to read a record from Kafka topic '{}'", topic, t));
 
         if (config.getOptionalValue("retry", Boolean.class).orElse(true)) {
             Integer max = config.getOptionalValue("retry-attempts", Integer.class).orElse(5);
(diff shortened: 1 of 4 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaSource.java
```diff
@@ -60,14 +60,17 @@
...
+                .doOnError(t -> LOGGER.error("Unable to read a record from Kafka topic '{}'", topic, t));
 
         if (config.getOptionalValue("retry", Boolean.class).orElse(true)) {
             Integer max = config.getOptionalValue("retry-attempts", Integer.class).orElse(5);
             flowable = flowable
-                    .retryWhen(attempts -> attempts
-                            .zipWith(Flowable.range(1, max), (n, i) -> i)
-                            .flatMap(i -> Flowable.timer(i, TimeUnit.SECONDS)));
+                    .retryWhen(attempts -> {
+                        return attempts
```
kenfinnigan: Not huge, but is the use of `return` needed, appears to be a single line statement? Is the code change actually changing anything?

PR #391, opened 2020-01-24
Change connector name to smallrye-vertx-eventbus

boosey:
The connector name was coded incorrectly as smallrye.vertx.eventbus. The correct name is smallrye-vertx-eventbus.


cescoffier:
Amazing! Thanks!

Issue #393, opened 2020-01-26
Examples do not work for me


misl:
My Environment:
```
.../smallrye-reactive-messaging/examples/amqp-quickstart$ mvn -version
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /opt/apache-maven-3.6.3
Java version: 11.0.5, vendor: Private Build, runtime: /usr/lib/jvm/java-11-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.0.0-38-generic", arch: "amd64", family: "unix"
```
Seems like none of the examples will start from the command-line. 

| Example | Issue |
|-------------|----------|
| quickstart |  No mvn exec:java; java -jar also fails |
| amqp-quickstart | mvn exec:java --> java.lang.ClassNotFoundException: org.slf4j.LoggerFactory |
| awssns-quickstart | mvn exec:java --> java.lang.ClassNotFoundException: org.slf4j.LoggerFactory  |
| cloud-events | No mvn exec:java; java -jar also fails |
| gcp-pubsub-quickstart | mvn exec:java --> java.lang.ClassNotFoundException: org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder |
| kafka-quickstart | mvn exec:java --> java.lang.ClassNotFoundException: org.slf4j.LoggerFactory |
| kafka-quickstart-kotlin | mvn exec:java --> java.lang.ClassNotFoundException: org.slf4j.LoggerFactory |
| mqtt-quickstart | mvn exec:java --> java.lang.ClassNotFoundException: org.slf4j.LoggerFactory |
| mqtt-server-quickstart | No mvn exec:java; jaa -jar also fails |

**What I did:**
- Build the whole project (`mvn clean install`)
- Went through every single example to try 'mvn exec:java'

Can someone please verify this, just to make sure it is not local system that is broken.

kenfinnigan:
Is this issue fixed now?

misl:
For the most part, but I believe AWS and GCP quickstart are still failing.

cescoffier:
I'm starting thinking that the mono-repo approach is not the best. I'm not an expert in AWS and GCP meaning that I cannot really help. It might be better to move them in separate repository (under smallrye of course). @kenfinnigan would this be possible?

kenfinnigan:
So moving the AWS and GCP examples into a separate SmallRye repo?

[...]

PR #394, opened 2020-01-26
#393: Fixed examples/quickstart

misl:
Fixed mvn exec:java for quickstart
- Added README.md 
- Aligned package and classname with other quickstarts


misl:
- AWSSNS: can not be tested without AWS account
- GCP-PubSub: runtime failure after applying similar changes to pom file.

No changes have been committed for these 2 quickstart projects.

PR #398, opened 2020-01-30
Externalize the in-memory connector

### LINKED ISSUE #324: Externalize the in-memory connector to its own module

cescoffier:
Fix #324

* Move the in-memory connector to its own module
* Update doc
* Small refactoring of the tests

### DIFF
--- a/smallrye-reactive-messaging-in-memory/src/test/java/io/smallrye/reactive/messaging/connectors/WeldTestBase.java
+++ b/smallrye-reactive-messaging-in-memory/src/test/java/io/smallrye/reactive/messaging/connectors/WeldTestBase.java
@@ -1,4 +1,4 @@
-package io.smallrye.reactive.messaging;
+package io.smallrye.reactive.messaging.connectors;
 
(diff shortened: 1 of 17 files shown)


smallrye-reactive-messaging-in-memory/src/test/java/io/smallrye/reactive/messaging/connectors/WeldTestBase.java
```diff
@@ -0,0 +1,154 @@
+package io.smallrye.reactive.messaging.connectors;
```
kenfinnigan: Is this class somewhere it doesn't get pulled in as a dependency? Do we need a common test dependency?

cescoffier: At some point, we may want to externalize some of the duplicated test classes somewhere. I also need to check if I cannot use Weld unit directly.

PR #399, opened 2020-01-31
disable java 13 builds as the jdk cannot be downloaded at the moment

cescoffier:
This would need to be reverted once fixed


cescoffier:
Ok, it seems that we won't have to merge this. The issue came from the Adopt OpenJDK API and it has been fixed.

PR #404, opened 2020-02-04
For preliminary review: Message implementation by the framework - first drop

pgrabarsky:
The aim here is to have Message implemented by the framework implementation provider only (e.g. not by connector providers):

Introduction of MessageBuilder with ServiceLoader
Message interface cleanup
The provider module compiles and UTs are successul
The kafka module is no more subclassing Message, and UT are successul
Other modules still need to be reworked


api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/MessageBuilderProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+public abstract class MessageBuilderProvider {
+
+    private static volatile MessageBuilderProvider instance = null;
+
+    public abstract <T> MessageBuilder<T> newBuilder();
+
+    public static MessageBuilderProvider instance() {
+        if (instance == null) {
+            synchronized (ConfigProviderResolver.class) {
```
kenfinnigan: I'm presuming copy/paste error, that it should be synchronized on `MessageBuilderProvider`?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -82,7 +82,14 @@ public KafkaSource(Vertx vertx, Config config, String servers) {
                             // The Kafka subscription must happen on the subscription.
                             this.consumer.subscribe(topic);
                         }))
-                .map(rec -> new IncomingKafkaRecord<>(consumer, rec));
+                .map(rec -> {
+                        // TODO: the actual commit strategy should be implemented
+                        return Message.newBuilder()
+                            .payload(rec.value())
+                            .metadata(new IncomingKafkaRecordMetadata<>(rec))
```
kenfinnigan: If we move to builders for `Message`, should we also create builders for `Metadata`? Or is that specific to connectors so ok?

pgrabarsky:
> Overall it looks like the API usage is now a lot more verbose than it was before.
> 
> It makes me wonder whether it's possible to have both. The Builder API for those that want more flexibility at the expense of code verbosity, and those that want the simpler API? Could `KafkaRecord` use the builder underneath?

Hi @kenfinnigan 

Thanks for the comment, here is my answer:

It would be possible to add some shortcut like `Message::fromPayload` that to could replace sometimes this kind of construct: `payload -> Message.newBuilder().payload(payload).build()`. My feeling is that there is nothing really against it. The question is more about whether this construct is really frequent enough to do it. 

For `Messages` with special metadata, the overall compacity of expression could be maintained by having connector provider implementing some helpers functions, for instance:

```
class KafkaMessageHelper {
    public <K,T> Message<T> newIncoming(String topic, K key, T value, long timestamp, int partition) {
        IncomingKafkaMetadata ikmd = ...
        return Message.<T>newBuilder().payload(value).metadata(ikmd).build();
    }
}
```

I would say this is mostly internal connector provider concerne as they should be the only one to create an actually new `Message` only from their metadata.

From framework end-user perspective:

* He should not use it very often: In several cases, it will simply not create a new `Message` but only update the payload.

* When he will create a new message from an existing one (e.g. for child message semantic), this kind of builder is likely to be natural (however, the API for this precise operation still needs to be defined).

* When he will create a `Message` from scratch, it might make sense to have this kind of builder API, a bit more verbose. For instance to compare to the current more compact syntax you can find `KafkaRecord.of(K key, T value)` and `Message.of(T payload, Metadata metadata) ` with `KafkaRecord` being a `Message`. It seems this could create confusion.

So, I would say, maybe but nothing really better is obvious to me.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/spi/MessageBuilderProvider.java
pgrabarsky: Yes thanks! I'll fix it.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pgrabarsky: This would make sense when there is lot of parameters involved. I didn't want to refactor the metadata in this PR. So the current status is that `OutgoingKafkaRecordMetadata` has a builder but not `IncomingKafkaRecordMetadata`. Maybe it is because the end user is more likely to create an `OutgoingKafkaRecordMetadata` (?).

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -35,217 +35,6 @@
  */
 public interface Message<T> {
```
cescoffier: Ok, this definitely need more discussion because it's it completely derailing with the spec. I doubt we can pass the TCK with such changes.

cescoffier:
We would need to find a way to be not so incompatible with the spec.
Maybe we can add a SmallRye `Messages` class like:

```
public class Messages {
     MessageBuilder<T> newMessage();
}
```

The user would be able to use this directly.

kenfinnigan:
I've had another look through the changes and had some additional thoughts.

I agree with @cescoffier that we shouldn't be breaking the spec API if we can avoid it. Can you add back the methods on the `Message` class in the copied MP Reactive Messaging spec API, and move the `MessageBuilder` into a specific SmallRye API package? We should only be including classes into the `org.eclipse.microprofile` namespace if it's already part of the specification.

As @cescoffier suggested we could add `Messages` into the implementation API package and expose the builder with a method there. Such a class could then also be used for identifying the required acknowledgment coordination between messages.

I think for now we should remove the `MessageBuilderProvider` spi, as it's a complication I'm not sure we need from the beginning.

pgrabarsky:
I have moved to MessageBuilder to a smallrye package.

For the `Messages.newMessage()`, I'm not comfortable to have a `newMessage()` returning a builder. This looks to far to any convention.

Regarding the coordination of acknowledgement into the builder, I tend to agree as it was my initial proposal. This is prototyped here: https://github.com/AmadeusITGroup/odyssey-reactive-messaging/blob/master/reactive-messaging/src/main/java/com/amadeus/middleware/odyssey/reactive/messaging/core/impl/AbstractMessageBuilder.java#L62. However the Message acknowledgement API is different to enable the scenario described in the project README.md.

For the rest, I'm not getting the point. The purpose of this PR is to separate the API from the implementation for `Message`.

cescoffier:
Closing - it's been implemented in `master`.

PR #410, opened 2020-02-10
AMQP connector: configuration via global AmqpClientOptions bean

famartinrh:
This PR adds on top of the feature to configure the amqp connector via CDI using AmqpClientOptions beans. 
With this PR will be possible to configure the amqp clients for all amqp channels at once specifiying one env var `amqp-client-options-name` which have to have the name of an AmqpClientOptions bean.
This change is backwards compatible and the usage of this new env var is not mandatory, however if used must contain the name of an existing AmqpClientOptions bean (which makes total sense)

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -55,6 +55,10 @@ public class AmqpConnector implements IncomingConnectorFactory, OutgoingConnecto
     private Instance<AmqpClientOptions> clientOptions;
 
     @Inject
+    @ConfigProperty(name = "amqp-client-options-name")
+    private Optional<String> defaultClientOptionsName;
+
+    @Inject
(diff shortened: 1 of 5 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -114,21 +118,26 @@ private AmqpClient createClient(Config config) {
...
         return client;
     }
 
+    private AmqpClient createClientFromClientOptionsBean(String optionsBeanName) {
+        Instance<AmqpClientOptions> options = clientOptions.select(NamedLiteral.of(optionsBeanName));
+        if (options.isUnsatisfied()) {
+            throw new IllegalStateException(
+                    "Cannot find a " + AmqpClientOptions.class.getName() + " bean named " + optionsBeanName);
+        }
+        LOGGER.debug("Creating amqp client from bean named '{}'", optionsBeanName);
```
cescoffier: Wondering about the case of `amqp`? But you know better than me on this.

PR #415, opened 2020-02-11
Change default for retry-attempts to infinite retries

thescouser89:
If not specified by the user, the default retry behaviour is to always
retry, with an exponential backoff wait time, up to the retry max wait
time (30 seconds by default).

If the user specify a retry-attempts, then the nubmer of retries will be
limited to that number.

A new env var is also introduced, 'retry-max-wait'. It is used to set
the max wait time in seconds between retries, with a default of 30
seconds.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
@@ -6,6 +6,7 @@ import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.serialization.StringDeserializer;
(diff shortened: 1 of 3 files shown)


kenfinnigan:
Wouldn't this result in possible overconsumption of resources for attempting a call continually that won't succeed?

How would anyone know that something is broken?

thescouser89:
Hi Ken! My usecase is from the server standpoint. I've been using the kafka consumer part from Quarkus and the Kafka server have issues from time to time. So after just 5 times, the consumer just stops retrying to connect and do nothing. This is problematic because I would expect the consumer to always try to reconnect to the server in case of failure (think of the example of a database connection?)

I'm unsure how always retrying would create an overconsumption of resources here. I believe it would just be a retry every 10 seconds?

I'm new to all this so please be patient as I try to understand it more :)

kenfinnigan:
Well, I could be way off base too. I'm not an expert either.

Why not just set the configuration for a different number of retry attempts?

thescouser89:
This is what I'm doing right now (100 million haa). But assuming that the server runs forever, we'll hit that number of retries eventually, and the consumer will hang again. That's why I wanted it to be infinite.

thescouser89:
Any suggestions on what I could change / improve to get this PR merged?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -65,11 +65,17 @@ public KafkaSource(Vertx vertx, Config config, String servers) {
                 .doOnError(t -> LOGGER.error("Unable to read a record from Kafka topic '{}'", topic, t));
 
         if (config.getOptionalValue("retry", Boolean.class).orElse(true)) {
-            Integer max = config.getOptionalValue("retry-attempts", Integer.class).orElse(5);
-            flowable = flowable
+            Integer max = config.getOptionalValue("retry-attempts", Integer.class).orElse(null);
```
cescoffier: Minor - I generally use -1 for infinite, but `null` is ok.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -65,11 +65,17 @@ public KafkaSource(Vertx vertx, Config config, String servers) {
...
+
+            if (max == null) {
+                // always retry
+                flowable = flowable.retryWhen(attempts -> attempts.flatMap(i -> Flowable.timer(10, TimeUnit.SECONDS)));
```
cescoffier: Wondering if we should implement some kind of exponential backoff. WDYT?

thescouser89: Exponential backoff sounds good! Do you think it makes sense to backing off up to a max level?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
thescouser89: Noted! Will change it to -1

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
cescoffier: Yes, something like 30 seconds max between attempt.

thescouser89:
Hi @cescoffier ! I've pushed my changes based on your suggestions. If that doesn't look too bad, could you please also point me to the docs repo so that I can specify the new defaults?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -65,11 +65,30 @@ public KafkaSource(Vertx vertx, Config config, String servers) {
...
-            flowable = flowable
-                    .retryWhen(attempts -> attempts
-                            .zipWith(Flowable.range(1, max), (n, i) -> i)
-                            .flatMap(i -> Flowable.timer(i, TimeUnit.SECONDS)));
+            int max = config.getOptionalValue("retry-attempts", Integer.class).orElse(-1);
+            int retryMaxWait = config.getOptionalValue("retry-max-wait", Integer.class).orElse(30);
+
+            if (max == -1) {
+                // always retry
+                final int[] CURRENT_WAIT_SECOND = { 1 };
```
cescoffier: Should be an AtomicInteger.

thescouser89:
Ah I found out where to edit in the doc. Changed to `AtomicInteger` and updated the doc with new env var and defaults. Thanks for the review! :)

PR #442, opened 2020-03-09
check build again Java 14-EA and Java 15-EA

cescoffier:


cescoffier:
Rest-assured (because of the dependency on Groovy) does not support 14 and 15.

PR #446, opened 2020-03-10
Move documentation to Antora

jmartisk:
This is an example how we can make this work with the new infrastructure
@cescoffier please have a look, for example I'm not sure if the solution of copying the example source files is a good idea, but Antora does not support including files from outside its structure (and I even tried creating a symlink from antora structure to the directory containing the examples, but that didn't work either)


doc/antora-playbook.yml
```diff
@@ -0,0 +1,20 @@
...
+content:
+  sources:
+    - url: ..
+      start_path: doc
+      branches: HEAD
+ui:
+  bundle:
+    url: https://github.com/smallrye/smallrye-antora-ui/blob/master/build/ui-bundle.zip?raw=true
+    snapshot: true
+asciidoc:
```
jmartisk: unfortunately we have to include variables in the playbook for now, Antora 2.2 doesn't support them in component descriptor (`antora.yml`) so the same set of variables will need to be for the main playbook (within `smallrye.github.io` repository) for the time being

cescoffier:
I need to have a deeper look before merging. Typically, I may need to have both documentation for a bit (or at least a redirect). Also how it is going to be released it something important.

cescoffier:
@jmartisk You moved the snippets but are they compile?  Before they were in their own module to be compiled so if I was changing something I was breaking the build.

jmartisk:
If you need to keep the original documentation as well, no problem I can make a copy instead of moving it, but it will be a bit of a pain to maintain it.

> You moved the snippets but are they compile? Before they were in their own module to be compiled so if I was changing something I was breaking the build.

Well, the copies are not part of any Maven project now, I just copied out the files :( To get them compiled, I think we'd have to move the whole Maven project into the Antora structure..

cescoffier:
@jmartisk I'm reworking the REactive Messaging documentation completely. So, I'm going to include this.

jmartisk:
@cescoffier Ok, awesome. I suppose that means this can be closed. I will keep the `docs-test` branch in my fork so you can use it as reference.

PR #456, opened 2020-03-20
New documentation using Antora and @ConnectorAttribute

cescoffier:
This PR is multiple goals:

1. It provides support for `@ConnectorAttribute` that allows connectors to describe their configuration attributes. The HTTP and AMQP connectors are now using this approach. Other connectors will follow.
2. It provides a processor that reads the `@ConnectorAttribute` and generates configuration classes. It avoids a lot of boilerplate (and error-prone) code. Also, it homogenizes the error messages.
3. The processor also generates `adoc` files containing tables listing the attribute. These files are intended to be used in the documentation. 
4. This PR also switches to Antora instead of _raw_ asciidoc.

About 4., the support is a bit rough at the moment. 
In the documentation module, to build the documentation, you need to run:
```
mvn clean process-resources
cd target/antora
git init && git add -A && git commit -m "initialize a dummy repository"
antora generate antora-playbook.yml --clean
open build/site/index.html
```


cescoffier:
@jmartisk Can you check the antora part (in the documentation module). It's the first time I use Antora so... it's probably all wrong! Hint to improve the inclusion of "examples" and relative files is more than welcome.

documentation/src/main/antora/antora-playbook.yml
```diff
@@ -0,0 +1,18 @@
+# partial playbook just for this project to be able to compile documentation without pushing it to GitHub
+# usage: antora generate antora-playbook.yml
+site:
+    title: SmallRye Reactive Messaging Documentation
+    start_page: smallrye-reactive-messaging::index.adoc
+content:
+    sources:
+        - url: ./
```
jmartisk: Antora needs to run from a git repo, but if you want to spare yourself the need to do `git init` you can change this to

```
content:
    sources:
        - url: ../../..
          start_path: documentation/target/antora
          branches: HEAD                    
``` 

and it will reuse the existing git repo

jmartisk:
Hmm the main problem I see here is that you need to run a Maven build and include its results in the git repo (in particular, the directory `documentation/target/antora`, which, to make it worse, is currently `.gitignore`'d because it contains `target`). Ideally, the contents of `documentation/target/antora` would not be generated, but static. 

Is there a particular reason why the `documentation` module needs to be a Maven project and not a static directory that is compliant with Antora out of the box? I would guess you'd maybe want it to be a Maven project for running tests or so, but there are currently no runnable tests in it AFAICS

cescoffier:
That is very unfortunate. The documentation is containing code and filtering. 
So basically what you are saying it that I need to not use `target` but a repository from the source tree.

jmartisk:
I'm basically saying that the `antora` directory should optimally not be generated by Maven, but static.

The filtering is just to fill values of the `*-version` properties or does it do something else?
Including properties only works in the "local" `antora-playbook.yml` file for now, for generating the live website a different playbook will be used (https://github.com/smallrye/smallrye.github.io/blob/develop/docs/antora-playbook.yml), Antora 2.2 does not yet support declaring properties in the module descriptor (`antora.yml`) so we'd have to manage the versions in the main smallrye.github.io repository anyway. Unless we try updating to using Antora 2.3 Alpha - I think that should already support including properties in module descriptors. I could try experimenting with it next week to see if it does not break anything (it is only an Alpha for now...)

Anyway that still would not solve the problem that the descriptor would be generated by a Maven build. Do you consider it too much of a burden to update the versions in the descriptor/playbook manually (when they change) rather than through Maven filtering?

jmartisk:
If we turned the directory into a static directory but still wanted to be able to run a Maven build on it (for example, to check that it compiles), I suppose we could introduce a Maven module in a separate directory that would be "joined" into the `antora` directory through symbolic links (its `src/main/java` would be a symlink pointing at `antora/modules/ROOT/examples` etc.) I hope Maven would be able to handle such symlinks (Antora unfortunately cannot)

cescoffier:
My docs are not totally static (and this would be the same for mutiny and so on).
Without filtering:

1. you can be sure that the doc and the project are going to use different dependency versions
2. how can you automate release as the version is in the `pom.xml` file - so need filtering.

I can, in the worse case, having an output directory with the output from maven that would be updated every time someone edits the doc. This is kind of cumbersome, but I extend my maven build to somehow detect this and fail the build.

I would also need to customize the "edit this page" to use the right one.

Maybe I can reduce this to only filter the `antora.yaml` file, and keep the rest of the doc there.
It would still be a problem for the example inclusion.

cescoffier:
@kenfinnigan please read the antora discussion. I'm not totally sure it would work for the reactive projects.

cescoffier:
I don't totally understand the symbolic link approach. I would prefer avoiding having to deal with symbolic link as that would mean the project cannot be built on windows.

cescoffier:
@jmartisk what does antora need to aggregate the doc? The HTML output or the sources?

cescoffier:
@jmartisk Ok, I've read the Antora doc last night and I start to see some light.

The content is now in `src/main/doc`. That's what your aggregator will consume. 
The only filtered file in the `playbook` and I believe you have an aggregator one. 
BTW, I don't know how we will keep the variable from this one up to date.

I'm still working on the migration. Wil take a bit of time.

jmartisk:
> I don't totally understand the symbolic link approach. I would prefer avoiding having to deal with symbolic link as that would mean the project cannot be built on windows.

That'a fair point, I forgot about windows, that would probably break

> @jmartisk what does antora need to aggregate the doc? The HTML output or the sources?

No, it does not need HTML output. The repo (~what you push to github) needs to contain the `antora.yml` descriptor and the `modules` directory containing all the assets. And all this should basically be static files. If that's not doable, we could potentially include some sort of custom building logic for your repo specifically (the Semaphore job for deploying the `smallrye.io` site would clone your repo, execute some build steps, and then generate the documentation from the local clone instead of using the github upstream directly). It would not be very nice, but it would be doable.


> The only filtered file in the `playbook` and I believe you have an aggregator one.
> BTW, I don't know how we will keep the variable from this one up to date.

I'll see if we can go and upgrade to antora 2.3 Alpha, because with 2.2 we need to keep variables in the playbooks, which is very much stupid, because the master playbook that will build your documentation is in a different repo, so that's very hard to update by a script.. I'll keep you updated on my findings

documentation/src/main/doc/antora.yml
```diff
@@ -0,0 +1,5 @@
+name: smallrye-reactive-messaging
+title: SmallRye Reactive Messaging
+version: ${project.version}
```
jmartisk: Does this need to be a variable? I mean, even if your `project.version` is for example `1.3.5-SNAPSHOT`, the "documentation version" will probably be 1.3 (we will not produce a different doc version for each micro release or so)
What I was expecting to do with this is that we simply put in the `major.minor` version here and update it manually once in a while (which will basically be only when you create a new branch targeting a different minor version, so not very often)

cescoffier: I caught it yesterday. I missed it.

The issue is that it avoids the release process automation without adding a commit in the process. Adding a commit is not impossible, just tricky as I will rely on regexp and replacement. 

Note that most of the time we are releasing minor, so it would require an increment every time (more or less).

cescoffier:
@jmartisk @kenfinnigan I've done a few more experiments over the weekend. I will migrate the documentation of reactive messaging 2.0 to Antora (but kept it separate from the rest of smallrye as it is Today until all projects migrate). At least I can give a few guidelines about how to get it working with Maven and structure the doc.

jmartisk:
Ok so I tried locally updating to Antora `2.3.0-beta.1` which was just released this weekend and it looks good, the main goodie is that we can include variables in the `antora.yml` descriptors now. 

So basically if you really need to do some building/filtering steps before passing stuff to Antora, I guess you could generate the `antora` directory with a Maven build, and that directory would also include the `antora.yml` descriptor with the correct variables after filtering. The `antora-playbook.yml` that is used for "partial" build doesn't have to contain them anymore with 2.3.

The somewhat unfortunate thing is that we will have to update the Semaphore deployment job to run a Maven build in your repo to generate the `antora` directory, because we probably won't push that into GitHub (it's generated so it's probably not a good idea to push it?)

cescoffier:
@jmartisk cool! Thanks for the update. I will try to play around. Semaphore is not an issue as the build process will be migrated to Github Actions (following what I did on Mutiny).

So just to summarize, with this new version, the playbook file becomes much simpler (no variables) but the `antora.yml` file would contain the variables. I need to see how I can generate and keep updated this file at development time (bump of a dependency version) and during the release process (bump of the project version). Note that it makes dependabot useless for a few dependencies (SmallRye Config,. Mutiny...).

documentation/src/main/doc/antora.yml
cescoffier: I found a way to extract this in Maven.

jmartisk:
I noticed you're including the `ui-bundle.zip` in the repo, is that intentional? Any reason why it can't be taken from https://github.com/smallrye/smallrye-antora-ui?

cescoffier:
> I noticed you're including the ui-bundle.zip in the repo, is that intentional? Any reason why it can't be taken from https://github.com/smallrye/smallrye-antora-ui?

Yes, I needed different menu items and footer.

jmartisk:
Well, but once we build the docs for the main `smallrye.github.io` site, it will use the "official" bundle. If you need these changes to be visible on the main site, we will need to apply them in the main bundle definition.  I don't know what exactly you changed, but given the customizability of the UI, it shouldn't be a big deal I think?

cescoffier:
The menu you added in the bundle list all the SamllRye projects. It makes sense when you have multiple projects. 

I also added a link to the Github repo in the footer.

jmartisk:
Oh, I see now, yes that makes sense. Perhaps this could be reused by other projects to generate their own specific site, so we could consider maintaining the source as an alternative branch in https://github.com/smallrye/smallrye-antora-ui . I suppose you'll want to upload the source somewhere anyway.

cescoffier:
Yes, I will upload the source.

I’m looking into Antora UI to see if that case be handled directly in your template.

PR #467, opened 2020-03-26
Fixes #79 - Add support for @Blocking

### LINKED ISSUE #79: Add support for @Blocking
`@Blocking` indicates that the method is blocking and must be called on a worker thread.

The `@Blocking` annotation can indicate the max level of concurrency (`concurrency` attribute, defaulting to 1.

kenfinnigan:


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/ExecutionHolder.java
```diff
@@ -0,0 +1,43 @@
...
+        if (internalVertxInstance) {
+            vertx.close().await().indefinitely();
+        }
+    }
+
+    @PostConstruct
+    public void init() {
+        if (instanceOfVertx == null || instanceOfVertx.isUnsatisfied()) {
+            internalVertxInstance = true;
+            this.vertx = Vertx.vertx();
```
cescoffier: Maybe we should log a message.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+            @Observes(notifyObserver = Reception.IF_EXISTS) @Priority(100) @BeforeDestroyed(ApplicationScoped.class) Object event) {
+        if (!workerExecutors.isEmpty()) {
+            for (WorkerExecutor executor : workerExecutors.values()) {
+                executor.close();
+            }
+        }
+    }
+
+    public <T> Uni<T> executeWork(Handler<Promise<T>> blockingCodeHandler, String workerName, boolean ordered) {
+        Objects.requireNonNull(blockingCodeHandler, "Code to execute not provided");
```
cescoffier: Action instead of code?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+
+        if (workerName == null) {
+            return executionHolder.vertx().executeBlocking(blockingCodeHandler, ordered);
+        } else {
+            return getWorker(workerName).executeBlocking(blockingCodeHandler, ordered);
+        }
+    }
+
+    private WorkerExecutor getWorker(String workerName) {
```
cescoffier: Need to be synchronized I guess.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/BlockingProcessorTest.java
```diff
@@ -0,0 +1,94 @@
...
+        MyCollector collector = container.select(MyCollector.class).get();
+        await().until(collector::hasCompleted);
+        assertThat(collector.payloads()).isEqualTo(LIST);
+
+        BeanConsumingMessagesAndProducingItems bean = container.select(BeanConsumingMessagesAndProducingItems.class).get();
+
+        List<String> threadNames = bean.threads().stream().distinct().collect(Collectors.toList());
+        assertThat(threadNames.contains(Thread.currentThread().getName())).isFalse();
+        for (String name : threadNames) {
+            assertThat(name.startsWith("vert.x-worker-thread-")).isTrue();
```
cescoffier: I'm not sure we need to be that precise.

api/src/main/java/io/smallrye/reactive/messaging/MediatorConfiguration.java
```diff
@@ -41,6 +41,12 @@
 
     int getNumberOfSubscriberBeforeConnecting();
 
+    boolean isBlocking();
+
+    String getWorkerPoolName();
+
+    boolean isOrderedExecution();
```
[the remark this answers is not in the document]
kenfinnigan: Latter?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
```diff
@@ -73,6 +80,21 @@ public void initialize(Object bean) {
         }
     }
 
+    protected <T> Uni<T> invokeBlocking(Object... args) {
+        try {
+            Objects.requireNonNull(this.invoker, "Invoker not initialized");
```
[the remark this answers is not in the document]
kenfinnigan: I will see if somewhere fits

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/BlockingProcessorTest.java
kenfinnigan: Well, it makes sure it's on the worker thread ;-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -39,26 +41,17 @@
     private List<KafkaSource> sources = new CopyOnWriteArrayList<>();
     private List<KafkaSink> sinks = new CopyOnWriteArrayList<>();
 
-    private boolean internalVertxInstance = false;
     private Vertx vertx;
 
-    public void terminate(@Observes @BeforeDestroyed(ApplicationScoped.class) Object event) {
+    public void terminate(
+            @Observes(notifyObserver = Reception.IF_EXISTS) @Priority(50) @BeforeDestroyed(ApplicationScoped.class) Object event) {
```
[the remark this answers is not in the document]
kenfinnigan: `@Observes(notifyObserver = Reception.IF_EXISTS)` means that an instance of this bean isn't created to call the `terminate()` method if one doesn't exist.

`@Priority` then defines the ordering, with lower numbers being triggered first

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/ExecutionHolder.java
kenfinnigan: That we're creating an instance? sure

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/MediatorConfigurationSupport.java
```diff
@@ -490,6 +490,29 @@ public Integer processBroadcast(Object outgoing, Supplier<Integer> supplier) {
...
+                || validationOutput.production.equals(MediatorConfiguration.Production.INDIVIDUAL_PAYLOAD)
+                || validationOutput.production.equals(MediatorConfiguration.Production.NONE))) {
+            throw getBlockingError("The @Blocking annotation is only supported for methods returning Message or a payload");
+        }
+
+        if (!(validationOutput.consumption.equals(MediatorConfiguration.Consumption.MESSAGE)
+                || validationOutput.consumption.equals(MediatorConfiguration.Consumption.PAYLOAD)
+                || validationOutput.consumption.equals(MediatorConfiguration.Consumption.NONE))) {
+            throw getBlockingError(
+                    "The @Blocking annotation is only supported for methods with parameters of Message or a payload");
```
cescoffier: Same comment.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/MediatorConfigurationSupport.java
```diff
@@ -490,6 +490,29 @@ public Integer processBroadcast(Object outgoing, Supplier<Integer> supplier) {
...
+        }
+
+        if (ClassUtils.isAssignable(returnType, CompletionStage.class)) {
```
cescoffier: You need to also check for Uni.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/SubscriberMediator.java
```diff
@@ -127,20 +127,36 @@ public void onComplete() {
...
-                })
-                .ignore();
+        if (configuration.isBlocking()) {
+            this.subscriber = ReactiveStreams.<Message<?>> builder()
+                    .flatMapCompletionStage(managePreProcessingAck())
+                    .flatMapCompletionStage(message -> ((Uni<?>) invokeBlocking(message.getPayload()))
+                            .subscribeAsCompletionStage()
+                            .thenApply(x -> message))
+                    .flatMapCompletionStage(x -> {
+                        if (configuration.getAcknowledgment() == Acknowledgment.Strategy.POST_PROCESSING) {
```
cescoffier: We may want to refactor and extract this code as it's now duplicated a few times. It can be done in another PR.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherMediator.java
```diff
@@ -110,17 +110,33 @@ private void produceAPublisherOfMessages() {
...
-            Message message = invoke();
-            Objects.requireNonNull(message,
-                    "The method " + configuration.methodAsString() + " returned an invalid value: null");
-            return message;
-        }));
+        if (configuration.isBlocking()) {
+            setPublisher(ReactiveStreams.<Uni<?>> generate(this::invokeBlocking)
+                    .flatMapCompletionStage(Uni::subscribeAsCompletionStage)
+                    .map(message -> {
+                        Objects.requireNonNull(message,
```
cescoffier: This should have been tested before (I've added a comment about this case).

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -234,44 +234,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
-                    .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> (Message) invoke(input.getPayload()))
-                    .buildRs();
+            if (configuration.isBlocking()) {
+                this.processor = ReactiveStreams.<Message> builder()
+                        .flatMapCompletionStage(managePreProcessingAck())
+                        .<Message> flatMapCompletionStage(input -> ((Uni<?>) invokeBlocking(input.getPayload()))
+                                .subscribeAsCompletionStage()
+                                .thenApply(result -> (Message) result))
+                        .buildRs();
```
cescoffier: There is a bug here. It may have been there before.
POST_PROCESSING acknowledgment is not handled correctly (well not handled at all).

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -234,44 +234,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
+            } else {
+                this.processor = ReactiveStreams.<Message> builder()
+                        .flatMapCompletionStage(managePreProcessingAck())
+                        .map(input -> (Message) invoke(input.getPayload()))
```
cescoffier: Same here.
POST_PROCESSING acknowledgment is not handled.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -234,44 +234,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
-            this.processor = ReactiveStreams.<Message> builder()
-                    .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> (Message) invoke(input))
-                    .buildRs();
+            if (configuration.isBlocking()) {
+                this.processor = ReactiveStreams.<Message> builder()
+                        .flatMapCompletionStage(managePreProcessingAck())
+                        .<Message> flatMapCompletionStage(input -> ((Uni<?>) invokeBlocking(input))
+                                .subscribeAsCompletionStage()
+                                .thenApply(result -> (Message) result))
```
cescoffier: POST_PROCESSING acknowledgment is not handled correctly

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -234,44 +234,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
+                        .buildRs();
+            } else {
+                this.processor = ReactiveStreams.<Message> builder()
+                        .flatMapCompletionStage(managePreProcessingAck())
+                        .map(input -> (Message) invoke(input))
```
cescoffier: POST_PROCESSING acknowledgment is not handled correctly

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private void addWorker(Method method) {
+        Objects.requireNonNull(method, "Method was empty");
+
+        Blocking blocking = method.getAnnotation(Blocking.class);
+
+        // Validate @Blocking is used in conjunction with @Incomings, @Incoming, or @Outgoing
+        if (!(method.isAnnotationPresent(Incomings.class) || method.isAnnotationPresent(Incoming.class)
+                || method.isAnnotationPresent(Outgoing.class))) {
+            throw getBlockingError(method, "no @Incomings, @Incoming, or @Outgoing present");
```
cescoffier: I would remove the @Incomings

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            // Validate @Blocking worker pool has configuration to define concurrency
+            String workerConfigKey = WORKER_CONFIG_PREFIX + "." + blocking.value() + "." + WORKER_CONCURRENCY;
+            Optional<Integer> concurrency = configInstance.get().getOptionalValue(workerConfigKey, Integer.class);
+            if (!concurrency.isPresent()) {
+                throw getBlockingError(method, workerConfigKey + " was not defined");
+            }
+
+            workerDefinitions.put(blocking.value(), concurrency.get());
```
cescoffier: Maybe a log message (info) indicating the that the worker pool has been created with concurrency x

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    private IllegalArgumentException getBlockingError(Method method, String message) {
+        return new IllegalArgumentException(
```
cescoffier: Don't we have this method somewhere else?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/InvalidBlockingProcessorShapeTest.java
```diff
@@ -0,0 +1,253 @@
...
+        @Blocking
+        @Incoming("count")
+        @Outgoing("sink")
+        public CompletionStage<Message<String>> process(Message<Integer> value) {
+            return CompletableFuture.supplyAsync(() -> Integer.toString(value.getPayload() + 1))
+                    .thenApply(Message::of);
+        }
+    }
+
+    @Test(expected = DeploymentException.class)
```
cescoffier: Need to also test when the method returns an Uni

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/InvalidBlockingProcessorShapeTest.java
```diff
@@ -0,0 +1,253 @@
...
+        addBeanClass(BeanProducingAPublisherBuilderOfMessagesAndConsumingIndividualMessage.class);
+        initialize();
+    }
+
+    @ApplicationScoped
+    public static class BeanProducingAPublisherBuilderOfMessagesAndConsumingIndividualMessage {
+        @Blocking
+        @Incoming("count")
+        @Outgoing("sink")
+        public PublisherBuilder<Message<String>> process(Message<Integer> message) {
```
cescoffier: I would add specific tests for Multi.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/InvalidBlockingPublisherShapeTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Test(expected = DeploymentException.class)
+    public void testProduceCompletionStageOfPayload() {
+        addBeanClass(BeanReturningCompletionStageOfPayload.class);
+        initialize();
+    }
+
+    @ApplicationScoped
+    public static class BeanReturningCompletionStageOfPayload {
```
cescoffier: Would need tests with Unis

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/InvalidBlockingSubscriberShapeTest.java
```diff
@@ -0,0 +1,179 @@
...
+        @Blocking
+        @Incoming("subscriber")
+        public Subscriber<Message<String>> create() {
+            return ReactiveStreams.<Message<String>> builder()
+                    .ignore().build();
+        }
+    }
+
+    @Test(expected = DeploymentException.class)
+    public void testThatWeCanConsumeMessagesFromAMethodReturningACompletionStage() {
```
cescoffier: Need the same with Uni.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/beans/BeanReturningPayloads.java
```diff
@@ -0,0 +1,28 @@
...
+@ApplicationScoped
+public class BeanReturningPayloads {
+    private AtomicInteger count = new AtomicInteger();
+    private List<String> threads = new CopyOnWriteArrayList<>();
+
+    @Blocking
+    @Outgoing("infinite-producer")
+    public int create() {
+        threads.add(Thread.currentThread().getName());
+        return count.incrementAndGet();
```
cescoffier: I would actually do something really blocking to check the concurrency.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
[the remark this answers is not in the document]
cescoffier: So it makes the hypothesis that the processing is synchronous/blocking? Or priority cannot be enforced right?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -98,7 +98,7 @@ public KafkaSource(Vertx vertx, Config config, String servers) {
 
     public void closeQuietly() {
         try {
-            this.consumer.closeAndAwait();
+            this.consumer.closeAndForget();
```
cescoffier: With the priority thingy this should not be required anymore.

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
```diff
@@ -144,11 +144,7 @@ private static MqttServerOptions mqttServerOptions(Config config) {
     }
 
     synchronized void close() {
-        mqttServer.close()
-                .onFailure().invoke(t -> logger.warn("An exception has been caught while closing the MQTT server", t))
-                .onItem().invoke(x -> logger.debug("MQTT server closed"))
-                .onFailure().recoverWithItem((Void) null)
-                .await().indefinitely();
+        mqttServer.closeAndForget();
```
cescoffier: With the @Priority, this should not be required anymore.

api/src/main/java/io/smallrye/reactive/messaging/MediatorConfiguration.java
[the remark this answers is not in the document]
cescoffier: +1

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/BlockingProcessorTest.java
cescoffier: Ok, fair enough :-)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/ExecutionHolder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public void terminate(
+            @Observes(notifyObserver = Reception.IF_EXISTS) @Priority(200) @BeforeDestroyed(ApplicationScoped.class) Object event) {
+        if (internalVertxInstance) {
+            vertx.close().await().indefinitely();
+        }
+    }
+
+    @PostConstruct
+    public void init() {
```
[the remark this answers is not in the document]
mkouba: It does not have to be synchronized but the `vertx` field should be volatile. A better approach would be to make the  `vertx` field final and use constructor injection:
```java
private final Vertx vertx;

// Dummy no-args constructor needed for client proxies, not needed on quarkus ;-)
ExecutionHolder() {
}

@Inject
ExecutionHolder(Instance<Vertx> instanceOfVertx) {
  this.vertx = ...
}
```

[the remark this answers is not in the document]
cescoffier: Yes, that would be much better!

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
```diff
@@ -0,0 +1,33 @@
...
+     * Indicates the name of the worker pool to use for execution.
+     * By default all executions will be performed on the default worker pool.
+     *
+     * @return custom worker pool name for blocking execution.
+     */
+    String value() default NO_VALUE;
+
+    /**
+     * Indicates whether the execution on the worker pool should be ordered.
+     * By default all executions are ordered.
```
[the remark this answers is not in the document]
kenfinnigan: did you really mean "combustion" above?

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpTestBase.java
```diff
@@ -45,7 +46,8 @@ public void tearDown() throws InterruptedException {
 
         CountDownLatch latch = new CountDownLatch(1);
         usage.close();
-        vertx.close().subscribe().with(x -> latch.countDown(), f -> latch.countDown());
+        executionHolder.vertx().close().subscribe().with(x -> latch.countDown(), f -> latch.countDown());
```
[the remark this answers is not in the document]
kenfinnigan: I just retained the code that was there previously as well.

But if the `vertx().closeAndAwait()` inside `executionHolder.terminate()` is fine, I can remove the line

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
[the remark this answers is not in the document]
kenfinnigan: All observers are synchronous by default unless you specify an async observer.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
kenfinnigan: Do you mean revert to the original code or remove?

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
kenfinnigan: Same question here, remove the close or go back to original code?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
kenfinnigan: I was wondering that yesterday, as it seemed it was the only one without it. Should've asked then!

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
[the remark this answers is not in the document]
cescoffier: :-) ... computation.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/PublisherMediator.java
kenfinnigan: So remove the null check?

cescoffier: If you follow the other approach in `invokeBlocking` yes.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/ExecutionHolder.java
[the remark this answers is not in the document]
kenfinnigan: Set `Vertx` to null in the default constructor?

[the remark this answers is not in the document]
kenfinnigan: It also impacts tests that don't use CDI container, as I need to set the Vertx instance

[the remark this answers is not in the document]
mkouba: > Set Vertx to null in the default constructor?

I think so. And make sure it's package-private. It should be only used when instantiating a client proxy.

> It also impacts tests that don't use CDI container, as I need to set the Vertx instance.

In that case you can add one more constructor that accepts a `Vertx` instance directly.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
kenfinnigan: Both these methods are also in `MediatorConfigurationSupport`.

Similar methods for Incoming and Outgoing are also duplicated between that class and `DefaultMediatorConfiguration`

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
cescoffier: Can you open an issue?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -236,44 +245,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
+            }
         } else {
-            this.processor = ReactiveStreams.<Message<?>> builder()
-                    .flatMapCompletionStage(managePreProcessingAck())
-                    .map(input -> (Message<?>) invoke(input))
-                    .buildRs();
+            if (configuration.isBlocking()) {
+                this.processor = ReactiveStreams.<Message<?>> builder()
+                        .flatMapCompletionStage(managePreProcessingAck())
+                        .<Message<?>> flatMapCompletionStage(input -> ((Uni<?>) invokeBlocking(input))
```
cescoffier, line 265: We must check that the returned Uni is not `null`.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
```diff
@@ -236,44 +245,92 @@ private void processMethodReturningAPublisherOfPayloadsAndConsumingPayloads() {
...
+                                .subscribeAsCompletionStage()
+                                .thenApply(result -> (Message<?>) result))
```
cescoffier, line 267: we need to verify that `result` is not null

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Indicates the name of the worker pool to use for execution.
+     * By default all executions will be performed on the default worker pool.
+     *
+     * The maximum concurrency of a custom worker pool can be set with the following configuration key:
+     * <code>smallrye.messaging.worker.{pool-name}.max-concurrency</code>
+     *
+     * @return custom worker pool name for blocking execution.
+     */
+    String value() default DEFAULT_WORKER_POOL;
```
Emily-Jiang, line 30: For the value, is `.` allowed as part of the name?

kenfinnigan: Probably not, just like other configs like topic name

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/MediatorConfigurationSupport.java
kenfinnigan: I added it below

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
kenfinnigan: I was wrong, the way it's currently coded it can be any value

documentation/src/main/doc/modules/ROOT/pages/advanced/blocking.adoc
```diff
@@ -0,0 +1,50 @@
+== @Blocking
+
+The `@Blocking` annotation can be used on a method annotated with `@Incomings`, `@Incoming`,
```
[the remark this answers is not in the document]
kenfinnigan: Think that popped back through rebasing, sorry

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpSinkTest.java
```diff
@@ -33,7 +33,6 @@
     @After
     public void cleanup() {
         if (provider != null) {
```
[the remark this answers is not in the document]
kenfinnigan: The method became `terminate` -> https://github.com/smallrye/smallrye-reactive-messaging/pull/467/files#diff-3a4f486c21639cc4d8651925bb53e79dL338

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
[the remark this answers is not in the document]
kenfinnigan: So, I think we can move it to `initialize()` of each mediator after they've called `super.initialize()`, but it does mean duplicating that check across the mediators.

What do you think?

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
Emily-Jiang: Had a look at vertx.createSharedWorkerExecutor(). It seems there is no restriction on the name as far as javadoc is concerned.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/ProcessorMediator.java
kenfinnigan: https://github.com/smallrye/smallrye-reactive-messaging/issues/480

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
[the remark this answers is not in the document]
cescoffier: let's keep it there for now. We can revisit later.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
cescoffier: yes. closeAndAwait();

smallrye-reactive-messaging-mqtt-server/src/main/java/io/smallrye/reactive/messaging/mqtt/server/MqttServerSource.java
cescoffier: `closeAndAwait`

kenfinnigan: Should be the right code now

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
kenfinnigan: Should be the right code now

api/src/main/java/io/smallrye/reactive/messaging/annotations/Blocking.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Identifies that a particular method performs blocking operations,
+ * and as such should be executed on separate worker.
+ *
+ * When possible, the initial threads execution context is _restored_
```
hutchig, line 13: I think the term "initial threads" could be a bit confusing for some non-English speakers as the apostrophe is missing it is close to "the initiating thread's context is propogated". It noun phrase would be better as
"When possible, the context on the thread used to execute the method is restored to the context present prior to the method being executed"    (if you want that meaning). 
Or "When possible, the context present on the worker thread when the blocking method is run is set to match the context used to run non-blocking methods." (if you want that meaning)....and apologies for "bike shedding".

kenfinnigan: What about changing "initial" to "triggering"?

cescoffier: "When supported, the caller execution context is captured and restored after the execution of the blocking method. This means that the rest of the pipeline would run in the same context (same thread). Typically, this is the case if the caller is using a Vert.x Context (event loop or worker)."

cescoffier:
Woohoo! LGTM!

Waiting for CI!

Issue #329, opened 2019-11-25
Increase the concurrency in the Kafka connector


cescoffier:
In the Kafka connector, it should be possible to configure the number of Kafka consumers so the application can consume more messages.

The number of in-flight messages in the pipeline must also be improved. This requires changing how the acknowledgment is handled to enqueue them and commit several records in one batch.

gsmet:
@cescoffier if you prepare a rough design, maybe that could be a good issue for @fmhwong?

cescoffier:
It's not clear yet what's the best way to achieve this. The easiest would be to create multiple instances of Kafka consumers. However, I'm not very satisfied with this approach and I would prefer something a bit more fine-grain - without mentioning the ordering issue and useless resource consumption if you create more consumers than partitions.

overthetop:
@cescoffier @gsmet  A good approach is what has been done in Spring Boot Kafka. 
You set the concurrency for example to 10 and under the hood that translates to 10 native kafka consumers for that specific consumer group. This is exactly how you'll do it if you are using the native kafka client to consumer data. It's really up to the developer to set the right concurrecy because he is the one who knows what's the best way to consume from his topic (how slow the processing will be, how many partitions he wants to process in each consumer etc.). Another really handy thing will be to be able to inject the native KafkaConsumer object in the processing method, so that developers can have full control over the kafka client (be able to pause() and resume() for example). Of course that's another topic (that functionality is also availale in Spring boot kafka)

einarjohnson:
Hi, are there any plans on the roadmap to support concurrency when processing messages?

cescoffier:
You can already do that using @Blocking.

einarjohnson:
@cescoffier , thanks for the quick reply. another question, can I use the blocking annotation in conjunction with the smallrye fault tolerance @retry annotation?

cescoffier:
Yes.

PyAntony:
@cescoffier Hi. Can you please give an example (or post a link) of how to achieve this? I know how blocking works, but what would be the best way to achieve multiple consumers with @Blocking?

PR #497, opened 2020-04-12
Improve concurrency handling in the Kafka sink

cescoffier:
Instead of relying on acknowledgment, it does handle a proper request protocol based on the max number of messages that can wait for acknowledgment concurrently.


cescoffier:
Related to #329

PR #503, opened 2020-04-14
add lower-level worker pool registration method

michalszynkiewicz:


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -38,9 +37,6 @@
     @Inject
     private ExecutionHolder executionHolder;
 
-    @Inject
```
kenfinnigan, line 41: What's the reason for not injecting the Config and using the Provider?

I thought that was a source of issues?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
```diff
@@ -105,39 +101,45 @@ private WorkerExecutor getWorker(String workerName) {
...
-                "Invalid method annotated with @Blocking: " + methodAsString(method) + " - " + message);
+    private void defineWorker(Method method) {
+        Objects.requireNonNull(method, "Method was empty");
+
+        Blocking blocking = method.getAnnotation(Blocking.class);
+
+        String methodName = method.getName();
+        String className = method.getDeclaringClass().getName();
+
+        // Validate @Blocking is used in conjunction with @Incoming, or @Outgoing
```
kenfinnigan, line 133: Should this validation be in the other method so it doesn't need to be repeated in a runtime?

Not sure what that means for parameters though

michalszynkiewicz: I don't know. This check was done this way before my changes too.

kenfinnigan: Maybe pass an array of annotations on the method?

kenfinnigan: Not sure what you mean. Quarkus doesn't do this check?

My presumption was that Quarkus is calling the defineWorker version that doesn't have the check. If it's calling the one with the check, that's fine

michalszynkiewicz: oh, sorry, I misunderstood your question :)
It seems I missed it and Quarkus does not have this validation. I created an issue for it, will fix it tomorrow: https://github.com/quarkusio/quarkus/issues/8567

For Quarkus we don't have info on the present annotations here, it's taken from Jandex so has to be done before calling `defineWorker`

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
michalszynkiewicz: When I had this code in Quarkus, it caused problems, I will try to roll back to Instance<Config>

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/connectors/WorkerPoolRegistry.java
michalszynkiewicz: @kenfinnigan I checked Quarkus once again and there `@Blocking` on methods that don't have `@Incoming` or `@Outgoing` will be ignored. That's because the logic for blocking is applied only to methods from `MediatorBuildItem`s.

Anyway, such validation makes sense and I will add it to Quarkus.

PR #520, opened 2020-04-23
Do not swallow the main exception when reporting a synchronous initialisation failure

### LINKED ISSUE #515: Exception thrown from connector during startup discarded by provider with cause thrown instead
If an exception occurs when a connector creates its subscriber or producer, if that exception has a cause, the exception itself is discarded and the cause is reported as a deployment error.

This happens here: https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/ReactiveMessagingExtension.java#L135

This was unhelpful in my connector, because the exception that was discarded contained the information about which channel the connector failed for.

cescoffier:
Fix #515


cescoffier:
@Azquelt Can you check?

PR #527, opened 2020-04-28
Fix AMQP Outgoing metadata handling

cescoffier:
It was erasing the previously made decisions.
Fix https://github.com/quarkusio/quarkus/issues/8274.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.Priority;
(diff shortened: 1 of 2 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -345,19 +345,32 @@ private String getActualAddress(Message<?> message, io.vertx.mutiny.amqp.AmqpMes
...
-        builder.contentType(metadata.map(OutgoingAmqpMetadata::getContentType).orElse(null));
-        builder.correlationId(metadata.map(OutgoingAmqpMetadata::getCorrelationId).orElse(null));
-        builder.groupId(metadata.map(OutgoingAmqpMetadata::getGroupId).orElse(null));
-        builder.id(metadata.map(OutgoingAmqpMetadata::getId).orElse(null));
-        int priority = metadata.map(OutgoingAmqpMetadata::getPriority).orElse(-1);
-        if (priority >= 0) {
-            builder.priority((short) priority);
-        }
-        builder.subject(metadata.map(OutgoingAmqpMetadata::getSubject).orElse(null));
+        metadata.ifPresent(meta -> {
```
gsmet, line 348: Any chance you could avoid the lambda? It doesn't bring much here so better not pay for its cost.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -345,19 +345,32 @@ private String getActualAddress(Message<?> message, io.vertx.mutiny.amqp.AmqpMes
...
+            if (meta.getAddress() != null) {
+                builder.address(meta.getAddress());
+            }
+            if (meta.getProperties() != null && !meta.getProperties().isEmpty()) {
+                builder.applicationProperties(meta.getProperties());
+            }
+            if (meta.getContentEncoding() != null) {
+                builder.contentEncoding(meta.getContentEncoding());
```
kenfinnigan, line 356: Is it missing `builder.groupId`?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
cescoffier: Reactive MEssaging and Reactive is general uses lots of lambdas. I can remove this one, but there will be tons of them around.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
cescoffier: Fixed.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
cescoffier: Done.

gsmet: Yeah, maybe it's just me but I fail to see how a lambda is better than a simple `if (optional.isPresent())`.

I can see how it gets useful in more complex cases but I would try to avoid them for the simple ones.

Anyway, ignore me, I'm just ranting :).

PR #530, opened 2020-04-29
Fix the In Memory connector

cescoffier:
* Cannot switch both incoming and outgoing due to the new validation mechanism (spec enforcement)
* Update documentation
* Add jandex plugin - required to fix https://github.com/quarkusio/quarkus/issues/6427

### DIFF
--- a/smallrye-reactive-messaging-in-memory/src/main/java/io/smallrye/reactive/messaging/connectors/InMemoryConnector.java
+++ b/smallrye-reactive-messaging-in-memory/src/main/java/io/smallrye/reactive/messaging/connectors/InMemoryConnector.java
@@ -1,9 +1,6 @@
 package io.smallrye.reactive.messaging.connectors;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-in-memory/src/main/java/io/smallrye/reactive/messaging/connectors/InMemoryConnector.java
```diff
@@ -48,40 +45,79 @@
...
      * <p>
-     * With the {@link InMemorySource}, you can send messages to the channel, mocking the incoming messages.
      *
-     * You can also retrieve an {@link InMemorySink} using:
+     * @param channels the channels to switch, must not be {@code null}, must not contain {@code null}, must not contain
+     *        a blank value
+     * @return The map of properties that have been defined. The method sets the system properties, but give
+     *         you this map to pass the properties around if needed.
+     */
+    public static Map<String, String> switchIncomingChannelsToInMemory(String... channels) {
```
kenfinnigan, line 67: Do the methods need to return a `Map` of the values being set?

Given they're set as system properties, what's the use case?

cescoffier: Quarkus Test Resources when using native test.

PR #549, opened 2020-05-18
Implement AMQP reconnection

cescoffier:
If the AMQP connection gets cut, it now tries to reconnect. 
This works for both sender and receiver.

This PR also refactors the connection and client management.


cescoffier:
Important: This feature depends on Vert.x 3.9.1

cescoffier:
Fix https://github.com/quarkusio/quarkus/issues/7404.

PR #563, opened 2020-05-27
Fixes #471

### LINKED ISSUE #471: Allows emitter to be annotated with @Broadcast

kenfinnigan:
Adds support for `@Broadcast` on `Emitter` to broadcast to all subscribers

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
@@ -1,0 +1,47 @@
+package io.smallrye.reactive.messaging.extension;
+
+import org.eclipse.microprofile.reactive.messaging.OnOverflow;
+
+import io.smallrye.reactive.messaging.annotations.Broadcast;
+
+public class EmitterConfiguration {
+    private final String name;
+    private OnOverflow.Strategy overflowBufferStrategy = null;
+    private long overflowBufferSize = -1;
+    private Boolean broadcast = Boolean.FALSE;
+    private int numberOfSubscriberBeforeConnecting = -1;
+
+    public EmitterConfiguration(String name, OnOverflow onOverflow, Broadcast broadcast) {
+        this.name = name;
+
+        if (onOverflow != null) {
+            this.overflowBufferStrategy = onOverflow.value();
+            this.overflowBufferSize = onOverflow.bufferSize();
+        }
+
+        if (broadcast != null) {
+            this.broadcast = Boolean.TRUE;
+            this.numberOfSubscriberBeforeConnecting = broadcast.value();
+        }
+    }
+
+    public String getName() {
+        return this.name;
+    }
+
+    public OnOverflow.Strategy getOverflowBufferStrategy() {
+        return this.overflowBufferStrategy;
+    }
+
+    public long getOverflowBufferSize() {
+        return this.overflowBufferSize;
+    }
+
+    public Boolean isBroadcast() {
+        return broadcast;
+    }
+
+    public int getNumberOfSubscriberBeforeConnecting() {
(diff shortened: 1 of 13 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
```diff
@@ -0,0 +1,47 @@
+package io.smallrye.reactive.messaging.extension;
+
+import org.eclipse.microprofile.reactive.messaging.OnOverflow;
+
+import io.smallrye.reactive.messaging.annotations.Broadcast;
+
+public class EmitterConfiguration {
+    private final String name;
+    private OnOverflow.Strategy overflowBufferStrategy = null;
```
cescoffier, line 9: Should we make all the fields final and public?

That would avoid getters, and be "like a record".

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
```diff
@@ -581,7 +583,24 @@ public void consume(final String b) {
 
     @Test // Reproduce #511
     public void testWeCanHaveSeveralSubscribers() {
-        EmitterImpl<String> emitter = new EmitterImpl<>("my-channel", "BUFFER", 128, 128);
+        OnOverflow overflow = new OnOverflow() {
```
cescoffier, line 586: Should we define an Overflow literal?

kenfinnigan:
Yeah, I was so excited I got it working almost first go I forgot about docs!

Will do that now

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
kenfinnigan: It would mean the constructor setting values to null based on what's based in vs setting the default on the field itself.

Is that ok?

cescoffier: The constructor would have to set the fields to `null` / `false`/ `-1`. 
I don't have a strong opinion about it, I'm not sure it simplifies much.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
kenfinnigan: It's in a test, so not sure if it's that important

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
kenfinnigan: So... Leave or change? I don't mind

kenfinnigan: I will include the change for now

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/EmitterInjectionTest.java
cescoffier: Right, we can keep it like this.

kenfinnigan:
Looks like Quay is having issues, containers can't be loaded for tests

kenfinnigan:
Failure on JDK 13 with AMQP, but think it's unrelated to the change

Issue #566, opened 2020-05-29
Processor with optional outgoing payload or message


zeljkot:
The current implementation of the processor that returns payload must return a value. This means that the processor cannot decide that for some incoming message there shouldn't be outgoing message; for example, if an incoming message does not satisfy some criteria or if there was an error.

Neither `null` nor `Optional.empty` would suppress outgoing message as expected and without error.

Please provide a way to suppress outgoing messages for some incoming messages.

cescoffier:
For the time being the following approaches work:

```
@Incoming("in")
    @Outgoing("out-1")
    public Multi<String> processPayload(String s) {
        if (s.equalsIgnoreCase("skip")) {
            return Multi.createFrom().empty();
        }
        return Multi.createFrom().item(s.toUpperCase());
    }

    @Incoming("in")
    @Outgoing("out-2")
    public Multi<Message<String>> processMessage(Message<String> m) {
        String s = m.getPayload();
        if (s.equalsIgnoreCase("skip")) {
            return Multi.createFrom().empty();
        }
        return Multi.createFrom().item(m.withPayload(s.toUpperCase()));
    }

    @Incoming("in")
    @Outgoing("out-3")
    public Multi<String> processPayloadStream(Multi<String> stream) {
        return stream
            .transform().byFilteringItemsWith(s -> !s.equalsIgnoreCase("skip"))
            .onItem().apply(String::toUpperCase);
    }

    @Incoming("in")
    @Outgoing("out-4")
    public Multi<Message<String>> processMessageStream(Multi<Message<String>> stream) {
        return stream
            .transform().byFilteringItemsWith(m -> !m.getPayload().equalsIgnoreCase("skip"))
            .onItem().apply(m -> m.withPayload(m.getPayload().toUpperCase()));
    }
```

I'm adding support by returning `null` or producing `null` in a `Uni`

cescoffier:
#573 implements the `null` behavior

PR #573, opened 2020-06-01
Implement new fault-tolerance model based on Message negative acknowledgement

cescoffier:
This PR adds support for `nack` and update connector to handle failure on nacks.
For example, the Kafka connector handles nacks by either using a fail-stop, continuing the processing or writing the failing record to a dead letter queue.
The AMQP, MQTT and Camel connectors have been updated too.

With nack, the emitter can also report processing failure.

This PR also implement the possibility to skip messages (#566)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -201,49 +210,179 @@ public Metadata getMetadata() {
...
+    @Experimental("nack support is a SmallRye-only feature")
+    static <T> Message<T> of(T payload, Metadata metadata,
+            Supplier<CompletionStage<Void>> ack, Function<Throwable, CompletionStage<Void>> nack) {
+        if (payload == null) {
+            throw new IllegalArgumentException("`payload` must not be `null`");
+        }
+        if (metadata == null) {
+            metadata = Metadata.empty();
+        }
+        Metadata actual = metadata;
```
kenfinnigan, line 307: What's the purpose of this extra assignment?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -297,6 +444,24 @@ default Metadata getMetadata() {
...
+     *
+     * @param reason the reason of the nack, must not be {@code null}
+     * @return a completion stage completed when the message is negative-acknowledgement has completed. If the
+     *         negative acknowledgement fails, the completion stage propagates the failure.
+     */
+    @Experimental("nack support is a SmallRye-only feature")
+    default CompletionStage<Void> nack(Throwable reason) {
+        Function<Throwable, CompletionStage<Void>> nack = getNack();
+        if (nack == null) {
+            return CompletableFuture.completedFuture(null);
```
kenfinnigan, line 459: Has the throwable been logged anywhere prior to this point?

Would it be beneficial to log out the throwable as we're swallowing it? How would a developer know something went wrong without a configured `nack`?

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: Because it must be (implicitly) final to be used in the inner class.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: That's a good point. When you create an instance of Message, you should pass the `nack` and `ack` function, but I agree. If by any chance you use this default implementation, we should log the reason (if there is no nack function)

cescoffier: Fixed.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -297,6 +449,32 @@ default Metadata getMetadata() {
...
+     */
+    @Experimental("nack support is a SmallRye-only feature")
+    default CompletionStage<Void> nack(Throwable reason) {
+        if (reason == null) {
+            throw new IllegalArgumentException("The reason must not be `null`");
+        }
+        Function<Throwable, CompletionStage<Void>> nack = getNack();
+        if (nack == null) {
+            LOGGER
+                    .warn("A message has been nacked, but no nack function has been provided. The reason was: {}",
```
kenfinnigan, line 468: Does `{}` bring in the message? Thought the format was `%s` or the like?

cescoffier: Using SLF4J, not JBoss logging.

kenfinnigan: Ah ok, we may want to re-evaluate that in the future

cescoffier:
This PR must be rebased with the fix for #574.

PR #575, opened 2020-06-01
Call the `accepted` method on the right context

### LINKED ISSUE #574: AMQP connector can make illegal threading use of vertx-proton, potentially corrupting the wire output and causing connection failure
@hguerrero asked about some issues he was having while trying out a example using Quarkus with smallrye-reactive-messaging-amqp with the following consuming code:

```
    @Incoming("example-in")
    public CompletionStage<Void> receive(String message) {
        return CompletableFuture.runAsync(() -> {
            System.out.println(message);
        });
    }
```
From a protocol trace it could be seen the remote server was specifically closing the connection. However it also showed two identical logs of sending a disposition for a received message just before this.

```
[1813009805:0] -> Disposition{role=RECEIVER, first=2, last=2, settled=true, state=Accepted{}, batchable=false}
[1813009805:0] -> Disposition{role=RECEIVER, first=2, last=2, settled=true, state=Accepted{}, batchable=false}
[1813009805:0] <- Close{error=null}
```

Looking into this it becomss clear that there are threading issues at play. Vertx-proton (and below that proton-j) is inherently single threaded, for use from the event loop thread it is created against only. The above code provokes the smallrye bits try and set a disposition from on the ForkJoin thread pool, whose thread makes it through to vertx-amqp-client and down into vertx-proton (and proton-j) itself. This causes illegal concurrent usage along with vertx-proton's own single threaded use of proton-j from its event loop thread, and can lead to the output stream getting trashed, and then depending on what is actually written the broker can barf on it and provoke it to close the connection (which ideally it would do with error detail set; still to look at that).

cescoffier:
Fix #574

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessage.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessage.java
@@ -4,16 +4,13 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
 import org.apache.qpid.proton.amqp.Binary;
-import org.apache.qpid.proton.amqp.messaging.AmqpSequence;
-import org.apache.qpid.proton.amqp.messaging.AmqpValue;
-import org.apache.qpid.proton.amqp.messaging.Data;
-import org.apache.qpid.proton.amqp.messaging.Header;
-import org.apache.qpid.proton.amqp.messaging.Section;
+import org.apache.qpid.proton.amqp.messaging.*;
 import org.apache.qpid.proton.message.Message;
 import org.apache.qpid.proton.message.MessageError;
 import org.eclipse.microprofile.reactive.messaging.Metadata;
 
 import io.vertx.core.json.JsonObject;
+import io.vertx.mutiny.core.Context;
 import io.vertx.mutiny.core.buffer.Buffer;
 
 public class AmqpMessage<T> implements org.eclipse.microprofile.reactive.messaging.Message<T> {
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessage.java
```diff
@@ -87,8 +86,15 @@ public AmqpMessage(io.vertx.amqp.AmqpMessage msg) {
...
-        this.message.accepted();
-        return CompletableFuture.completedFuture(null);
+        // We must switch to the context having created the message.
+        // This context is passed when this instance of message is created.
+        // It's more a Vert.x AMQP client issue which should ensure calling `accepted` on the right context.
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        this.context.runOnContext(x -> {
+            this.message.accepted();
+            future.complete(null);
+        });
```
gemmellr, line 96: Should it check it needs to do this? Doesn't this also rather change the behaviour anywhere ack() was already being called on the correct thread?

cescoffier: Do you mean implementing a trampoline? 

Hopefully, this is a temporary fix and the Vert.x AMQP client would do it more consistently.

gemmellr: I just mean that if the ack method was already being used in other places on the correct event loop thread then it would have actually done the work when called previously, whereas now it will only do it later, possibly after whatever the thread attempts next once it has called ack. The difference may or may not be important. I'm mainly just surprised to see that change rather than only the off-event-loop handling.

cescoffier: So you mean implementing a trampoline to check if we are on the correct context, and if not doing this logic. It can be done. I would have preferred having it done in the vert.x amqp client than here, as this is just a temporary fix for the real bug (which is in the Vert.x AMQP client).

gemmellr: Yep. I realise you see this as a temporary fix I just figured it would have much the same effect, such that the unaffected bits dont see any change in behaviour now, or again later when the temp fix is removed and it goes back.

gemmellr: Looks good

PR #586, opened 2020-06-03
Consistent naming in poms

phillip-kruger:
A small PR that add consistent naming in the poms. That might help navigating the modules (especially for a newcomer)

Before:
before

After:
after

Signed-off-by:Phillip Kruger <phillip.kruger@gmail.com>

### DIFF
--- a/examples/quickstart/pom.xml
+++ b/examples/quickstart/pom.xml
@@ -12,7 +12,7 @@
 
   <artifactId>reactive-messaging-quickstart</artifactId>
   
-  <name>SmallRye Reactive Messaging : Quickstart :: Memory</name>    
+  <name>SmallRye Reactive Messaging : Quickstart :: General</name>    
 
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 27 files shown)


examples/quickstart/pom.xml
```diff
@@ -11,6 +11,8 @@
   </parent>
 
   <artifactId>reactive-messaging-quickstart</artifactId>
+  
+  <name>SmallRye Reactive Messaging : Quickstart :: Memory</name>
```
kenfinnigan, line 15: I think this should just be "Quickstart :: General" as though it's using in memory, I don't think that was the main purpose

smallrye-connector-attribute-processor/pom.xml
```diff
@@ -12,6 +12,8 @@
 
   <artifactId>smallrye-connector-attribute-processor</artifactId>
 
+  <name>SmallRye Reactive Messaging : Attribute Processor</name>
```
kenfinnigan, line 15: "Connector Attribute Processor"

PR #594, opened 2020-06-08
Credit-based back-pressure for the outbound AMQP connector

### LINKED ISSUE #491: Handle credit based back pressure in the AMQP connector
Instead of acknowledge-based back-pressure, switch to a credit-based backpressure.

It requires changing the AMQP "sink" to _request_ the number of messages equal to the number of (AMQP) credits. When the request number is exhausted, check if we have new credit. Retry-periodically until we do.

cescoffier:
Fixed #491.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -3,15 +3,11 @@ package io.smallrye.reactive.messaging.amqp;
 import static io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction.*;
 import static java.time.Duration.ofSeconds;
 
-import java.time.Instant;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -133,7 +123,7 @@ void setup(ExecutionHolder executionHolder) {
 
         AmqpClient client = AmqpClientHelper.createClient(this, ic, clientOptions);
         String link = ic.getLinkName().orElseGet(ic::getChannel);
-        ConnectionHolder holder = new ConnectionHolder(client, ic);
+        ConnectionHolder holder = new ConnectionHolder(client, ic, getVertx());
```
kenfinnigan, line 126: Should we be passing Vertx or the ExecutionHolder?

cescoffier: Here, we only need the vert.x instance. But maybe later we will need more.

Issue #577, opened 2020-06-01
Add support for Kafka's ConsumerRebalanceListener


pcasaes:
Sometimes _earliest_ and _latest_ don't quite cut it, not to mention some interesting deliver exactly once cases as described in the **Kafka: The Definitive Guide**. There would be a fait bit of work cut out though. From what I can understand from the code  subscription is started during configuration. The listener would have to be registered before subscription somehow. Fortunately Vert.x' Kafka client implementation already has support for this.

cescoffier:
Can do give me a bit more details? Do you mean adding the ability to use a custom "poller" ?

cescoffier:
I finally understand...
https://www.learningjournal.guru/courses/kafka/kafka-foundation-training/rebalance-listener/ provides a good example. 

It's definitely something we should add.

pcasaes:
That's a good example. The Listener is always called in the first poll within the poll loop thread. It is also called whenever there's a rebalance which can because by an unhealthy consumer, but it can also be caused by adding/removing consumers as well as increase the number of partitions.

One example I have is a kafka backed in memory cache. The topic has a retention period but I don't want to wait for kafka to clean up the partitions. I also don't want my warm up to take too long loading up from earliest since it definitely will have data that is too old. I can use a rebalance to do an offset by timestamp look up (there's an index in kafka for that) and then set the offsets accordingly.

cescoffier:
It should be rather simple to integrate.

cescoffier:
So, I was thinking to integrate it as follow:

* the `ConsumerRebalanceListener` would be exposed as a bean, maybe `@Named` if needed
* if `@Named` is used, add an attribute to configure the name
* The Kafka Connector would retrieve the bean using the `bean manager`
* In Kafka Source, use the `subscribe` method that supports the listener, like in:
```
this.consumer.getDelegate().unwrap().subscribe(topic, listener);
```

@pcasaes fancy a PR? I can guide you.

pcasaes:
Sounds good. I'll give it a shot.

cescoffier:
Thanks @pcasaes, don't hesitate to ping me if you have any questions.

PR #601, opened 2020-06-11
Add support for Kafka's ConsumerRebalanceListener

pcasaes:
Details https://github.com/smallrye/smallrye-reactive-messaging/issues/577

Had to create a custom listener interface that also receives the KafkaConsumer beside the assigned topics with partitions. This is necessary for use cases that involves seeking to specific offsets during rebalance/initialization.

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -69,4 +69,8 @@ Type: _string_ | false |
 
 Type: _int_ | false | `1`
 
+| *consumer-rebalance-listener.name* | The name set in `javax.inject.Named` of a bean that implements `io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener`. If set the listener will be applied to the consumer.
+
+Type: _string_ | false | 
+
 |===
--- a/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
+++ b/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
@@ -1,0 +1,60 @@
+package inbound;
+
+import io.smallrye.mutiny.Uni;
+import io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener;
+import io.vertx.kafka.client.common.TopicPartition;
+import io.vertx.mutiny.kafka.client.consumer.KafkaConsumer;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Named;
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+@ApplicationScoped
+@Named("rebalanced-example.rebalancer")
+public class KafkaRebalancedConsumerRebalanceListener implements KafkaConsumerRebalanceListener {
+
+    private static final Logger LOGGER = Logger.getLogger(KafkaRebalancedConsumerRebalanceListener.class.getName());
+
+    /**
+     * When receiving a list of partitions will search for the earliest offset within 10 minutes
+     * and seek the consumer to it.
+     *
+     * @param consumer        underlying consumer
+     * @param topicPartitions set of assigned topic partitions
+     * @return A {@link Uni} indicating operations complete or failure
+     */
+    @Override
+    public Uni<Void> onPartitionsAssigned(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions) {
+        long now = System.currentTimeMillis();
+        long shouldStartAt = now - 600_000L; //10 minute ago
+
+        return Uni
+            .combine()
+            .all()
+            .unis(topicPartitions
+                .stream()
+                .map(topicPartition -> {
+                    LOGGER.info("Assigned " + topicPartition);
+                    return consumer.offsetsForTimes(topicPartition, shouldStartAt)
+                        .onItem()
+                        .invoke(o -> LOGGER.info("Seeking to " + o))
+                        .onItem()
+                        .produceUni(o -> consumer
+                            .seek(topicPartition, o == null ? 0L : o.getOffset())
+                            .onItem()
+                            .invoke(v -> LOGGER.info("Seeked to " + o))
+                        );
+                })
+                .collect(Collectors.toList()))
+            .combinedWith(a -> null);
+    }
+
+    @Override
+    public Uni<Void> onPartitionsRevoked(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions) {
+        return Uni
+            .createFrom()
+            .nullItem();
+    }
+}
--- a/documentation/src/main/doc/modules/kafka/pages/consumer-rebalance-listener.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/consumer-rebalance-listener.adoc
@@ -1,0 +1,36 @@
+[#kafka-consumer-rebalance-listener]
+=== Consumer Rebalance Listener
+
+An implementation of the consumer re-balance listener can be provided which affords us fine grain controls of the assigned
+offset. Common uses are storing offsets in a separate store to enable deliver exactly-once semantics, and starting from
+a specific time window.
+
+Whenever the topic partitions assigned method is called the consumer will pause. It will only resume once the returned
+Uni succeeds. In the case of failure it will retry until success or until the consumer session times out in which case
+it will resume again forcing a new consumer re-balance.
+
+==== Example
+
+In this example we will set-up a consumer that always starts on messages from at most 10 minutes ago. First we need to provide
+a bean managed implementation of `io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener` annotated with
+`javax.inject.Named`. We then must configure our inbound connector to use this named bean.
+
+[source, java]
+----
+include::example$inbound/KafkaRebalancedConsumerRebalanceListener.java[]
+----
+
+[source, java]
+----
+include::example$inbound/KafkaRebalancedConsumer.java[]
+----
+
+To configure the inbound connector to use the provided listener we either set the consumer rebalance listener's name:
+
+* `mp.messaging.incoming.rebalanced-example.consumer-rebalance-listener.name=rebalanced-example.rebalancer`
+
+Or have the listener's name be the same as the group id:
+
+* `mp.messaging.incoming.rebalanced-example.group.id=rebalanced-example.rebalancer`
+
+Setting the consumer re-balance listener's name takes precedence over using the group id.
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
@@ -66,12 +66,16 @@ import io.vertx.mutiny.core.Vertx;
 @ConnectorAttribute(name = "partition", type = "int", direction = Direction.OUTGOING, description = "The target partition id. -1 to let the client determine the partition", defaultValue = "-1")
 @ConnectorAttribute(name = "waitForWriteCompletion", type = "boolean", direction = Direction.OUTGOING, description = "Whether the client waits for Kafka to acknowledge the written record before acknowledging the message", defaultValue = "true")
 @ConnectorAttribute(name = "max-inflight-messages", type = "int", direction = Direction.OUTGOING, description = "The maximum number of messages to be written to Kafka concurrently - The default value is the value from the `max.in.flight.requests.per.connection` Kafka property. It configures the maximum number of unacknowledged requests the client before blocking. Note that if this setting is set to be greater than 1 and there are failed sends, there is a risk of message re-ordering due to retries.", defaultValue = "5")
+@ConnectorAttribute(name = "consumer-rebalance-listener.name", type = "string", direction = Direction.INCOMING, description = "The name set in `javax.inject.Named` of a bean that implements `io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener`. If set the listener will be applied to the consumer.")
 public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnectorFactory {
 
     public static final String CONNECTOR_NAME = "smallrye-kafka";
 
     @Inject
     ExecutionHolder executionHolder;
+
+    @Inject
+    Instance<KafkaConsumerRebalanceListener> consumerRebalanceListeners;
 
     private final List<KafkaSource<?, ?>> sources = new CopyOnWriteArrayList<>();
     private final List<KafkaSink> sinks = new CopyOnWriteArrayList<>();
@@ -106,7 +110,7 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
         }
 
         if (partitions == 1) {
-            KafkaSource<Object, Object> source = new KafkaSource<>(vertx, ic);
+            KafkaSource<Object, Object> source = new KafkaSource<>(vertx, ic, consumerRebalanceListeners);
             sources.add(source);
 
             boolean broadcast = ic.getBroadcast();
@@ -120,7 +124,7 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
         // create an instance of source per partitions.
         List<Publisher<IncomingKafkaRecord<Object, Object>>> streams = new ArrayList<>();
         for (int i = 0; i < partitions; i++) {
-            KafkaSource<Object, Object> source = new KafkaSource<>(vertx, ic);
+            KafkaSource<Object, Object> source = new KafkaSource<>(vertx, ic, consumerRebalanceListeners);
             sources.add(source);
             streams.add(source.getStream());
         }
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
@@ -1,0 +1,63 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import java.util.Set;
+
+import io.smallrye.mutiny.Uni;
+import io.vertx.kafka.client.common.TopicPartition;
+import io.vertx.mutiny.kafka.client.consumer.KafkaConsumer;
+
+/**
+ *
+ * When implemented by a managed bean annotated with {@link javax.inject.Named} and
+ * configured against an inbound connector will be applied as a consumer re-balance listener
+ * to that inbound connector's consumer.
+ *
+ *
+ * To configure which listener you want to use, set the name in the inbound connector's consumer re-balance listener attribute,
+ * ex:
+ * {@code
+ *  mp.messaging.incoming.example.consumer-rebalance- listener.name=ExampleConsumerRebalanceListener
+ * }
+ * {@code @Named("ExampleConsumerRebalanceListener")}
+ *
+ * Alternatively, name your listener (using the {@code @Named} annotation) to be the group id used by the connector, ex:
+ * {@code
+ *  mp.messaging.incoming.example.group.id=my-group
+ * }
+ * {@code @Named("my-group")}
+ *
+ * Setting the consumer re-balance listener name takes precedence over using the group id.
+ *
+ * For more details:
+ *
+ * @see org.apache.kafka.clients.consumer.ConsumerRebalanceListener
+ */
+public interface KafkaConsumerRebalanceListener {
+
+    /**
+     * Called when the consumer is assigned topic partitions
+     * This method might be called for each consumer available to the connector
+     *
+     * The consumer will be paused until the returned {@link Uni}
+     * indicates success. On failure will retry using an exponential back off until the consumer can
+     * be considered timed-out by Kafka, in which case will resume anyway triggering a new re-balance.
+     *
+     * @see KafkaConsumer#pause()
+     * @see KafkaConsumer#resume()
+     *
+     * @param consumer underlying consumer
+     * @param topicPartitions set of assigned topic partitions
+     * @return A {@link Uni} indicating operations complete or failure
+     */
+    Uni<Void> onPartitionsAssigned(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions);
+
+    /**
+     * Called when the consumer is revoked topic partitions
+     * This method might be called for each consumer available to the connector
+     *
+     * @param consumer underlying consumer
+     * @param topicPartitions set of revoked topic partitions
+     * @return A {@link Uni} indicating operations complete or failure
+     */
+    Uni<Void> onPartitionsRevoked(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions);
+}
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
@@ -93,4 +93,41 @@ public interface KafkaLogging extends BasicLogger {
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 18218, value = "An exception has been caught while closing the Kafka consumer")
     void exceptionOnClose(@Cause Throwable t);
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 18219, value = "Loading KafkaConsumerRebalanceListener from configured name '%s'")
+    void loadingConsumerRebalanceListenerFromConfiguredName(String configuredName);
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 18220, value = "Loading KafkaConsumerRebalanceListener from group id '%s'")
+    void loadingConsumerRebalanceListenerFromGroupId(String consumerGroup);
+
+    @LogMessage(level = Logger.Level.ERROR)
+    @Message(id = 18221, value = "Unable to execute consumer assigned re-balance listener for group '%s'. The consumer has been paused. Will retry until the consumer session times out in which case will resume to force a new re-balance attempt.")
+    void unableToExecuteConsumerAssignedRebalanceListener(String consumerGroup, @Cause Throwable t);
+
+    @LogMessage(level = Logger.Level.ERROR)
+    @Message(id = 18222, value = "Unable to execute consumer revoked re-balance listener for group '%s'")
+    void unableToExecuteConsumerRevokedRebalanceListener(String consumerGroup, @Cause Throwable t);
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 18223, value = "Executing consumer assigned re-balance listener for group '%s'")
+    void executingConsumerAssignedRebalanceListener(String consumerGroup);
+
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 18224, value = "Executing consumer revoked re-balance listener for group '%s'")
+    void executingConsumerRevokedRebalanceListener(String consumerGroup);
+
+    @LogMessage(level = Logger.Level.INFO)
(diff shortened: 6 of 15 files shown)


documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
```diff
@@ -65,4 +65,8 @@ Type: _string_ | false |
 
 Type: _string_ | false | 
 
+| *consumer-rebalance-listener.name* | The name set in javax.inject.Named of a bean that implements io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener. If set the listener will be applied to the consumer.
```
cescoffier, line 68: Can use use `javax.inject.Named` and `io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener` (purely cosmetic comment)

cescoffier: The change must be done in the connector class.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -61,13 +62,17 @@
...
+@ConnectorAttribute(name = "consumer-rebalance-listener.name", type = "string", direction = Direction.INCOMING, description = "The name set in javax.inject.Named of a bean that implements io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener. If set the listener will be applied to the consumer.")
 public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnectorFactory {
 
     public static final String CONNECTOR_NAME = "smallrye-kafka";
 
     @Inject
     private ExecutionHolder executionHolder;
 
+    @Inject
+    private BeanManager beanManager;
```
cescoffier, line 74: You can also use an `Instance<KafkaConsumerRebalanceListener>` - the API is a bit simpler. 
See https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java#L79 and https://github.com/smallrye/smallrye-reactive-messaging/blob/a9b33e291f150ef72b840daa99deb2738044e17a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java#L21 as an example.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,37 @@
...
+ * configured against a consumer, ex:
+ * mp.messaging.incoming.example.consumer-rebalance-listener.name=ExampleConsumerRebalanceListener
+ *
+ * Will be applied as a consumer rebalance listener to the consumer.
+ *
+ * For more details
+ *
+ * @see org.apache.kafka.clients.consumer.ConsumerRebalanceListener
+ */
+public interface KafkaConsumerRebalanceListener {
```
cescoffier, line 20: What's the reason for not using `org.apache.kafka.clients.consumer.ConsumerRebalanceListener` directly?

cescoffier: And then I read the next file and understood it. Sorry!

cescoffier:
Please rebase. I should fix the CI issue.

pcasaes:
I thought the same thing. I wanted to to get the code out for review first.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
pcasaes: Sure thing!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
pcasaes: For documentation's sake it's because most of the time you'll want to manipulate the consumer object (like seeking to an offset). The original listener interface is designed with an imperative synchronous style in mind https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html

I originally wrote it with the the vanilla listener interface then saw the problem 😆

With an example in the docs it'll be clearer.

pcasaes:
@cescoffier Added an example in the docs. I thought occurred to me though. We could dispense with the `consumer-rebalance-listener.name` configuration if we were to use the same `channel-name` on the listener. What do you think?

Pro: less configuration work.
Con: Multiple inbound connectors won't be able to share a listener.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * When receiving a list of partitions will search for the earliest offset within 10 minutes
+     * and seek the consumer to it. These operations are asynchronous so the inbound connector
+     * WILL continue to receive messages from the subscribed topic that MIGHT be older than 10 minutes.
+     *
+     * @param consumer underlying consumer
+     * @param topicPartitions set of assigned topic partitions
+     */
+    @Override
+    public void onPartitionsAssigned(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions) {
```
cescoffier, line 35: Instead of "void" should it returns a `Uni<Void>` indicating when the action has completed?

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Because the operations in {@link #onPartitionsAssigned(KafkaConsumer, Set)} are asynchronous we
+     * MIGHT need a way to inform the inbound connector to ignore incoming messages until this listener
+     * has done its job.
+     *
+     * @param message
+     * @return true if the consumer has been properly set-up for the messages partition
+     */
+    public boolean isConsumerReady(IncomingKafkaRecord<?, ?> message) {
```
cescoffier, line 78: Returning Uni<Void> would remove the need for this method.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
```diff
@@ -0,0 +1,86 @@
...
+                                .seek(topicPartition, offsetAndTimestamp.getOffset())
+                                .subscribe()
+                                .with(v -> LOGGER.info("Seeked to " + offsetAndTimestamp));
+                        }
+                    });
+            });
+    }
+
+    @Override
+    public void onPartitionsRevoked(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions) {
```
cescoffier, line 66: Same, here, should it return a `Uni<Void>`?

pcasaes:
I was able to get it working but the solution isn't very elegant. It doesn't solve the fundamental problem which is that we need to wait for the partition assigned Uni to complete before continuing with the poll. I'll take a look at kafka mutiny module and see what can be done.

pcasaes:
@cescoffier I got it working well enough. The KafkaConsumer class has pause and resume methods which made the work possible.

Did you get chance to think about my comments on https://github.com/smallrye/smallrye-reactive-messaging/pull/601#issuecomment-642926677 ?

cescoffier:
> @cescoffier Added an example in the docs. I thought occurred to me though. We could dispense with the consumer-rebalance-listener.name configuration if we were to use the same channel-name on the listener. What do you think?

> Pro: less configuration work.
> Con: Multiple inbound connectors won't be able to share a listener.

@pcasaes Interesting idea. The questions are:

1. do we want to use a default one (if there is one exposed without a name)?
2. do we want to use one having the channel name as name?

About 1:

I'm not sure, it's a fairly advanced use case, you probably don't want to use it unconsciously. So, I would say no: it must have a name.

About 2:

I like the idea. At least the developer is aware of it. It does not reduce the flexibility as you can still have a specific name and configure the attribute in that case. So, you can still have a listener shared by multiple channels.

The question is about channel name or topic name. As it's quite Kafka specific, I would say that it should use the topic name, and not the channel name. 

So:

1. if the configuration attribute is not set and there is a listener named after the topic name you are consuming, you use that one. A log message should be added to indicate the choice.
2. if the configuration attribute is set and there is no listener with that name, you fail.
3. if the configuration attribute is set and there is a listener with that name, you use that one.

WDYT?

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
```diff
@@ -0,0 +1,71 @@
...
+                            .onItem()
+                            .invoke(v -> LOGGER.info("Seeked to " + o))
+                        );
+                })
+                .collect(Collectors.toList()))
+            .combinedWith(a -> null)
+            .onItemOrFailure()
+            .apply((a, t) -> {
+                // once the seek is complete let's resume the consumer
+                consumer.resume();
```
cescoffier, line 60: Do you want to resume on failure?

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param consumer        underlying consumer
+     * @param topicPartitions set of assigned topic partitions
+     * @return An observable
+     */
+    @Override
+    public Uni<Void> onPartitionsAssigned(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions) {
+        // we must pause the consumer otherwise the inbound connector will continue to receive messages
+        // while we seek to the correct offset.
+        consumer.pause();
```
cescoffier, line 33: Question: should it be done by the connector, or by the implementation of the`KafkaConsumerRebalanceListener`?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Called when the consumer is assigned topic partitions
+     * This method might be called for each consumer available to the connector
+     *
+     * @param consumer underlying consumer
+     * @param topicPartitions set of assigned topic partitions
+     * @return An observable
```
cescoffier, line 29: so, it's not an `observable`, it's a Uni indicating when the operation has completed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    Uni<Void> onPartitionsAssigned(KafkaConsumer<?, ?> consumer, Set<TopicPartition> topicPartitions);
+
+    /**
+     * Called when the consumer is revoked topic partitions
+     * This method might be called for each consumer available to the connector
+     *
+     * @param consumer underlying consumer
+     * @param topicPartitions set of revoked topic partitions
+     * @return An observable
```
cescoffier, line 39: so, it's not an observable, it's a Uni indicating when the operation has completed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
cescoffier: or failed :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
```diff
@@ -93,4 +93,12 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 18218, value = "An exception has been caught while closing the Kafka consumer")
     void exceptionOnClose(@Cause Throwable t);
+
+    @LogMessage(level = Logger.Level.ERROR)
+    @Message(id = 18219, value = "Unable to execute consumer rebalance listener for group '%s'")
+    void unableToExecuteConsumerReblanceListener(String consumerGroup, @Cause Throwable t);
```
cescoffier, line 99: Great! You already use the new logging system!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -59,14 +68,47 @@ public KafkaSource(Vertx vertx, KafkaConnectorIncomingConfiguration config) {
...
+        final List<Publisher<KafkaConsumerRecord<K, V>>> multis = new ArrayList<>();
+        config
+                .getConsumerRebalanceListenerName()
+                .map(name -> {
+                    log.info("Loading KafkaConsumerRebalanceListener " + name);
+                    return NamedLiteral.of(name);
+                })
+                .map(consumerRebalanceListeners::select)
+                .map(Instance::get)
+                .ifPresent(listener -> multis.add(Multi
```
cescoffier, line 81: Do we want to handle the case with multiple listeners for a single topic?

pcasaes:
I gave it some thought. I understand why we wouldn't want to use the channel name but the topic name doesn't seem quite right either. You couldn't conceivably set up two connectors against the same topic. How about the consumer group? That makes more sense. It's in the name after all: consumer rebalance listener. And setting up two consumers instances on the same group but with different setups seems incorrect.

pcasaes:
@cescoffier I think it's done. Let me know what you think.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: There won't be multiple listeners. But the listener is called whenever there's a rebalance not just during startup.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
pcasaes: That's up to the user really.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
pcasaes: I thought about this. The listener isn't always used to seek to separate offset, so it's really up to the user.

pcasaes: On further thought. It does make sense to do it. By the kafka documentation it is expected that the operations in the listener are completed before receiving events. If the user wishes to override they can resume from withing their implementation.

cescoffier: Agreed.

cescoffier: So it needs to be changed in this example and explained in the doc.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
cescoffier: Got it!

pcasaes:
Went ahead and moved it to the resume top to the connector. In the case of a failure kafka will handle it. If a consumer instance is paused for a certain amount of time it will trigger a consumer rebalance.

pcasaes:
@cescoffier 

1. Used the logger for all logs

2. Removed the merging Multi (that was a relic from an older attempt).

3. We have to resume the consumer for it to trigger a rebalance
This one was crucial and was discovered by adding a test for this case. So if the rebalance Uni results in an error we wait _session.timeout.ms_ + _max.poll.interval.ms_ before resuming the consumer.

pcasaes:
I think I can do better with shorter retries.

pcasaes:
@cescoffier 
If assigning fails will keep retrying using an exponential back off until the kafka consumer session can be considered timed out. According to the documentation it's _max.poll.interval.ms_ or _max.poll.interval.ms + session.timeout.ms_ if _group.instance.id_ is not null. Once timed out we'll resume the consumer which will trigger a new consumer re-balance. Added tests to both cases, success on a simple retry and success only after triggering a new consumer re-balance.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
pcasaes: done

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
pcasaes: done

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
pcasaes: done

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -59,7 +66,74 @@ public KafkaSource(Vertx vertx, KafkaConnectorIncomingConfiguration config) {
         kafkaConfiguration.remove("broadcast");
         kafkaConfiguration.remove("partitions");
```
cescoffier, line 67: The new attribute need to be removed from the kafkaConfiguration map. It avoids a warning.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -59,7 +66,74 @@ public KafkaSource(Vertx vertx, KafkaConnectorIncomingConfiguration config) {
...
+                        log.executingConsumerAssignedRebalanceListener(group);
+                        listener.onPartitionsAssigned(kafkaConsumer, set)
+                                .onFailure().invoke(t -> log.unableToExecuteConsumerAssignedRebalanceListener(group, t))
+                                .onFailure().retry().withBackOff(Duration.ofSeconds(1), Duration.ofSeconds(10)).withJitter(0.0)
+                                .atMost(consumerReEnableRetryMaxAttempts)
+                                .onFailure().invoke(t -> {
+                                    log.reEnablingConsumerforGroup(group);
+                                    kafkaConsumer.resume();
+                                })
+                                .onItem().invoke(a -> kafkaConsumer.resume())
```
cescoffier, line 122: You may resume twice (not sure if it's harmful). I would recommend moving the onFailure and onItem in the `subscribe` part.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -59,7 +66,74 @@ public KafkaSource(Vertx vertx, KafkaConnectorIncomingConfiguration config) {
...
+                    // than consumerReEnableWaitTime
+                    final long consumerReEnableRetryMaxAttempts = 1
+                            + Math.max(0, 3 + (consumerReEnableWaitTime - 14_000) / 10_000);
+
+                    kafkaConsumer.partitionsAssignedHandler(set -> {
+                        kafkaConsumer.pause();
+                        log.executingConsumerAssignedRebalanceListener(group);
+                        listener.onPartitionsAssigned(kafkaConsumer, set)
+                                .onFailure().invoke(t -> log.unableToExecuteConsumerAssignedRebalanceListener(group, t))
+                                .onFailure().retry().withBackOff(Duration.ofSeconds(1), Duration.ofSeconds(10)).withJitter(0.0)
```
cescoffier, line 116: Should we remove the jitter?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -59,7 +66,74 @@ public KafkaSource(Vertx vertx, KafkaConnectorIncomingConfiguration config) {
...
+                                })
+                                .onItem().invoke(a -> kafkaConsumer.resume())
+                                .subscribe()
+                                .with(a -> log.executedConsumerAssignedRebalanceListener(group));
+                    });
+
+                    kafkaConsumer.partitionsRevokedHandler(set -> {
+                        log.executingConsumerRevokedRebalanceListener(group);
+                        listener.onPartitionsRevoked(kafkaConsumer, set)
+                                .onFailure().invoke(t -> log.unableToExecuteConsumerRevokedRebalanceListener(group, t))
```
cescoffier, line 130: I would recommend moving this to the subscribe.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceTest.java
```diff
@@ -388,15 +506,64 @@ public void testSourceWithEmptyOptionalConfiguration() {
...
         Weld weld = baseWeld();
         addConfig(config);
         weld.addBeanClass(ConsumptionBean.class);
+        weld.addBeanClass(ConsumptionConsumerRebalanceListener.class);
         weld.disableDiscovery();
         container = weld.initialize();
         return container.getBeanManager().createInstance().select(ConsumptionBean.class).get();
     }
 
+    private ConsumptionBeanWithoutAck deployWithoutAck(MapBasedConfig config) {
```
cescoffier, line 554: A test checking that we fail if a listener is configured but cannot be found should be added.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: I think so. I'm worried that it would cause the resume to start before the session times out. Unfortunately `UniRetry` doesn't have a `untilTime` method. That would've been super helpful. It has an `until` but that doesn't do an exponential backoff.

documentation/src/main/doc/modules/kafka/pages/consumer-rebalance-listener.adoc
```diff
@@ -0,0 +1,32 @@
+[#kafka-consumer-rebalance-listener]
+=== Consumer Rebalance Listener
+
+An implementation of the consumer re-balance listener can be provided which affords us fine grain controls of the assigned
+offset. Common uses are storing offsets in a separate store to enable deliver exactly-once semantics, and starting from
+a specific time window.
```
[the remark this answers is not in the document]
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,58 @@
...
+import io.vertx.mutiny.kafka.client.consumer.KafkaConsumer;
+
+/**
+ *
+ * When implemented by a managed bean annotated with {@link javax.inject.Named} and
+ * configured against an inbound connector will be applied as a consumer re-balance listener
+ * to that inbound connector's consumer.
+ *
+ *
+ * To configure set the name in the inbound connector's consumer re-balance listener, ex:
```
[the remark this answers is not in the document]
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,58 @@
...
+ * mp.messaging.incoming.example.consumer-rebalance-listener.name=ExampleConsumerRebalanceListener
+ * @Named("ExampleConsumerRebalanceListener")
+ *
+ * Or set the name to be the same as the group id, ex:
```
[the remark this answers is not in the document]
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ *
+ * When implemented by a managed bean annotated with {@link javax.inject.Named} and
+ * configured against an inbound connector will be applied as a consumer re-balance listener
+ * to that inbound connector's consumer.
+ *
+ *
+ * To configure set the name in the inbound connector's consumer re-balance listener, ex:
+ * mp.messaging.incoming.example.consumer-rebalance-listener.name=ExampleConsumerRebalanceListener
```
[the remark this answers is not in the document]
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumerRebalanceListener.java
```diff
@@ -0,0 +1,58 @@
...
+ * @Named("ExampleConsumerRebalanceListener")
+ *
+ * Or set the name to be the same as the group id, ex:
+ * mp.messaging.incoming.example.group.id=my-group
```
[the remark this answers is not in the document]
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: done

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceTest.java
pcasaes: done

pcasaes:
All done

cescoffier:
Amazing! Thanks for the hard work @pcasaes!

PR #609, opened 2020-06-16
Configure Kafka Source to use the same group-id if configured for con…

pcasaes:
…currency (partitions > 1)

If the user configured `group-id` in the inbound connector configs there is no issue. But if not configured we were generating a unique group id per `KafkaSource`. In this case each `KafkaSource` would receive all events from the Topic with no load balancing.


pcasaes:
Hi @cescoffier while working on the kafka consumer rebalance I noticed a bug on the concurrent consumer work. It's a small bug with a simple fix.

cescoffier:
Thanks!

PR #624, opened 2020-06-24
Fix nack propagation for Kafka and add test

kenfinnigan:


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/OutgoingKafkaRecord.java
```diff
@@ -37,7 +40,8 @@ public OutgoingKafkaRecord(String topic, K key, T value, Instant timestamp, int
                 .orElseThrow(() -> ex.illegalArgumentNoMetadata(OutgoingKafkaRecordMetadata.class));
```
cescoffier, line 40: That needs to be changed too. You can't expect to have OutgoingKafkaRecordMetadata, also you need to be sure you keep the existing set. To do that, the `OutgoingKafkaRecord ` constructor should receive the original Metadata.

kenfinnigan: I tried retaining the incoming metadata, but managed to break other tests! I can try again though.

What would be a good way to test that the current approach is losing metadata?

cescoffier:
Same as for ack.
Get a message with some metadata.
In another method, create a new Outgoing Kafka Record from this first
message
In another method (consumer), retrieve the message and check that the
initial metadata are still there (as well as the Outgoing Kafka metadata)

Le jeu. 25 juin 2020 à 13:47, Ken Finnigan <notifications@github.com> a
écrit :

kenfinnigan:
Do you happen to know the test name that does it for ack? If not I can find it

kenfinnigan:
So when you say we can't expect to have `OutgoingKafkaRecordMetadata` present, what do we do if it isn't there?

Currently it throws the illegal argument exception, but you're implying it shouldn't do that right? If not, what instead, as I don't think we can then set the topic, etc on the outgoing record without it?

cescoffier:
I guess this method was created to build an `OutgoingKafkaRecord` from a message containing the right metadata. But in your case, you want to create a new `OutgoingKafkaRecord` even if there are no Kafka metadata. The `with` method would set different attributes. If there is no `OutgoingKafkaRecordMetadata`, create one with default values.

kenfinnigan:
Thanks

kenfinnigan:
Should be good now

kenfinnigan:
MQTT Server failures on JDK 11 are unrelated, I think.

Is there a race condition?

Issue #631, opened 2020-06-28
Performance degradation while producing messages to Kafka


pcasaes:
While testing Quarkus 1.6.0.CR1 I noticed a large performance hit while producing Kafka messages in comparison with 1.5.2.Final (smallrye reactive messaging 2.0.3 vs 2.1.0). The impact expressed itself as an order of magnitude increase in latency and CPU. 

I was able to track it down to this commit:
e439631487a76fdae84b7ae88cb0c3c59f613ef9

Lines 80-89 on KafkaSink.java
```
                return Uni.createFrom()
                        .<Void> emitter(
                                e -> stream.send((ProducerRecord) record, ar -> handleWriteResult(ar, message, record, e)))
                        .onFailure().retry()
                        .withBackOff(Duration.ofSeconds(1), Duration.ofSeconds(20)).atMost(config.getRetries())
                        .onFailure().recoverWithUni(t -> {
                            // Even with the retry we still fail, nack the message.
                            log.nackingMessage(message, actualTopic, t);
                            return Uni.createFrom().completionStage(message.nack(t));
                        });
```

I reverted just these lines locally and the performance penalty disappeared 
```
                return Uni.createFrom()
                        .emitter(e -> stream.write((ProducerRecord) record, ar -> handleWriteResult(ar, message, record, e)));
```

It is generally not recommended to retry producing Kafka messages. Doing so can lead to duplicate messages since the Kafka java library already retries indefinitely (by default).

I have Flamegraphs which I'll make available on the mailing group.

cescoffier:
These lines were added to handle the case where you don't have a Kafka broker ready (and so no topic in topology yet). It might be only necessary on startup.

Another solution is to introduce the long time overdue readiness check which will hold sending until we are ready.

cescoffier:
Fixed by #632

PR #632, opened 2020-06-28
#631 Memoize KafkaSink retries config and retry only if above 0

pcasaes:
@cescoffier 
This solves issue #631 

The problem was in the the `config.getRetries()` which is performing an expensive regular expression and it was doing this for every produced message. The solution here is to memo the retry value to a field and only apply the retry code if above 0.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
@@ -36,17 +36,17 @@ public class KafkaSink {
     private final int partition;
     private final String key;
     private final String topic;
-    private final KafkaConnectorOutgoingConfiguration config;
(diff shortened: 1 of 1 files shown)


pcasaes:
Thanks @kenfinnigan . I fixed an error in the commit message.

cescoffier:
Thanks! 

As it seems important enough, I'm going to cut a micro release with this fix.

pcasaes:
I took a look at the config code performing that regular expression
https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java

I did a quick test replacing the regexp with a simple for loop and string builder. Saw performance gains above 2x. Worth opening a PR for that?

Looking at the code also made me realize that setting the retry property and not relying on the default would mitigate #631

cescoffier:
Yes, go ahead!

pcasaes:
@kenfinnigan maybe you can take a look
https://github.com/smallrye/smallrye-config/pull/341

Noticed you were one of the last to touch that file.

PR #638, opened 2020-07-02
Provide the infrastructure to support health reporting

cescoffier:
Preliminary work to support health support.

What's not covered by this PR:

* the module exposing the reports to MP Metrics
* the update of the existing connectors to support readiness and liveness health checks

The implemented approach avoids having the provider and connectors depending on MP Health directly.

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/health/HealthReport.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/health/HealthReport.java
@@ -1,0 +1,147 @@
+package io.smallrye.reactive.messaging.health;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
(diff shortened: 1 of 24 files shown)


cescoffier:
When we did that for Metrics it's been a bit of a mess to mitigate the optionality in Quarkus (we had to update some bean class at build time to veto them). That "fix" would not work here. The approach I used avoids this, as we would have a new module exposing the MP Health from these internal types.

kenfinnigan:
And the new module would be in SRye RM or in Quarkus?

cescoffier:
Smallrye, it will be in SmallRye Reactive Messaging.

cescoffier:
let me add this module as part of this PR, so you can see the "whole" picture (minus the connector. part)

cescoffier:
@kenfinnigan I've added the MP Health support

PR #639, opened 2020-07-02
Fixes #629

### LINKED ISSUE #629: @Blocking prevents exception from being logged
With a method such as:

```
@Incoming("kafka-something")
@Blocking
public void onMessage(byte[] message) {
    throw new RuntimeException("Fakeception");
}
```

The exception results in the appropriate failure processing, but the actual exception isn't logged anywhere for subsequent investigation.

See https://github.com/quarkusio/quarkus/issues/10253 for full details

kenfinnigan:
- @Blocking and @Incoming method exceptions are now logged
- JUnit 4/5 co-existing now for unit tests
- Add SmallRye Testing to capture log output from a test

### DIFF
--- a/smallrye-reactive-messaging-provider/pom.xml
+++ b/smallrye-reactive-messaging-provider/pom.xml
@@ -11,7 +11,7 @@
   <artifactId>smallrye-reactive-messaging-provider</artifactId>
 
   <name>SmallRye Reactive Messaging : Provider</name>
-  
+
   <dependencies>
     <dependency>
       <groupId>io.smallrye.reactive</groupId>
@@ -30,7 +30,7 @@
       <artifactId>microprofile-metrics-api</artifactId>
       <scope>provided</scope>
     </dependency>
-    
+
     <dependency>
       <groupId>io.reactivex.rxjava2</groupId>
       <artifactId>rxjava</artifactId>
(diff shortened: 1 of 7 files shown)


smallrye-reactive-messaging-provider/pom.xml
```diff
@@ -53,6 +53,18 @@
...
+    <dependency>
+      <groupId>io.smallrye.testing</groupId>
+      <artifactId>smallrye-testing-utilities</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.logmanager</groupId>
+      <artifactId>jboss-logmanager</artifactId>
+      <version>2.1.17.Final</version>
```
cescoffier, line 65: Should we declare this version in the root pom? Maybe even the dependency in the dependency management?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/BlockingExceptionTest.java
```diff
@@ -0,0 +1,39 @@
...
+        addBeanClass(IncomingBlockingExceptionBean.class);
+        initialize();
+
+        IncomingBlockingExceptionBean bean = container.getBeanManager().createInstance()
+                .select(IncomingBlockingExceptionBean.class).get();
+
+        await().until(() -> bean.list().size() == 5);
+        assertThat(bean.list()).contains("a", "b", "d", "e", "f");
+
+        //TODO: Assert the log output contains an exception
```
cescoffier, line 32: Is it done?

smallrye-reactive-messaging-provider/pom.xml
kenfinnigan: Done

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/blocking/BlockingExceptionTest.java
kenfinnigan: Oops, removed

PR #646, opened 2020-07-06
Implement readiness and liveness health check support for the Kafka connector

cescoffier:
The Kafka connector reports the readiness and liveness of each channel managed by the connector.

### Readiness

On the inbound side (receiving records from Kafka), the readiness check verifies that:

* the broker is available
* the Kafka topic is created (available in the broker).
* no failures have been caught

On the outbound side (writing records to Kafka), the readiness check verifies that:

* the broker is available
* the Kafka topic is created (available in the broker).

### Liveness

On the inbound side (receiving records from Kafka), the liveness check verifies that:

* no failures have been caught
* the client is connected to the broker

On the outbound side (writing records to Kafka), the liveness check verifies that:

* no failures have been caught

Note that a message processing failures _nacks_ the message which is then handled by the failure-strategy.
It the responsibility of the failure-strategy to report the failure and influence the outcome of the liveness checks.
The `fail` failure strategy reports the failure and so the liveness check will report the failure.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
@@ -4,9 +4,8 @@ import static io.smallrye.reactive.messaging.kafka.i18n.KafkaExceptions.ex;
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
 import java.time.Duration;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
+import java.util.stream.Collectors;
 
 import javax.enterprise.inject.Instance;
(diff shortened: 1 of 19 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -187,4 +211,48 @@ public void closeQuietly() {
...
+            } else {
+                builder.add(configuration.getChannel(), true);
+            }
+        }
+
+        // If health is disable do not add anything to the builder.
+    }
+
+    private String getTopic() {
+        return configuration.getTopic().orElseGet(configuration::getChannel);
```
kenfinnigan, line 234: As this is now being used in at least a couple of places, should we be capturing the topic name instead of calling configuration more than once?

cescoffier: Yes, good idea! Just did it.

Issue #562, opened 2020-05-27
OpenTelemetry support for Kafka and AMQP


cescoffier:
This issue is about adding OpenTelemetry support (Trace / Span) to the Kafka and AMQP connectors both for the inbound and outbound direction. 

This issue should introduce a new type of metadata (`OpenTelemetryMetadata`), which would be created by the inbound Kafka and AMQP connector by reading the headers/annotations from the incoming record/message. 

On the outbound side, if the message produced by the application contains `OpenTelemetryMetadata,` the trace/span should be added to the sent record/message.

It should also be possible to create a new "empty" (initial) `OpenTelemetryMetadata` if the application is the source of the message. It should also be possible to customize the data if needed (like a method receiving the metadata and updating them).

CC @pavolloffay, @kenfinnigan

pavolloffay:
I like the intention to use OpenTelemetry instead OpenTracing. However before pulling this into the specification we have to create MP-OpenTelemetry project. I am not saying we cannot create POC in SR in the meantime.

kenfinnigan:
@pavolloffay, we're not putting this into the spec anytime soon.

We will implement it entirely in SmallRye and validate the APIs through proven usage before we propose anything to MicroProfile.

pavolloffay:
Sounds good :+1:

objectiser:
@kenfinnigan Would it be better to implement the OpenTelemetry support in the upstream kafka and amqp projects?

For example, the kafka instrumentation would also be of interest to the strimzi project.

kenfinnigan:
The instrumentation isn't being done within the client libraries of Kafka or AMQP, but in the reactive messaging connectors

objectiser:
Understand that - but just wondering if it would be better if it was done in the kafka/amqp client libs? Or is there something smallrye specific that would not be possible in client lib instrumentation?

kenfinnigan:
Ideally, I wouldn't want to have to make updates to a separate library for each reactive messaging connector.

It will take far longer to implement with that approach.

Maybe down the road we consider contributing things into those client libraries, but I don't see the benefit in starting there

hacst:
[...]

PR #659, opened 2020-07-16
Initial OpenTelemetry tracing for Kafka

kenfinnigan:
- Fixes Kafka part of #562

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -17,6 +17,10 @@ Type: _string_ | false |
 
 Type: _boolean_ | false | `true`
 
+| *tracing-enabled* | Whether tracing is enabled (default) or disabled
+
+Type: _boolean_ | false | `true`
+
 | *topics* | A comma-separating list of topics to be consumed. Cannot be used with the `topic` or `pattern` properties
 
 Type: _string_ | false | 
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,8 @@
 
     <jboss-log-manager.version>2.1.17.Final</jboss-log-manager.version>
 
+    <opentelemetry.version>0.6.0</opentelemetry.version>
+
     <sonar.coverage.jacoco.xmlReportPaths>target/site/jacoco/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
 
     <smallrye-reactive-converter-api.version>1.1.0</smallrye-reactive-converter-api.version>
@@ -212,6 +214,22 @@
         <groupId>org.jboss.logmanager</groupId>
         <artifactId>jboss-logmanager</artifactId>
         <version>${jboss-log-manager.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-api</artifactId>
+        <version>${opentelemetry.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-sdk</artifactId>
+        <version>${opentelemetry.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-exporters-inmemory</artifactId>
+        <version>${opentelemetry.version}</version>
       </dependency>
 
       <!-- From this project -->
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
 
     <jboss-log-manager.version>2.1.17.Final</jboss-log-manager.version>
 
-    <opentelemetry.version>0.6.0</opentelemetry.version>
+    <opentelemetry.version>0.7.1</opentelemetry.version>
 
     <sonar.coverage.jacoco.xmlReportPaths>target/site/jacoco/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
 
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
@@ -7,14 +7,14 @@ import java.util.function.Supplier;
 import org.apache.kafka.common.header.Headers;
 import org.eclipse.microprofile.reactive.messaging.Metadata;
 
-import io.smallrye.reactive.messaging.ce.CloudEventMetadata;
-import io.smallrye.reactive.messaging.kafka.commit.KafkaCommitHandler;
-import io.smallrye.reactive.messaging.kafka.fault.KafkaFailureHandler;
-import io.smallrye.reactive.messaging.kafka.impl.ce.KafkaCloudEventHelper;
 import io.grpc.Context;
 import io.opentelemetry.OpenTelemetry;
 import io.opentelemetry.trace.TracingContextUtils;
 import io.smallrye.reactive.messaging.TracingMetadata;
+import io.smallrye.reactive.messaging.ce.CloudEventMetadata;
+import io.smallrye.reactive.messaging.kafka.commit.KafkaCommitHandler;
+import io.smallrye.reactive.messaging.kafka.fault.KafkaFailureHandler;
+import io.smallrye.reactive.messaging.kafka.impl.ce.KafkaCloudEventHelper;
 import io.smallrye.reactive.messaging.kafka.tracing.HeaderExtractAdapter;
 import io.vertx.mutiny.kafka.client.consumer.KafkaConsumerRecord;
 
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
@@ -29,7 +29,8 @@ public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
     public IncomingKafkaRecord(KafkaConsumerRecord<K, T> record,
             KafkaCommitHandler commitHandler,
             KafkaFailureHandler onNack,
-            boolean cloudEventEnabled) {
+            boolean cloudEventEnabled,
+            boolean tracingEnabled) {
         this.commitHandler = commitHandler;
         this.kafkaMetadata = new IncomingKafkaRecordMetadata<>(record);
 
@@ -60,19 +61,20 @@ public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
             metadata = Metadata.of(this.kafkaMetadata);
         }
 
-        TracingMetadata tracingMetadata = TracingMetadata.empty();
-        if (record.headers() != null) {
-            // Read tracing headers
-            Context context = OpenTelemetry.getPropagators().getHttpTextFormat()
-                .extract(Context.current(), kafkaMetadata.getHeaders(), HeaderExtractAdapter.GETTER);
-            tracingMetadata = TracingMetadata
-                .withPrevious(TracingContextUtils.getSpanWithoutDefault(context));
+        if (tracingEnabled) {
+            TracingMetadata tracingMetadata = TracingMetadata.empty();
+            if (record.headers() != null) {
+                // Read tracing headers
+                Context context = OpenTelemetry.getPropagators().getHttpTextFormat()
+                        .extract(Context.current(), kafkaMetadata.getHeaders(), HeaderExtractAdapter.GETTER);
+                tracingMetadata = TracingMetadata
+                        .withPrevious(TracingContextUtils.getSpanWithoutDefault(context));
+            }
+
+            metadata = metadata.with(tracingMetadata);
         }
 
-        this.metadata = Metadata.of(this.kafkaMetadata, tracingMetadata);
-
         this.metadata = metadata;
-
         this.onNack = onNack;
         if (payload == null && !payloadSet) {
             this.payload = record.value();
@@ -135,7 +137,7 @@ public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
         return onNack.handle(this, reason);
     }
 
-    public void injectTracingMetadata(TracingMetadata tracingMetadata) {
+    public synchronized void injectTracingMetadata(TracingMetadata tracingMetadata) {
         metadata = metadata.with(tracingMetadata);
     }
 }
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
@@ -11,11 +11,16 @@ import io.smallrye.reactive.messaging.ce.CloudEventMetadata;
 import io.smallrye.reactive.messaging.kafka.commit.KafkaCommitHandler;
 import io.smallrye.reactive.messaging.kafka.fault.KafkaFailureHandler;
 import io.smallrye.reactive.messaging.kafka.impl.ce.KafkaCloudEventHelper;
+import io.grpc.Context;
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.trace.TracingContextUtils;
+import io.smallrye.reactive.messaging.TracingMetadata;
+import io.smallrye.reactive.messaging.kafka.tracing.HeaderExtractAdapter;
 import io.vertx.mutiny.kafka.client.consumer.KafkaConsumerRecord;
 
 public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
 
-    private final Metadata metadata;
+    private Metadata metadata;
     private final IncomingKafkaRecordMetadata<K, T> kafkaMetadata;
     private final KafkaCommitHandler commitHandler;
     private final KafkaFailureHandler onNack;
@@ -55,7 +60,19 @@ public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
             metadata = Metadata.of(this.kafkaMetadata);
         }
 
+        TracingMetadata tracingMetadata = TracingMetadata.empty();
+        if (record.headers() != null) {
+            // Read tracing headers
+            Context context = OpenTelemetry.getPropagators().getHttpTextFormat()
+                .extract(Context.current(), kafkaMetadata.getHeaders(), HeaderExtractAdapter.GETTER);
+            tracingMetadata = TracingMetadata
+                .withPrevious(TracingContextUtils.getSpanWithoutDefault(context));
+        }
+
+        this.metadata = Metadata.of(this.kafkaMetadata, tracingMetadata);
+
         this.metadata = metadata;
+
         this.onNack = onNack;
         if (payload == null && !payloadSet) {
             this.payload = record.value();
@@ -116,5 +133,9 @@ public class IncomingKafkaRecord<K, T> implements KafkaRecord<K, T> {
     @Override
     public CompletionStage<Void> nack(Throwable reason) {
         return onNack.handle(this, reason);
+    }
+
+    public void injectTracingMetadata(TracingMetadata tracingMetadata) {
+        metadata = metadata.with(tracingMetadata);
     }
 }
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
@@ -92,7 +92,7 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
 
     public static final String CONNECTOR_NAME = "smallrye-kafka";
 
-    public static Tracer TRACER = OpenTelemetry.getTracerProvider().get("io.smallrye.reactive.messaging.kafka");
+    public static final Tracer TRACER = OpenTelemetry.getTracerProvider().get("io.smallrye.reactive.messaging.kafka");
 
     @Inject
     ExecutionHolder executionHolder;
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
@@ -32,6 +32,8 @@ import org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams;
 import org.eclipse.microprofile.reactive.streams.operators.SubscriberBuilder;
 import org.reactivestreams.Publisher;
 
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.trace.Tracer;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.reactive.messaging.annotations.ConnectorAttribute;
 import io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction;
@@ -48,7 +50,7 @@ import io.vertx.mutiny.core.Vertx;
 @ConnectorAttribute(name = "bootstrap.servers", alias = "kafka.bootstrap.servers", type = "string", defaultValue = "localhost:9092", direction = Direction.INCOMING_AND_OUTGOING, description = "A comma-separated list of host:port to use for establishing the initial connection to the Kafka cluster.")
 @ConnectorAttribute(name = "topic", type = "string", direction = Direction.INCOMING_AND_OUTGOING, description = "The consumed / populated Kafka topic. If neither this property nor the `topics` properties are set, the channel name is used")
 @ConnectorAttribute(name = "health-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether health reporting is enabled (default) or disabled", defaultValue = "true")
-
+@ConnectorAttribute(name = "tracing-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether tracing is enabled (default) or disabled", defaultValue = "true")
 @ConnectorAttribute(name = "topics", type = "string", direction = Direction.INCOMING, description = "A comma-separating list of topics to be consumed. Cannot be used with the `topic` or `pattern` properties")
 @ConnectorAttribute(name = "pattern", type = "boolean", direction = Direction.INCOMING, description = "Indicate that the `topic` property is a regular expression. Must be used with the `topic` property. Cannot be used with the `topics` property", defaultValue = "false")
 @ConnectorAttribute(name = "key.deserializer", type = "string", direction = Direction.INCOMING, description = "The deserializer classname used to deserialize the record's key", defaultValue = "org.apache.kafka.common.serialization.StringDeserializer")
@@ -89,6 +91,8 @@ import io.vertx.mutiny.core.Vertx;
 public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnectorFactory, HealthReporter {
 
     public static final String CONNECTOR_NAME = "smallrye-kafka";
+
+    public static Tracer TRACER = OpenTelemetry.getTracerProvider().get("io.smallrye.reactive.messaging.kafka");
 
     @Inject
     ExecutionHolder executionHolder;
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
@@ -28,6 +28,7 @@ import io.opentelemetry.context.Scope;
 import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.SpanContext;
 import io.opentelemetry.trace.TracingContextUtils;
+import io.opentelemetry.trace.attributes.SemanticAttributes;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.subscription.UniEmitter;
 import io.smallrye.reactive.messaging.TracingMetadata;
@@ -252,9 +253,9 @@ public class KafkaSink {
 
             // Set Span attributes
             span.setAttribute("partition", partition);
-            span.setAttribute("messaging.system", "kafka");
-            span.setAttribute("messaging.destination", topic);
-            span.setAttribute("messaging.destination_kind", "topic");
+            SemanticAttributes.MESSAGING_SYSTEM.set(span, "kafka");
+            SemanticAttributes.MESSAGING_DESTINATION.set(span, topic);
+            SemanticAttributes.MESSAGING_DESTINATION_KIND.set(span, "topic");
 
             // Set span onto headers
             OpenTelemetry.getPropagators().getHttpTextFormat()
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
@@ -1,5 +1,6 @@
 package io.smallrye.reactive.messaging.kafka.impl;
 
+import static io.smallrye.reactive.messaging.kafka.KafkaConnector.TRACER;
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
 import java.time.Duration;
@@ -10,19 +11,31 @@ import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
+<<<<<<< HEAD
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.serialization.StringSerializer;
+=======
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.header.internals.RecordHeaders;
+>>>>>>> ced09d82... Initial OpenTelemetry work
 import org.eclipse.microprofile.reactive.messaging.Message;
 import org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams;
 import org.eclipse.microprofile.reactive.streams.operators.SubscriberBuilder;
 
+import io.grpc.Context;
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.SpanContext;
+import io.opentelemetry.trace.TracingContextUtils;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.subscription.UniEmitter;
+import io.smallrye.reactive.messaging.TracingMetadata;
 import io.smallrye.reactive.messaging.ce.OutgoingCloudEventMetadata;
 import io.smallrye.reactive.messaging.health.HealthReport;
 import io.smallrye.reactive.messaging.kafka.KafkaConnectorOutgoingConfiguration;
 import io.smallrye.reactive.messaging.kafka.OutgoingKafkaRecordMetadata;
 import io.smallrye.reactive.messaging.kafka.impl.ce.KafkaCloudEventHelper;
+import io.smallrye.reactive.messaging.kafka.tracing.HeaderInjectAdapter;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.json.JsonObject;
 import io.vertx.kafka.client.producer.KafkaWriteStream;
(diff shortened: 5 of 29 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
```diff
@@ -7,21 +7,36 @@
 import org.apache.kafka.common.header.Headers;
 import org.eclipse.microprofile.reactive.messaging.Metadata;
 
+import io.grpc.Context;
```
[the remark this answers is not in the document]
kenfinnigan: Yeah, OpenTelemetry using gRPC Context to pass things around.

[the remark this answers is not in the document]
cescoffier: We need to investigate how we can mitigate this in Quarkus. Depending on quarkus-grpc might be a bit too much.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
```diff
@@ -0,0 +1,359 @@
...
+import io.opentelemetry.trace.TracingContextUtils;
+import io.reactivex.Flowable;
+import io.smallrye.config.SmallRyeConfigProviderResolver;
+import io.smallrye.reactive.messaging.TracingMetadata;
+import io.smallrye.reactive.messaging.kafka.KafkaConnector;
+import io.smallrye.reactive.messaging.kafka.KafkaTestBase;
+import io.smallrye.reactive.messaging.kafka.KafkaUsage;
+import io.smallrye.reactive.messaging.kafka.MapBasedConfig;
+
+public class TracingPropagationTest extends KafkaTestBase {
```
pavolloffay, line 59: There are missing tests when an active span is created in the test an messages are produced and consumed. In this case the spans created by the instrumentation should by linked to the span in the test.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -158,6 +168,47 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
                 kafkaHeaders);
     }
 
+    private void createOutgoingTrace(Message<?> message, String topic, int partition, Headers headers) {
```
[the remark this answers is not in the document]
cescoffier: Yes, is it blocking?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -158,6 +168,47 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
...
+        if (configuration.getTracingEnabled()) {
+            Optional<TracingMetadata> tracingMetadata = TracingMetadata.fromMessage(message);
+
+            final Span.Builder spanBuilder = TRACER.spanBuilder(topic)
+                    .setSpanKind(Span.Kind.PRODUCER);
```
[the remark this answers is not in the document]
cescoffier: Partitions are important. They should be part of the trace (but optional)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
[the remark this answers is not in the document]
kenfinnigan: It's done right before creating the Kafka `ProducerRecord`, and I don't believe it's blocking

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -158,6 +168,47 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
                 kafkaHeaders);
     }
 
+    private void createOutgoingTrace(Message<?> message, String topic, int partition, Headers headers) {
+        if (configuration.getTracingEnabled()) {
+            Optional<TracingMetadata> tracingMetadata = TracingMetadata.fromMessage(message);
+
+            final Span.Builder spanBuilder = TRACER.spanBuilder(topic)
```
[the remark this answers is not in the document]
kenfinnigan: Why?

That's not what is recommended by current OpenTelemetry docs

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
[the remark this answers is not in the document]
kenfinnigan: On the outgoing trace I couldn't see a way to get the partitiion

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -158,6 +168,47 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
...
+                }
+            } else {
+                spanBuilder.setNoParent();
+            }
+
+            final Span span = spanBuilder.startSpan();
+            TracingContextUtils.currentContextWith(span);
+
+            // Set Span attributes
+            span.setAttribute("partition", partition);
```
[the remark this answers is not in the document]
kenfinnigan: In 0.5.0 they weren't there, I can use them now they're there

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -243,6 +248,36 @@ public synchronized void reportFailure(Throwable failure) {
...
+            // Set Span attributes
+            span.setAttribute("partition", kafkaRecord.getPartition());
+            span.setAttribute("offset", kafkaRecord.getOffset());
+            span.setAttribute("messaging.system", "kafka");
+            span.setAttribute("messaging.destination", kafkaRecord.getTopic());
+            span.setAttribute("messaging.destination_kind", "topic");
+
+            kafkaRecord.injectTracingMetadata(tracingMetadata.withSpan(span));
+
+            span.end();
```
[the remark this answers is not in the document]
kenfinnigan: The goal of the spans is not to track method execution, but simply to indicate a connector for `@Incoming` or `@Outgoing`.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderInjectAdapter.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.opentelemetry.context.propagation.HttpTextFormat;
+
+public class HeaderInjectAdapter implements HttpTextFormat.Setter<Headers> {
+    public static final HeaderInjectAdapter SETTER = new HeaderInjectAdapter();
+
+    @Override
+    public void set(Headers headers, String key, String value) {
+        if (headers != null) {
+            headers.remove(key).add(key, value.getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
kenfinnigan: Can't recall right now, but this is the pattern used in the Kafka instrumentation class

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaUsage.java
```diff
@@ -215,6 +220,26 @@ public void consumeStrings(String topicName, int count, long timeout, TimeUnit u
                 });
     }
 
+    public void consumeStringsWithTracing(String topicName, int count, long timeout, TimeUnit unit, Runnable completion,
+            BiConsumer<String, String> consumer, Consumer<Context> tracingConsumer) {
+        AtomicLong readCounter = new AtomicLong();
+        this.consumeStrings(this.continueIfNotExpired(() -> readCounter.get() < (long) count, timeout, unit), completion,
```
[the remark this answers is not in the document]
kenfinnigan: It counts up to the number of messages we expect to receive.

As for the nit, all the other methods use the same method

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
kenfinnigan: That's what `testFromKafkaToAppToKafka()` tests. Spans present on the message from Kafka that we receive into the app.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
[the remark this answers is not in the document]
pavolloffay: Where did you find the guidelines? 

The operation name should be low cardinality which topic name certainly is, but it should represent the operation wich in our case is either consume or produce.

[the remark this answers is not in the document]
pavolloffay: We should also add the topic name to tags. This will help with the data semantics and in aggregation jobs since we cannot rely on the operation name.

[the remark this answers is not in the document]
kenfinnigan: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md looks like it has been updated since last I saw, previously it only mentioned topic name.

Topic name is added to `messaging.destination`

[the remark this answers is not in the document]
pavolloffay: It might be the important fact here is to keep it low cardinality and then describe what it does. `produce` would be also good but including the topic name is even better.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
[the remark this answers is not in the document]
kenfinnigan: I'm a numpty, the partition is set further down as an attribute

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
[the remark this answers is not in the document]
kenfinnigan: Actually, they're only in `master` and not 0.6.0

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -243,6 +248,36 @@ public synchronized void reportFailure(Throwable failure) {
...
+            if (parentSpan != null && parentSpan.isValid()) {
+                spanBuilder.setParent(parentSpan);
+            } else {
+                spanBuilder.setNoParent();
+            }
+
+            final Span span = spanBuilder.startSpan();
+
+            // Set Span attributes
+            span.setAttribute("partition", kafkaRecord.getPartition());
```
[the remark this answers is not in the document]
kenfinnigan: Not released yet

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderInjectAdapter.java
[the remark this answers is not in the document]
kenfinnigan: The equals check on `RecordHeader` uses key and value, so I think it's necessary to remove the previous one using the key otherwise you could end up with different values as the underlying collection is a `List`.

Maybe I'm wrong, but probably better to be safe

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -158,6 +168,47 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
...
+    private void createOutgoingTrace(Message<?> message, String topic, int partition, Headers headers) {
+        if (configuration.getTracingEnabled()) {
+            Optional<TracingMetadata> tracingMetadata = TracingMetadata.fromMessage(message);
+
+            final Span.Builder spanBuilder = TRACER.spanBuilder(topic)
+                    .setSpanKind(Span.Kind.PRODUCER);
+
+            if (tracingMetadata.isPresent()) {
+                // Handle possible parent span
+                final SpanContext parentSpan = tracingMetadata.get().getPreviousSpanContext();
```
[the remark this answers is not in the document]
kenfinnigan: That's one of the reasons we didn't like the current Kafka instrumentation for OpenTelemetry: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/kafka-clients-0.11/src/main/java/io/opentelemetry/auto/instrumentation/kafkaclients/TracingIterator.java#L46

There's no guarantee that we process a method with `@Incoming` and then it's `@Outgoing` with such timing that there couldn't have been a different `@Incoming` being received before the other `@Outgoing` is processed, so relying on the context that's on the `TRACER` would lead to issues.

At least that's my understanding of things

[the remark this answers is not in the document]
kenfinnigan: `getPreviousSpanContext()` will return the `SpanContext`, if there is one, from any `@Incoming` that was processed

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
kenfinnigan: Ah, do you mean inside the method annotated with `@Incoming` or `@Outgoing` it creates a span?

kenfinnigan:
I have absolutely no idea why all the tests fail with ids not being in the expected set, even though it works fine locally

cescoffier:
I will look tomorrow.

kenfinnigan:
@pavolloffay any thoughts as to why it fails in CI but works locally?

pavolloffay:
I haven't looked, but it might be related to span reporting? E.g. the tests do no wait for spans being reported?

kenfinnigan:
What would be needed for that?

I presumed that waiting for the elements to be present in the output, and given it's using the log exporter, that the spans would be reported immediately.

pavolloffay:
Is it using log reporter or in-memory? Either way, it might be needed to wait for spans to be reported. I am usually using `Awaitility` lib to wait for a condition.

That said I am not sure why the tests are failing in this PR it might be bc of other reasons.

pavolloffay:
https://github.com/smallrye/smallrye-opentelemetry/pull/22/files#diff-9151ceec98feb657829d3d698bb7d7b1R109

kenfinnigan:
@pavolloffay @cescoffier This should be all good now, rebased and CI green

cescoffier:
Thanks I will have a look tomorrow.

Le mer. 19 août 2020 à 17:40, Ken Finnigan <notifications@github.com> a
écrit :

pom.xml
```diff
@@ -80,6 +80,8 @@
 
     <jboss-log-manager.version>2.1.17.Final</jboss-log-manager.version>
 
+    <opentelemetry.version>0.6.0</opentelemetry.version>
```
pavolloffay, line 83: Could we bump it to the latest version 0.7.0?

pavolloffay (commented):
nit use `SemanticAttributes` for attributes (I had a comment about it before) 

Previously squashed comment:

> Ah, do you mean inside the method annotated with @Incoming or @Outgoing it creates a span? 

The current span should be available in these methods. Do you create spans for each of these methods?

kenfinnigan:
I've updated to OTeL 0.7.1, and as part of that I was able to use some new `SemanticAttributes` that became available

kenfinnigan:
Rebased again

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
```diff
@@ -17,6 +17,10 @@ Type: _string_ | false |
 
 Type: _boolean_ | false | `true`
 
+| *tracing-enabled* | Whether tracing is enabled (default) or disabled
+
+Type: _boolean_ | false | `true`
```
cescoffier, line 22: In Quarkus, it might be a bit more complicated than just `true`. Does it require the tracing extension to be there?

pom.xml
```diff
@@ -214,6 +216,22 @@
...
+        <version>${opentelemetry.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-sdk</artifactId>
+        <version>${opentelemetry.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-exporters-inmemory</artifactId>
```
cescoffier, line 231: For my own understanding, it is a regular dependency or a test dependency?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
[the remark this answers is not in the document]
cescoffier: Any info about this?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
```diff
@@ -23,8 +28,18 @@ public IncomingKafkaRecord(
             KafkaCommitHandler commitHandler,
             KafkaFailureHandler onNack) {
         this.kafkaMetadata = new IncomingKafkaRecordMetadata<>(record);
-        this.metadata = Metadata.of(this.kafkaMetadata);
         this.commitHandler = commitHandler;
+
+        TracingMetadata tracingMetadata = TracingMetadata.empty();
```
cescoffier, line 33: Should it be added only if tracing is enabled?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
```diff
@@ -81,4 +96,8 @@ public Metadata getMetadata() {
     public CompletionStage<Void> nack(Throwable reason) {
         return onNack.handle(this, reason);
     }
+
+    public void injectTracingMetadata(TracingMetadata tracingMetadata) {
```
cescoffier, line 100: with metadata not final anymore, we must synchronize access.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -79,6 +82,8 @@
 
     public static final String CONNECTOR_NAME = "smallrye-kafka";
 
+    public static final Tracer TRACER = OpenTelemetry.getTracerProvider().get("io.smallrye.reactive.messaging.kafka");
```
cescoffier, line 85: so even if tracing is disabled we instantiate the tracer? Does it make sense?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -168,6 +180,48 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
                 kafkaHeaders);
     }
 
+    private void createOutgoingTrace(Message<?> message, String topic, int partition, Headers headers) {
+        if (configuration.getTracingEnabled()) {
+            Optional<TracingMetadata> tracingMetadata = TracingMetadata.fromMessage(message);
+
+            final Span.Builder spanBuilder = TRACER.spanBuilder(topic + " send")
```
cescoffier, line 187: is the "send" word defined somewhere? In Kafka, we use "write" most of the time.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -168,6 +180,48 @@ private void handleWriteResult(AsyncResult<?> ar, Message<?> message, ProducerRe
                 kafkaHeaders);
     }
 
+    private void createOutgoingTrace(Message<?> message, String topic, int partition, Headers headers) {
+        if (configuration.getTracingEnabled()) {
```
cescoffier, line 184: This whole block should be externalized in a carefully tested method.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -225,6 +230,7 @@ public KafkaSource(Vertx vertx,
                     }
                 })
                 .map(rec -> commitHandler.received(new IncomingKafkaRecord<>(rec, commitHandler, failureHandler)))
+                .onItem().invoke(this::incomingTrace)
```
cescoffier, line 233: let's build a conditional pipeline instead of testing if tracing is enabled every time.

```
multi = ...
if (getTracingEnabled()) {
    multi = .... 
}

File: smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (cescoffier) [reply smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java:298]: We could still register the reception timestamp. 
Note that it's not the record timestamp.

File: smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderExtractAdapter.java [hunk 21]
```diff
@@ -0,0 +1,21 @@
...
+import org.apache.kafka.common.header.Headers;
+
+import io.opentelemetry.context.propagation.HttpTextFormat;
+
+public class HeaderExtractAdapter implements HttpTextFormat.Getter<Headers> {
+    public static final HeaderExtractAdapter GETTER = new HeaderExtractAdapter();
+
+    @Override
+    public String get(Headers headers, String key) {
+        final Header header = headers.lastHeader(key);
```
cescoffier, line 15: should it fail on multiple values? Is it even possible?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderInjectAdapter.java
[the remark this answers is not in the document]
cescoffier: I was thinking add was appending, but let's check.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/DefaultConfigTest.java
```diff
@@ -89,12 +89,14 @@ private MapBasedConfig getKafkaSinkConfigForMyAppProcessingData(String topicOut,
         Map<String, Object> config = new HashMap<>();
         config.put(prefix + "connector", KafkaConnector.CONNECTOR_NAME);
         config.put(prefix + "topic", topicOut);
+        config.put(prefix + "tracing-enabled", false);
```
cescoffier, line 92: That is a bit cumbersome and error-prone.
Is there something we can do in the base class?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaTestBase.java
```diff
@@ -66,7 +66,9 @@ public void setup() {
 
     @After
     public void tearDown() {
-        vertx.close();
+        if (vertx != null) {
+            vertx.close();
```
cescoffier, line 70: we should wait until the vert.x instance is closed if we want to be totally correct.

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
```diff
@@ -0,0 +1,392 @@
...
+        @Incoming("source")
+        @Outgoing("kafka")
+        public Message<Integer> processMessage(Message<Integer> input) {
+            return input.withPayload(input.getPayload() + 1);
+        }
+    }
+
+    @ApplicationScoped
+    public static class MyAppReceivingData {
+        private List<TracingMetadata> tracingMetadata = new ArrayList<>();
```
cescoffier, line 373: minor - can be final

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
```diff
@@ -0,0 +1,392 @@
...
+        private List<Integer> results = new CopyOnWriteArrayList<>();
```
cescoffier, line 374: minor - can be final

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
kenfinnigan: No, without an implementation of the API present, it's basically a No-Op

pom.xml
kenfinnigan: Testing only

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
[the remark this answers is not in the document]
kenfinnigan: It's only the grpc Context that's needed, not the entirety of grpc

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
kenfinnigan: Probably, will adjust

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
kenfinnigan: do you mean just making the method synchronized, or something different?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
kenfinnigan: There's a small cost yes, even to have the No Op tracer.

However, not setting it like this then means more work to ensure we don't create multiple versions later.

I don't mind doing that, but that would potentially slow things down a bit on the first use

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
kenfinnigan: Not sure what you mean. It's all in `createOutgoingTrace` method, are you suggesting something else?

Or do you mean there should be unit tests verifying the behavior of just this method?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
kenfinnigan: It's defined by OpenTelemetry specification. See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#conventions

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
kenfinnigan: Ok

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderExtractAdapter.java
kenfinnigan: I think it's specified to only retrieve the last one

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/tracing/HeaderInjectAdapter.java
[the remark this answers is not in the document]
kenfinnigan: Yes add is appending, so we need to remove the previous value to ensure we don't end up with multiples present

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/DefaultConfigTest.java
kenfinnigan: Will see if something can be done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
kenfinnigan: How do I do the `onItem()` so that it's invoked at the correct place in the stream?

Do you mean creating two pipelines, one with and one without the `onItem()`? Or do you mean multi into multi?

pom.xml
cescoffier: @kenfinnigan I believe this was done, no?

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
cescoffier: Perfect!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecord.java
cescoffier: yes, `synchronized`

pom.xml
kenfinnigan: Yes

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
cescoffier: ok, make sense.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
cescoffier: great, don't forget to add this to the documentation.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
cescoffier: Damned missed it.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
cescoffier: Pipelines are "append-only".

You create one, and append the `onItem()` if tracing is enabled:

```java
multi = ...
if (getTracingEnabled()) {
    multi = multi.onItem().invoke(this::incomingTrace); 
}
multi = multi....
```

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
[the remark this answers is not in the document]
kenfinnigan: There doesn't appear to be anything defined to represent that in OTeL at present.

Do you have a suggested name?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/DefaultConfigTest.java
kenfinnigan: I was thinking I could detect whether it was set in `MapBasedConfig` and then add it set to false in `write()` if it wasn't.

But I'm not sure that's going to be easy given the different key prefixes and an unknown number of them.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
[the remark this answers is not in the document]
cescoffier: @pavolloffay any hint?

cescoffier (approved):
LGTM! 

@pavolloffay is it ok for you?

For the record, @kenfinnigan did a few tests in Quarkus to check the behavior. This PR didn't break anything :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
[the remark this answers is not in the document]
pavolloffay: I am not sure I follow you here.

Do you want to record the reception timestamp as the span timestamp? The span start time should be overridable, if the timing you want to record is after the span start I suggest to add a timed event to the span.

kenfinnigan:
Any updates @pavolloffay? I'd like to get this merged

pavolloffay:
I have responsed to @cescoffier in https://github.com/smallrye/smallrye-reactive-messaging/pull/659/files#r478926384. 

I will have another look at the PR

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -264,6 +275,36 @@ public synchronized void reportFailure(Throwable failure) {
         failures.add(failure);
     }
 
+    public void incomingTrace(IncomingKafkaRecord<K, V> kafkaRecord) {
+        if (configuration.getTracingEnabled()) {
```
pavolloffay, line 279: The signature looks weird `getTracingEnabled()` it could be `.isTracingEnabled()`?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/tracing/TracingPropagationTest.java
```diff
@@ -0,0 +1,390 @@
...
+        if (container != null) {
+            container.close();
+        }
+
+        // Release the config objects
+        SmallRyeConfigProviderResolver.instance().releaseConfig(ConfigProvider.getConfig());
+    }
+
+    @Test
+    public void testFromAppToKafka() {
```
pavolloffay, line 90: Help me understand what this test does. 

The test class sends messages from producer to consumer. The span context is injected and extracted by the test class so what does it test? I am probably mising something here.

pavolloffay (approved):
LGTM

The tests are bit too hard to read more comments would help there.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
kenfinnigan: I think that's just how the methods are generated, as it's not manually coded

PR #672, opened 2020-07-20
Set fetch size to before pause value when resuming kafka source

pcasaes:
@cescoffier 
This is a doozy and I'm not this sure this fix is the correct way to go. When we have a re-balance listener configured and `partitions` is greater than 1 we will get the following if consuming messages at a high throughput (KafkaSource will then stop consuming):

```
2020-07-20 03:48:12,061 ERROR [io.sma.rea.mes.provider] (vert.x-eventloop-thread-1) SRMSG00201: Error caught during the stream processing: io.smallrye.mutiny.subscription.BackPressureFailure: Buffer full, cannot emit item
	at io.smallrye.mutiny.operators.multi.MultiFlatMapOp$FlatMapMainSubscriber.failOverflow(MultiFlatMapOp.java:539)
	at io.smallrye.mutiny.operators.multi.MultiFlatMapOp$FlatMapMainSubscriber.tryEmit(MultiFlatMapOp.java:240)
	at io.smallrye.mutiny.operators.multi.MultiFlatMapOp$FlatMapInner.onItem(MultiFlatMapOp.java:595)
	at io.smallrye.mutiny.subscription.MultiSubscriber.onNext(MultiSubscriber.java:61)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.helpers.HalfSerializer.onNext(HalfSerializer.java:31)
	at io.smallrye.mutiny.helpers.StrictMultiSubscriber.onItem(StrictMultiSubscriber.java:81)
	at io.smallrye.mutiny.operators.multi.MultiMapOp$MapProcessor.onItem(MultiMapOp.java:50)
	at io.smallrye.mutiny.subscription.MultiSubscriber.onNext(MultiSubscriber.java:61)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.helpers.HalfSerializer.onNext(HalfSerializer.java:31)
	at io.smallrye.mutiny.helpers.StrictMultiSubscriber.onItem(StrictMultiSubscriber.java:81)
	at io.smallrye.mutiny.operators.multi.MultiSignalConsumerOp$SignalSubscriber.onItem(MultiSignalConsumerOp.java:161)
	at io.smallrye.mutiny.subscription.MultiSubscriber.onNext(MultiSubscriber.java:61)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.helpers.HalfSerializer.onNext(HalfSerializer.java:31)
	at io.smallrye.mutiny.helpers.StrictMultiSubscriber.onItem(StrictMultiSubscriber.java:81)
	at io.smallrye.mutiny.operators.multi.MultiSignalConsumerOp$SignalSubscriber.onItem(MultiSignalConsumerOp.java:161)
	at io.smallrye.mutiny.subscription.MultiSubscriber.onNext(MultiSubscriber.java:61)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.lambda$onNext$1(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:217)
	at io.smallrye.mutiny.context.ContextPropagationMultiInterceptor$1.onNext(ContextPropagationMultiInterceptor.java:36)
	at io.smallrye.mutiny.vertx.MultiReadStream.lambda$subscribe$2(MultiReadStream.java:76)
	at io.vertx.kafka.client.consumer.impl.KafkaConsumerImpl.lambda$handler$1(KafkaConsumerImpl.java:79)
	at io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl.run(KafkaReadStreamImpl.java:229)
	at io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl.lambda$schedule$8(KafkaReadStreamImpl.java:184)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:366)
	at io.vertx.core.impl.EventLoopContext.lambda$executeAsync$0(EventLoopContext.java:38)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:500)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)

```

This happens because when resuming the Vert.x consumer it set sets the fetch amount to Long.MAX_VALUE
https://github.com/vert-x3/vertx-kafka-client/blob/3.9.1/src/main/java/io/vertx/kafka/client/consumer/impl/KafkaReadStreamImpl.java#L571

There's also a less severe side effect when `partitions` is 1. Because of the `MAX_VALUE` fetch amount it will consume more than 1 message concurrently when using a worker thread to perform blocking operations, ex:

```
    @Incoming("topic-in")
    public CompletionStage<Void> consume(IncomingKafkaRecord<String, String> record) {
        return CompletableFuture.runAsync(() -> {
            // perform blocking code.
        });
    }
```

Without a re-balance listener it will process the messages one at a time (`MultiFlapMapOp#drainLoop` always ends up requesting 1 in this case). Configuring a consumer re-balance listener shouldn't affect this behavior.

When `partitions` is 1 the the fetch amount should be 1 and when greater than 1 the amount should be 128. This resolves both issues.

What I don't like about this fix is that I arrived at these values empirically. It seems like this bug resides in the Vert.x Kafka code. Maybe resume should be setting the fetch amount to the previous non zero fetch amount before pausing. Something like:

```
  @Override
  public KafkaReadStreamImpl<K, V> resume() {
    return fetch(this.previousFetchAmountBeforePausing.get());
  }

```

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
@@ -150,6 +150,7 @@ public class KafkaSource<K, V> {
                     + 11_000L; // it's possible that it might expire 10 seconds before when we need it to
 
             kafkaConsumer.partitionsAssignedHandler(set -> {
+                final long currentDemand = kafkaConsumer.demand();
                 kafkaConsumer.pause();
                 log.executingConsumerAssignedRebalanceListener(group);
                 listener.onPartitionsAssigned(kafkaConsumer, set)
@@ -161,12 +162,12 @@ public class KafkaSource<K, V> {
                                 a -> {
                                     log.executedConsumerAssignedRebalanceListener(group);
                                     commitHandler.partitionsAssigned(vertx.getOrCreateContext(), set);
-                                    kafkaConsumer.resume();
+                                    kafkaConsumer.fetch(currentDemand);
                                 },
                                 t -> {
                                     log.reEnablingConsumerforGroup(group);
                                     commitHandler.partitionsAssigned(vertx.getOrCreateContext(), set);
-                                    kafkaConsumer.resume();
+                                    kafkaConsumer.fetch(currentDemand);
                                 });
             });
 


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -129,6 +129,10 @@ public KafkaSource(Vertx vertx,
                                                     "10000")))
                             + 11_000L; // it's possible that it might expire 10 seconds before when we need it to
 
+                    // When resuming the consumer we must set the fetch amount to what it was before pausing
+                    // When partitions is greater than 1 this value will be 128 otherwise it will be 1.
+                    final int resumeFetchAmount = config.getPartitions() > 1 ? 128 : 1;
```
cescoffier, line 134: Why 128?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -140,11 +144,11 @@ public KafkaSource(Vertx vertx,
                                 .with(
                                         a -> {
                                             log.executedConsumerAssignedRebalanceListener(group);
-                                            kafkaConsumer.resume();
+                                            kafkaConsumer.fetch(resumeFetchAmount);
```
cescoffier, line 147: What happens when the amount is consumed? Would it get more?

cescoffier:
We need some tests around this.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: When `partitions` is greater then 1 the initial fetch is the `concurrency` value from `MultiMerge` which by default is 128.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: It will get more. That's the current behavior without the consumer re-balance.

pcasaes:
If we'll pursue this fix I can write some tests. Basically something that replicates the BackPressureFailure as well as order of messages. But frankly fixing this one the Vert.x client seems like a more elegant solution. What do you think?

cescoffier:
Yes, I believe it's more a Vert.x Kafka Client issue.

pcasaes:
I opened a discussion on the Vert.x forum. Hopefully we'll get some good feedback.

pcasaes:
@cescoffier we got some work done the vertx kafka client
https://github.com/vert-x3/vertx-kafka-client/pull/180

Once 3.9.3 or 4.0.0 is out and integrated we can resolve this.

cescoffier:
Great!

Le ven. 24 juil. 2020 à 16:06, Paulo Casaes <notifications@github.com> a
écrit :

cescoffier (dismissed):

cescoffier:
@pcasaes Just checking - are we all good about this?

pcasaes:
We have to wait for vertx 3.9.3 or 4.0.0

pcasaes:
@cescoffier this is resolved

cescoffier:
Thanks, I will have a look.

I'm going to do a large refactoring of the connector as there are misusages (incorrect) of the kafka protocol.

pcasaes:
What are you thinking?

cescoffier:
I have been noticing issues on _large_ Kafka clusters. By looking into them, I realized some misconceptions.

First (and thank you for this), I'm going to make the `throttle` commit strategy the default, and disable auto-commit by default. 

I'm chasing a few non-closed timers and clients also. While this is generally not an issue in a real application, it's kind of a bad smell. I believe I got most of them yesterday. 

Then, there are a few issues regarding rebalancing. How the vert.x client handle rebalancing makes it impossible to use the synchronization barrier from the iterative rebalance protocol appropriately. When the protocol invokes the revoked partition callback, the callback must block the polling thread (which means it must be called from the polling thread). Once the callback returns, it sends a message to the leader to indicate that the revocation has been handled correctly. The callback must commit the latest offset _synchronously_ (commitSync), before returning. So there is a bit of change to do in this area. The same applies when the application is stopped (which could be considered as a kind of rebalance where all your partitions get revoked). 

At the moment, you can create multiple consumers to handle multiple partitions. These consumers are new _clients_. However, there is a better way to handle this case. Typically, it is possible to implement multi-threaded consumers that would remove the need to create multiple clients. The problem with multiple clients is the complexity of the metrics (multiple clients means multiple ids). 

Finally, in this effort to reduce the number of clients, we could reuse the same client on the _sink_ and _source_, if there are connected to the same broker (and share the same `client.id`). 

That give you some ideas. 

Testing is also something that needs to be done more rigorously (even if at the moment the tests already take 5 minutes on my machine).

pcasaes:
Nice! count on me for help.

> First (and thank you for this), I'm going to make the throttle commit strategy the default, and disable auto-commit by default.

I think this is a great idea but last night I noticed two issues with it. 

Minor issue: It only triggers a commit if there are incoming messages. Shouldn't be a problem in a real world situation. If it's EASY to solve we can fix it, otherwise I don't think it's worth the risk.

Major issue: 
> If too many received messages are not acked then the KafkaSource will be marked as unhealthy. Too many is defined as a power of two value greater than or equal to max.poll.records times 2 (default is 1024). This strategy guarantees at-least-once delivery even if the channel performs asynchronous processing.

So max.poll.records * 2 isn't good enough. The consumer can easily consume at a higher rate (1024 per 5 seconds), especially if you are replaying topics. We'd have to give this a good think. My original idea was to warn of situations where a poison record (one never acked or nacked) would leak to OOM. An EASY fix would be to change the value to one that is impossibly high but not so much that it would still lead to an OOM (might not even be possible with Kafka's killer throughput). I think we could do better though.

Side note: it would be nice to be able to configure the 5 second commit period value, maybe even use kafka's conf property for this.

> Then, there are a few issues regarding rebalancing. How the vert.x client handle rebalancing makes it impossible to use the synchronization barrier from the iterative rebalance protocol appropriately. When the protocol invokes the revoked partition callback, the callback must block the polling thread (which means it must be called from the polling thread). Once the callback returns, it sends a message to the leader to indicate that the revocation has been handled correctly. The callback must commit the latest offset synchronously (commitSync), before returning. So there is a bit of change to do in this area. The same applies when the application is stopped (which could be considered as a kind of rebalance where all your partitions get revoked).

Yes!!! This definitely irked me when I was working on this. Not sure how to do this without introducing changes to the vert.x client. I'm pretty sure that I tried using **commitSync** but it hanged the client :(

> At the moment, you can create multiple consumers to handle multiple partitions. These consumers are new clients. However, there is a better way to handle this case. Typically, it is possible to implement multi-threaded consumers that would remove the need to create multiple clients. The problem with multiple clients is the complexity of the metrics (multiple clients means multiple ids).

This sounds nice. It's actually been my experience that even with multiple clients the connector still only processes one message at a time for all clients.

> Finally, in this effort to reduce the number of clients, we could reuse the same client on the sink and source, if there are connected to the same broker (and share the same client.id).

This would be very nice.

cescoffier:
> Nice! count on me for help.

I will for sure!

> 
> > First (and thank you for this), I'm going to make the throttle commit strategy the default, and disable auto-commit by default.
> 
> I think this is a great idea but last night I noticed two issues with it.
> 
> Minor issue: It only triggers a commit if there are incoming messages. Shouldn't be a problem in a real world situation. If it's EASY to solve we can fix it, otherwise I don't think it's worth the risk.
> 

Yes, I've seen the issue. It mimick the `poll` mechanism from Kafka, and so is triggered when there is a message. I was thinking about changing this using a Vert.x timer (basically a scheduled task), which every x second (configurable) would commit the offset. 

> Major issue:
> 
> > If too many received messages are not acked then the KafkaSource will be marked as unhealthy. Too many is defined as a power of two value greater than or equal to max.poll.records times 2 (default is 1024). This strategy guarantees at-least-once delivery even if the channel performs asynchronous processing.
> 
> So max.poll.records * 2 isn't good enough. The consumer can easily consume at a higher rate (1024 per 5 seconds), especially if you are replaying topics. We'd have to give this a good think. My original idea was to warn of situations where a poison record (one never acked or nacked) would leak to OOM. An EASY fix would be to change the value to one that is impossibly high but not so much that it would still lead to an OOM (might not even be possible with Kafka's killer throughput). I think we could do better though.
> 

First the timer approach would mitigate a bit the issue. Now, we still have the issue with the poison pill which means that we won't commit any new message because one is missing. We can implement a timeout that consider that a record not acked in x seconds (also configuration) is nacked. I've seen the Kafka client using that approach too (if the period between two pools is too long). 


> Side note: it would be nice to be able to configure the 5 second commit period value, maybe even use kafka's conf property for this.

Definitely.

> 
> > Then, there are a few issues regarding rebalancing. How the vert.x client handle rebalancing makes it impossible to use the synchronization barrier from the iterative rebalance protocol appropriately. When the protocol invokes the revoked partition callback, the callback must block the polling thread (which means it must be called from the polling thread). Once the callback returns, it sends a message to the leader to indicate that the revocation has been handled correctly. The callback must commit the latest offset synchronously (commitSync), before returning. So there is a bit of change to do in this area. The same applies when the application is stopped (which could be considered as a kind of rebalance where all your partitions get revoked).
> 
> Yes!!! This definitely irked me when I was working on this. Not sure how to do this without introducing changes to the vert.x client. I'm pretty sure that I tried using **commitSync** but it hanged the client :(

Because you didn't find the _unwrap_ method :-) 
Yes, there are issues in the Vert.x Kafka client, and the Vert.x team is aware (I'm working with them). We will provide a new API for it, but first I need to better understand how to make it great. So, for now, we will, when needed, extract the underlying client and use the regular Kafka API.

> 
> > At the moment, you can create multiple consumers to handle multiple partitions. These consumers are new clients. However, there is a better way to handle this case. Typically, it is possible to implement multi-threaded consumers that would remove the need to create multiple clients. The problem with multiple clients is the complexity of the metrics (multiple clients means multiple ids).
> 
> This sounds nice. It's actually been my experience that even with multiple clients the connector still only processes one message at a time for all clients.

Even with _partitions_ set to more than 1? 

> 
> > Finally, in this effort to reduce the number of clients, we could reuse the same client on the sink and source, if there are connected to the same broker (and share the same client.id).
> 
> This would be very nice.

Yes, definitely something that will help observability.

pcasaes:
> > Nice! count on me for help.
> 
> I will for sure!
> 
> > > First (and thank you for this), I'm going to make the throttle commit strategy the default, and disable auto-commit by default.
> > 
> > 
> > I think this is a great idea but last night I noticed two issues with it.
> > Minor issue: It only triggers a commit if there are incoming messages. Shouldn't be a problem in a real world situation. If it's EASY to solve we can fix it, otherwise I don't think it's worth the risk.
> 
> Yes, I've seen the issue. It mimick the `poll` mechanism from Kafka, and so is triggered when there is a message. I was thinking about changing this using a Vert.x timer (basically a scheduled task), which every x second (configurable) would commit the offset.

Seems simple enough.  So acking a message will simply store it and a periodic task then does the committing, all in the same vert.x context.

> 
> > Major issue:
> > > If too many received messages are not acked then the KafkaSource will be marked as unhealthy. Too many is defined as a power of two value greater than or equal to max.poll.records times 2 (default is 1024). This strategy guarantees at-least-once delivery even if the channel performs asynchronous processing.
> > 
> > 
> > So max.poll.records * 2 isn't good enough. The consumer can easily consume at a higher rate (1024 per 5 seconds), especially if you are replaying topics. We'd have to give this a good think. My original idea was to warn of situations where a poison record (one never acked or nacked) would leak to OOM. An EASY fix would be to change the value to one that is impossibly high but not so much that it would still lead to an OOM (might not even be possible with Kafka's killer throughput). I think we could do better though.
> 
> First the timer approach would mitigate a bit the issue. Now, we still have the issue with the poison pill which means that we won't commit any new message because one is missing. We can implement a timeout that consider that a record not acked in x seconds (also configuration) is nacked. I've seen the Kafka client using that approach too (if the period between two pools is too long).

Much better! We can keep the received timestamp of each message and check the oldest timestamp for defining a healthy state.

> 
> > Side note: it would be nice to be able to configure the 5 second commit period value, maybe even use kafka's conf property for this.
> 
> Definitely.

Easy to do.

> 
> > > Then, there are a few issues regarding rebalancing. How the vert.x client handle rebalancing makes it impossible to use the synchronization barrier from the iterative rebalance protocol appropriately. When the protocol invokes the revoked partition callback, the callback must block the polling thread (which means it must be called from the polling thread). Once the callback returns, it sends a message to the leader to indicate that the revocation has been handled correctly. The callback must commit the latest offset synchronously (commitSync), before returning. So there is a bit of change to do in this area. The same applies when the application is stopped (which could be considered as a kind of rebalance where all your partitions get revoked).
> > 
> > 
> > Yes!!! This definitely irked me when I was working on this. Not sure how to do this without introducing changes to the vert.x client. I'm pretty sure that I tried using **commitSync** but it hanged the client :(
> 
> Because you didn't find the _unwrap_ method :-)
> Yes, there are issues in the Vert.x Kafka client, and the Vert.x team is aware (I'm working with them). We will provide a new API for it, but first I need to better understand how to make it great. So, for now, we will, when needed, extract the underlying client and use the regular Kafka API.

Cool!

> 
> > > At the moment, you can create multiple consumers to handle multiple partitions. These consumers are new clients. However, there is a better way to handle this case. Typically, it is possible to implement multi-threaded consumers that would remove the need to create multiple clients. The problem with multiple clients is the complexity of the metrics (multiple clients means multiple ids).
> > 
> > 
> > This sounds nice. It's actually been my experience that even with multiple clients the connector still only processes one message at a time for all clients.
> 
> Even with _partitions_ set to more than 1?

Yes. I set the topic partition to 2 and the connect partition to 2 as well. I believe you are using a Multi merge to join both streams (might be wrong). Each client does get its own vert.x event thread but the connector will only offer a next record after getting an ack from the previously offered record irrespective of client. I was able to confirm this with strategic debugging.

This can be worked around by configuring NONE as an Ack strategy, processing on a worker executor, returning CompletableFuture.complete(null) on the main thread and acking in the worker thread.

> 
> > > Finally, in this effort to reduce the number of clients, we could reuse the same client on the sink and source, if there are connected to the same broker (and share the same client.id).
> > 
> > 
> > This would be very nice.
> 
> Yes, definitely something that will help observability.

PR #680, opened 2020-07-27
Add metrics to the base scope

Azquelt:
Create metrics defined by the spec in the base scope, rather than the
application scope.

To match eclipse/microprofile-reactive-messaging#93


cescoffier:
@jmartisk can you double-check?

jmartisk (approved):

jmartisk:
LGTM

PR #687, opened 2020-08-02
Add commit strategies to support at least once delivery

pcasaes:
@cescoffier Here's a preliminary commit. As we discussed if **enable.auto.commit** is true then ACK will be a NOOP, otherwise we'll keep track of which offsets we should commit to and do so every 5 seconds. 

I'll have to give some thoughts on automated tests. Manually I tested a simple consumer against two partitions and would start and top a second process. This would force a re-balance.


pcasaes:
This won't always work. In most cases it will but if replaying from compacted logs we won't get monotonically sequential offsets. We'll have to queue up records as they are created in the event loop. This way we will get them in order.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -0,0 +1,143 @@
...
+        resetNextCommitTime();
+    }
+
+    @Override
+    public <K, V> IncomingKafkaRecord<K, V> received(IncomingKafkaRecord<K, V> record) {
+        TopicPartition recordsTopicPartition = getTopicPartition(record);
+        getReceivedOffsetQueue(recordsTopicPartition).offer(record.getOffset());
+
+        return record;
+    }
```
pcasaes, line 69: This solves the issue with compacted logs. Tested and working well, but it creates a new problem. 

The user might decide to use the **NONE** Acknowledgement Strategy together with **enable.auto.commit=false**.
https://smallrye.io/smallrye-reactive-messaging/2.1.1/apidocs/org/eclipse/microprofile/reactive/messaging/Acknowledgment.Strategy.html#NONE

 Basically the user wishes to never commit (there's a few use cases for this).

But with the implementation as is we will spring a leak. The received offset queue will grow without being emptied since there will be no call to **ack**. It looks to me that the Incoming publishers and connectors are decoupled in such a way that the connector doesn't know what strategy is configured.

I see two possible solutions:

1. Have the commit strategy be configurable like the nack strategy. Have the default depend on **enable.auto.commit** but let the user configure _ignore_ if they wish.

2. If we have more than **max.poll.records** (default 500) received messages without committing a single one then we can assume that the user configured to never commit. In this case we can just turn the strategy off.

1 is much simpler to maintain, but is a bit onerous on the user and requires proper documentation.
2 requires no action from the user (should just work), but is a harder to reason and maintain.


What do you think? If there was some way for the connector to know that **NONE** was used during setup that would be great, but I'm not seeing it.

cescoffier:
> This won't always work. In most cases, it will but if replaying from compacted logs we won't get monotonically sequential offsets. We'll have to queue up records as they are created in the event loop. This way we will get them in order.

You are absolutely right. My explanation was simplifying a bit. We need to track the received offsets.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: (1) definitely. We can use "enable.auto.commit" to determine the default. Adding another attribute to configure the strategy would be nice and allow configuring the application to use the current model. 

That's what I wrote on the SmallRye Mailing list (for traceability purpose):

> To implement this, we would need a new attribute to set the strategy:
> * commit every offset when the message is acked -> Today's solution
> * let kafka auto-commit -> enable the auto-commit and ignore acks from the app
> * batch/cursor tracking strategy -> the strategy described in this mail (which would need a > better name)
> Each of them has pros and cons, so they would need to be clearly documented.

pcasaes:
@cescoffier Should be done. Next step is documentation and some tests. I had to change the DLQ Nack to ack the message after producing to the DLQ. I think it would also be interesting to add a health check on the throttled commit strategy. If the amount of unacked messages becomes **too high** that's definitely not health.

pcasaes:
@cescoffier Documentation is done. Also added a health check to the _throttled_ strategy.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -111,52 +124,59 @@ public KafkaSource(Vertx vertx,
...
+                        .subscribe()
+                        .with(
+                                a -> {
+                                    log.executedConsumerAssignedRebalanceListener(group);
+                                    commitHandler.partitionsAssigned(vertx.getOrCreateContext(), set);
+                                    kafkaConsumer.resume();
+                                },
+                                t -> {
+                                    log.reEnablingConsumerforGroup(group);
+                                    commitHandler.partitionsAssigned(vertx.getOrCreateContext(), set);
```
pcasaes, line 163: Line 158 and 163 are the only actual changes. We have to inform the commit handler that a re-balance took place so it can reset its internal data structurers.

cescoffier:
Are we good to go with this one?

pcasaes:
Unless you see any issues I'm good with it.

cescoffier:
Thanks again!

PR #699, opened 2020-08-17
Added support to enclose the channel name with double quotes

### LINKED ISSUE #695: Channel name with dot is not supported
I'm trying to use the AMQP connector to consume a channel. 

But the channel was created with several dots in the name and I didn't find a way to make it work.

Looking at the code seems that it won't really work in the current implementation, as it removes the part before the first dot in the name. Any reason why?

```java
    static Map<String, ConnectorConfig> extractConfigurationFor(String prefix, Config root) {
        Iterable<String> names = root.getPropertyNames();
        Map<String, ConnectorConfig> configs = new HashMap<>();
        names.forEach(key -> {
            // $prefix$name.key=value (the prefix ends with a .)
            if (key.startsWith(prefix)) {
                // Extract the name
                String name = key.substring(prefix.length());
                if (name.contains(".")) { // We must remove the part after the first dot
                    String tmp = name;
                    name = tmp.substring(0, tmp.indexOf('.'));
                    configs.put(name, new ConnectorConfig(prefix, root, name));
                } else {
                    configs.put(name, new ConnectorConfig(prefix, root, name));
                }
            }
        });
        return configs;
    }
```

I would gladly try to contribute with a PR to make this possible.

marcelorubim:
Closes #695

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConfiguredChannelFactory.java
@@ -98,7 +98,9 @@ public class ConfiguredChannelFactory implements ChannelRegistar {
             if (key.startsWith(prefix)) {
                 // Extract the name
                 String name = key.substring(prefix.length());
-                if (name.contains(".")) { // We must remove the part after the first dot
+                if (name.charAt(0) == '"') { // Check if the name is enclosed by double quotes
+                    name = name.substring(1, name.lastIndexOf('"'));
+                } else if (name.contains(".")) { // We must remove the part after the first dot
                     String tmp = name;
                     name = tmp.substring(0, tmp.indexOf('.'));
                 }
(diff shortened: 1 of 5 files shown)


marcelorubim:
@cescoffier 
To avoid duplicate many tests, I just changed the ``dummy-sink`` to be without quotes.

So we can test both in the same test.

cescoffier:
Thanks!

eriknyk:
@marcelorubim  @cescoffier I'm using v2.4 and it still this not works.

```
mp.messaging.incoming."tc.payments.domain_event.job_created".connector=smallrye-kafka
mp.messaging.incoming."tc.payments.domain_event.job_created".group.id=demo3
mp.messaging.incoming."tc.payments.domain_event.job_created".value.deserializer=com.tiercollect.services.tigomoney.support.MessageDeserializer
mp.messaging.incoming."tc.payments.domain_event.job_created".broadcast=true

```

There is a problem with `mp.messaging.incoming."tc.payments.domain_event.job_created".value.deserializer`

Error Output:

```
Caused by: javax.enterprise.inject.spi.DeploymentException: java.lang.IllegalArgumentException: The attribute `value.deserializer` on connector 'smallrye-kafka' (channel: tc.payments.domain_event.job_created) must be set
```

Can you help with that please?

Best Regards.

PR #703, opened 2020-08-19
Add AMQP nack strategies for the modified delivery state

cescoffier:
Add two new nack strategies to the AMQP connector:

* modified-failed - marking the message modified and incrementing the delivery count (the message may be redelivered to the application)
* modified-failed-undeliverable-here - marking the message modified, incrementing the delivery count, and indicating the application cannot process the message. The broker cannot redeliver the message to the application.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/fault/AmqpModifiedFailedAndUndeliverableHere.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/fault/AmqpModifiedFailedAndUndeliverableHere.java
@@ -1,0 +1,33 @@
+package io.smallrye.reactive.messaging.amqp.fault;
+
+import static io.smallrye.reactive.messaging.amqp.i18n.AMQPLogging.log;
+
+import java.util.concurrent.CompletionStage;
+
+import io.smallrye.reactive.messaging.amqp.AmqpMessage;
+import io.smallrye.reactive.messaging.amqp.ConnectionHolder;
+import io.vertx.mutiny.core.Context;
+
+/**
+ * This nack strategy marking the message as {@code modified} and set the {@code delivery-failed} attribute to {@code true},
+ * as well as the {@code undeliverable-here} flag to {@code true}.
+ * <p>
+ * The message will not be redelivered on the same node, but may be redelivered on another node.
+ * <p>
+ * See http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-modified.
+ */
+public class AmqpModifiedFailedAndUndeliverableHere implements AmqpFailureHandler {
+
+    private final String channel;
+
+    public AmqpModifiedFailedAndUndeliverableHere(String channel) {
+        this.channel = channel;
+    }
+
+    @Override
+    public <V> CompletionStage<Void> handle(AmqpMessage<V> msg, Context context, Throwable reason) {
+        log.nackedModifiedFailedMessageAndUndeliverableHere(channel);
+        log.fullIgnoredFailure(reason);
+        return ConnectionHolder.runOnContext(context, () -> msg.getAmqpMessage().modified(true, true));
+    }
+}
--- a/smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
+++ b/smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
@@ -45,8 +45,17 @@ public class AmqpFailureHandlerTest extends AmqpTestBase {
         return container.getBeanManager().createInstance().select(MyReceiverBean.class).get();
     }
 
+    private MyReceiverBeanRecovering deployRecovering() {
+        Weld weld = new Weld();
+        weld.addBeanClass(MyReceiverBeanRecovering.class);
+
+        container = weld.initialize();
+        await().until(() -> container.select(MediatorManager.class).get().isInitialized());
+        return container.getBeanManager().createInstance().select(MyReceiverBeanRecovering.class).get();
+    }
+
     @Test
-    public void testFailStrategy() throws InterruptedException {
+    public void testFailStrategy() {
         String address = UUID.randomUUID().toString();
         getFailConfig(address);
         MyReceiverBean bean = deploy();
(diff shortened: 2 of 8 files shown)


cescoffier:
CC @gemmellr

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/fault/AmqpModifiedFailedAndUndeliverableHere.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private final String channel;
+
+    public AmqpModifiedFailedAndUndeliverableHere(String channel) {
+        this.channel = channel;
+    }
+
+    @Override
+    public <V> CompletionStage<Void> handle(AmqpMessage<V> msg, Context context, Throwable reason) {
+        log.nackedModifiedFailedMessage(channel);
```
gemmellr, line 29: This is using the same log message as the other failure handler. Seems like it should have its own variant, or the message should always indicate both the deliveryFailed and undeliverableHere value, with the latter passed in, otherwise it would be difficult to tell which handler variant actually occurred.

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
```diff
@@ -131,6 +140,46 @@ public void testRejectStrategy() {
...
+        MyReceiverBeanRecovering bean = deployRecovering();
+        AtomicInteger counter = new AtomicInteger();
+        AmqpConnector connector = container.getBeanManager().createInstance().select(AmqpConnector.class,
+                ConnectorLiteral.of(AmqpConnector.CONNECTOR_NAME)).get();
+        await().until(() -> isAmqpConnectorReady(connector));
+        await().until(() -> isAmqpConnectorAlive(connector));
+
+        usage.produceTenIntegers("modified-failed", counter::getAndIncrement);
+
+        await().atMost(2, TimeUnit.MINUTES).until(() -> bean.list().size() >= 10);
```
gemmellr, line 155: Seems odd to wait for >=10 but then assert exactly 13 values, this will surely be racey? Particularly as the 10th message will satistfy the await, but is also one of the ones to be redelivered.

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
```diff
@@ -131,6 +140,46 @@ public void testRejectStrategy() {
...
+        AtomicInteger counter = new AtomicInteger();
+        AmqpConnector connector = container.getBeanManager().createInstance().select(AmqpConnector.class,
+                ConnectorLiteral.of(AmqpConnector.CONNECTOR_NAME)).get();
+        await().until(() -> isAmqpConnectorReady(connector));
+        await().until(() -> isAmqpConnectorAlive(connector));
+
+        usage.produceTenIntegers("modified-failed-undeliverable-here", counter::getAndIncrement);
+
+        await().atMost(2, TimeUnit.MINUTES).until(() -> bean.list().size() >= 10);
+        // All messages should not have been received, 3 6 and 9 are NOT redelivered
```
gemmellr, line 176: Should have been received?

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
```diff
@@ -131,6 +140,46 @@ public void testRejectStrategy() {
...
+        AtomicInteger counter = new AtomicInteger();
+        AmqpConnector connector = container.getBeanManager().createInstance().select(AmqpConnector.class,
+                ConnectorLiteral.of(AmqpConnector.CONNECTOR_NAME)).get();
+        await().until(() -> isAmqpConnectorReady(connector));
+        await().until(() -> isAmqpConnectorAlive(connector));
+
+        usage.produceTenIntegers("modified-failed", counter::getAndIncrement);
+
+        await().atMost(2, TimeUnit.MINUTES).until(() -> bean.list().size() >= 10);
+        // All messages should not have been received + 3, 6 and 9 are redelivered
```
gemmellr, line 156: Should have been received?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/fault/AmqpModifiedFailedAndUndeliverableHere.java
cescoffier: +1

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
cescoffier: Good point!

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
cescoffier: yes, we capture the list before the failure.

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
cescoffier: yes, we capture the list before the failure.

cescoffier:
PR updated.

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpFailureHandlerTest.java
gemmellr: I mean that the comment says "All messages should **not** have been received", while the test expects they were all recieved (plus a few twice), and wont get that far unless enough were. It then goes on to assert the exact paylods, so it seems like the comment should be  "All messages should have been received". Or maybe "now" instead of "not" might work too.

Same for other instance below

gemmellr:
The main code seems fine overall, though I notice the 'modified failed' test failed in CI. It looks like this was as the receipt order was unexpected. I'd guess this was because one of the messages was received and modified-failed and could then be redelivered, but this occurred before some of the later messages were ever sent to the broker and/or consumer begin with. Likely as the producing and consuming is happening together, making their timing non deterministic.

cescoffier:
PR updated again.

gemmellr:
Looks good. An additional test might have been sending the messages in a couple of batches where only the last in the batch is modified-failed, and is then redelivered immediately, reciept verified, before the next batch is sent and the same done, meaning you can deterministically check the whole lot are received in an exact expected order.

This is a good example of where non-brokered tests (as per https://github.com/vert-x3/vertx-amqp-client/pull/45) of client behaviour are often better, as it could tightly controll the messages given to the consumer and also verify the acknowledgements.

gemmellr (approved):

Issue #588, opened 2020-06-05
Make using Kafka keys simpler


zeljkot:
Most annotated methods can work with two kinds of types: payload and message. Messages provide more flexibility but are more complicated to use because they:
* cannot be automatically acknowledged POST_PROCESSING 
* cannot be marked as @Blocking

With Kafka, it is quite common to ask for message key, but then methods must work with messages rather than payloads, so all those ease of use features cannot be utilized.

Please provide a way to get Kafka keys and still be able to use simpler reactive API.

In the meantime, here is a workaround:

```java
@Incoming("some-topic")
@Outgoing("some-topic-internal")
public Message<Tuple2<String, byte[]>> onMessage(KafkaRecord<String, byte[]> message) {
  return message.withPayload(Tuple2.of(message.getKey(), message.getPayload()));
}

@Incoming("some-topic-internal")
@Blocking
public void onMessage(Tuple2<String, byte[]> tuple) {
  doSomething(tuple.getItem1(), tuple.getItem2());
}
```

cescoffier:
Forgot to close this issue. It's been addressed in 2.5 with the support of `Record`.

PR #726, opened 2020-09-08
WIP - Metadata injection support

cescoffier:
This PR implement metadata injection, so methods can receive metadata object attached to incoming messages directly as a method parameter. 
It supports mandatory metadata and optional metadata:

```java
public String process(String payload, MyMetadata mandatory, Optional<MyOtherMetadata> optional) {...}
```

It only supports methods consuming a single payload and message.

TODO:

* [ ] - Documentation
* [ ] - Example with Kafka to extract the key

Related to #588

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -428,7 +428,6 @@ public interface Message<T> {
             throw new IllegalArgumentException("`className` must not be `null`");
         }
         for (Object meta : getMetadata()) {
-            System.out.println("Comparing " + meta.getClass().getName() + " and " + className);
             if (meta.getClass().getName().equals(className)) {
                 return Optional.of(meta);
             }
--- a/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
+++ b/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
@@ -418,6 +418,25 @@ public interface Message<T> {
     }
 
     /**
+     * Retrieves the metadata associated with the given classname.
+     *
+     * @param className the classname of the metadata to retrieve, must not be {@code null}
+     * @return an {@link Optional} containing the associated metadata, empty if none.
+     */
+    default Optional<?> getMetadata(String className) {
+        if (className == null) {
+            throw new IllegalArgumentException("`className` must not be `null`");
+        }
+        for (Object meta : getMetadata()) {
+            System.out.println("Comparing " + meta.getClass().getName() + " and " + className);
+            if (meta.getClass().getName().equals(className)) {
+                return Optional.of(meta);
+            }
+        }
+        return Optional.empty();
+    }
+
+    /**
      * @return the supplier used to retrieve the acknowledgement {@link CompletionStage}.
      */
     default Supplier<CompletionStage<Void>> getAck() {
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
@@ -1,25 +1,23 @@
 package io.smallrye.reactive.messaging;
 
-import static io.smallrye.reactive.messaging.i18n.ProviderExceptions.ex;
-import static io.smallrye.reactive.messaging.i18n.ProviderLogging.log;
-import static io.smallrye.reactive.messaging.i18n.ProviderMessages.msg;
-
-import java.util.Objects;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.function.Function;
-
-import javax.enterprise.inject.Instance;
-
+import io.smallrye.mutiny.Uni;
+import io.smallrye.reactive.messaging.connectors.WorkerPoolRegistry;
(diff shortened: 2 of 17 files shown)


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -417,6 +417,25 @@ default Metadata getMetadata() {
...
+     *
+     * @param className the classname of the metadata to retrieve, must not be {@code null}
+     * @return an {@link Optional} containing the associated metadata, empty if none.
+     */
+    default Optional<?> getMetadata(String className) {
+        if (className == null) {
+            throw new IllegalArgumentException("`className` must not be `null`");
+        }
+        for (Object meta : getMetadata()) {
+            System.out.println("Comparing " + meta.getClass().getName() + " and " + className);
```
kenfinnigan, line 431: Should that be a proper log message?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
```diff
@@ -65,13 +117,48 @@ public void initialize(Object bean) {
                     try {
                         return this.configuration.getMethod().invoke(bean, args);
                     } catch (Exception e) {
+                        e.printStackTrace();
```
kenfinnigan, line 120: Remove?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
```diff
@@ -65,13 +117,48 @@ public void initialize(Object bean) {
...
     }
 
+    @SuppressWarnings("unchecked")
+    protected <T> T invoke(Message<?> message) {
+        try {
+            Objects.requireNonNull(this.invoker, msg.invokerNotInitialized());
+            Object[] objects = invocationParameters.apply(message);
+            return (T) this.invoker.invoke(objects);
+        } catch (RuntimeException e) { // NOSONAR
+            log.methodException(configuration().methodAsString(), e);
```
kenfinnigan, line 135: Should this change to the exception type that logs a message and rethrows?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
```diff
@@ -65,13 +117,48 @@ public void initialize(Object bean) {
...
             }
         }
     }
 
+    @SuppressWarnings("unchecked")
+    protected <T> T invoke(Message<?> message) {
+        try {
+            Objects.requireNonNull(this.invoker, msg.invokerNotInitialized());
+            Object[] objects = invocationParameters.apply(message);
+            return (T) this.invoker.invoke(objects);
```
kenfinnigan, line 133: Do we need to store `objects` or can result be passed directly into call below?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
```diff
@@ -65,13 +117,48 @@ public void initialize(Object bean) {
...
+        } catch (RuntimeException e) { // NOSONAR
+            log.methodException(configuration().methodAsString(), e);
+            throw e;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    protected <T> Uni<T> invokeBlocking(Message<?> message) {
+        try {
+            Objects.requireNonNull(this.invoker, msg.invokerNotInitialized());
```
kenfinnigan, line 143: It would be nice if there was a way to not have to duplicate `invokeBlocking`, but I see you're wanting to do the parameter work in the blocking thread too.

kenfinnigan (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
cescoffier: Do you mean having a specific exception class that logs on creation? 

Here we log and rethrow to have a trace of the exception if the system recovers from the failure.

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
cescoffier: Fixed.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
cescoffier: Done.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
cescoffier: Done.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/AbstractMediator.java
cescoffier: A bit tricky, but yes the duplication is not great.

cescoffier:
Status: This PR is staled for now. Message Converter handles the main use case.

PR #734, opened 2020-09-11
Implement MutinyEmitter for Mutiny types

### LINKED ISSUE #710: Reactive messaging Emitter<T> interface is missing Mutiny integration
`Quarkus version: 1.7.0`

`package org.eclipse.microprofile.reactive.messaging;`

Why does the Emitter interface has only the `CompletionStage<Void> send(T msg)` method and is missing `Uni<Void>` send overload?

As shown here I can consume the messages using mutiny types https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/emitter/emitter.html#streams

`@Inject @Channel("my-channel") Multi<String> streamOfPayloads;`

But I can only send them using the Emitter that doesn't understand mutiny

`CompletionStage<Void> send(T msg)`

Does it make sense to have something like `Uni<Void> send(T msg);` and be consistent with the consuming interface?

kenfinnigan:
Fixes #710

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
@@ -17,9 +17,13 @@ public interface ChannelRegistry {
 
     void register(String name, Emitter<?> emitter);
 
+    void register(String name, MutinyEmitter<?> emitter);
+
     List<PublisherBuilder<? extends Message<?>>> getPublishers(String name);
 
     Emitter<?> getEmitter(String name);
+
+    MutinyEmitter<?> getMutinyEmitter(String name);
 
     List<SubscriberBuilder<? extends Message<?>, Void>> getSubscribers(String name);
 
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
@@ -1,0 +1,169 @@
+package io.smallrye.reactive.messaging.extension;
+
+import static io.smallrye.reactive.messaging.i18n.ProviderExceptions.ex;
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.messaging.OnOverflow;
+import org.reactivestreams.Publisher;
+
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.subscription.BackPressureStrategy;
+import io.smallrye.mutiny.subscription.MultiEmitter;
+import io.smallrye.reactive.messaging.EmitterBehavior;
+import io.smallrye.reactive.messaging.helpers.BroadcastHelper;
+
+public abstract class AbstractEmitter<T> implements EmitterBehavior {
+    protected final AtomicReference<MultiEmitter<? super Message<? extends T>>> internal = new AtomicReference<>();
+    protected final Multi<Message<? extends T>> publisher;
+
+    protected final String name;
+
+    protected final AtomicReference<Throwable> synchronousFailure = new AtomicReference<>();
+
+    @SuppressWarnings("unchecked")
+    public AbstractEmitter(EmitterConfiguration config, long defaultBufferSize) {
+        this.name = config.name;
+        if (defaultBufferSize <= 0) {
+            throw ex.illegalArgumentForDefaultBuffer();
+        }
+
+        Consumer<MultiEmitter<? super Message<? extends T>>> deferred = fe -> {
+            MultiEmitter<? super Message<? extends T>> previous = internal.getAndSet(fe);
+            if (previous != null) {
+                previous.complete();
+            }
+        };
+
+        Multi<Message<? extends T>> tempPublisher;
+        if (config.overflowBufferStrategy == null) {
+            Multi<Message<? extends T>> multi = Multi.createFrom().emitter(deferred, BackPressureStrategy.BUFFER);
+            tempPublisher = getPublisherUsingBufferStrategy(defaultBufferSize, multi);
+        } else {
+            tempPublisher = getPublisherForStrategy(config.overflowBufferStrategy, config.overflowBufferSize,
+                    defaultBufferSize, deferred);
+        }
+
+        if (config.broadcast) {
+            publisher = (Multi<Message<? extends T>>) BroadcastHelper
+                    .broadcastPublisher(tempPublisher, config.numberOfSubscriberBeforeConnecting).buildRs();
+        } else {
+            publisher = tempPublisher;
+        }
+    }
+
+    @Override
+    public synchronized void complete() {
+        verify(internal, name).complete();
+    }
+
+    @Override
+    public synchronized void error(Exception e) {
+        if (e == null) {
+            throw ex.illegalArgumentForException("null");
+        }
+        verify(internal, name).fail(e);
+    }
+
+    @Override
+    public synchronized boolean isCancelled() {
+        MultiEmitter<? super Message<? extends T>> emitter = internal.get();
+        return emitter == null || emitter.isCancelled();
+    }
+
+    @Override
+    public boolean hasRequests() {
+        MultiEmitter<? super Message<? extends T>> emitter = internal.get();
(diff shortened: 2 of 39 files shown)


kenfinnigan:
It's not complete but wanted to get some initial feedback. Still needs documentation.

Also, `sendAndAwait()` seems to hang indefinitely when running `MutinyEmitterAndAwaitInjectionTest`. Not sure if it's because I made a mistake in the implementation or in the test.

api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
```diff
@@ -17,10 +17,14 @@
 
     void register(String name, Emitter<?> emitter);
 
+    void register(String name, MutinyEmitter<?> emitter);
+
     List<PublisherBuilder<? extends Message<?>>> getPublishers(String name);
 
     Emitter<?> getEmitter(String name);
 
+    MutinyEmitter<?> getMutinyEmitter(String name);
```
cescoffier, line 26: Do we have to add this method or "translate" a _regular_ emitter to a MutinyEmitter in the injection point?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
```diff
@@ -0,0 +1,165 @@
...
+    public Publisher<Message<? extends T>> getPublisher() {
+        return publisher;
+    }
+
+    boolean isSubscribed() {
+        return internal.get() != null;
+    }
+
+    protected synchronized void emit(Message<? extends T> message) {
+        MultiEmitter<? super Message<? extends T>> emitter = verify(internal, name);
```
cescoffier, line 140: there is a missing check, probably missing before, `message` must not be `null`.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
```diff
@@ -20,8 +21,9 @@ public EmitterConfiguration() {
         // Used for proxies.
     }
 
-    public EmitterConfiguration(String name, OnOverflow onOverflow, Broadcast broadcast) {
+    public EmitterConfiguration(String name, Boolean isMutinyEmitter, OnOverflow onOverflow, Broadcast broadcast) {
```
cescoffier, line 24: Why `Boolean` as an object?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MutinyEmitterImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void sendAndAwait(T payload) {
+        send(payload).await().indefinitely();
+    }
+
+    @Override
+    public Cancellable sendAndForget(T payload) {
+        return send(payload).subscribe().with(x -> {});
```
cescoffier, line 42: Wondering if we need to log failure property?
In the next Mutiny version, it will be delegated to the droppedException handler.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MutinyEmitterImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public <M extends Message<? extends T>> void sendAndAwait(M msg) {
+        send(msg).await().indefinitely();
+    }
+
+    @Override
+    public <M extends Message<? extends T>> Cancellable sendAndForget(M msg) {
+        return send(msg).subscribe().with(x -> {});
```
cescoffier, line 62: Should we log the potential failure properly?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterAndAwaitInjectionTest.java
```diff
@@ -0,0 +1,675 @@
...
+
+        public MutinyEmitter<String> emitter() {
+            return emitter;
+        }
+
+        public List<String> list() {
+            return list;
+        }
+
+        public void run() {
```
cescoffier, line 460: Try running on another thread, as await is blocking the caller thread until acknowledgment.

cescoffier: Also, having a thread dump to verify where the blocking happens would be great.

cescoffier (changes requested):

api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
kenfinnigan: In theory we could do it all through `getEmitter`, but if I recall it would require hanging onto injection point type in the map of extension to know which is which

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
kenfinnigan: Previously it was done in the `send()` method, which it still is, but can add it here too as extra precaution

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
kenfinnigan: Don't know, can make it primitive

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MutinyEmitterImpl.java
kenfinnigan: I can add a handler to log the failure.

I was wondering if we should do anything like that

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MutinyEmitterImpl.java
kenfinnigan: Probably, same as above

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterAndAwaitInjectionTest.java
kenfinnigan: So `run()` on a separate thread? Can give that a go

api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
cescoffier: yeah, so forget about it. Your way is simpler.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
cescoffier: oh ok, missed it. So it's good.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/MutinyEmitterImpl.java
cescoffier: I believe we should add a handler and properly log.
The dropped exception handler from mutiny is too generic, and here we can make the log message useful for the user.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterAndAwaitInjectionTest.java
cescoffier: yes

kenfinnigan:
I'm going to close and reopen, something funky is happening

PR #736, opened 2020-09-12
Add support for Kafka Record

cescoffier:
Add support for Kafka record so the user can configure the key/value easily.
It also allows setting a `null` value.

- Add support for Kafka Records
- Disable Java 15 builds as the certificates look outdated.

### DIFF
--- a/documentation/src/main/doc/modules/kafka/pages/outbound.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/outbound.adoc
@@ -52,6 +52,23 @@ By default, the written record contains:
 * no key, or the key configured using the `key` attribute or the key passed in the metadata attached to the `Message`
 * the timestamp computed for the system clock (`now`) or the timestamp passed in the metadata attached to the `Message`
 
+=== Sending key/value pairs
+
+In the Kafka world, it's often necessary to send _records_, i.e. a key/value pair.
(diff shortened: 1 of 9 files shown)


documentation/src/main/doc/modules/kafka/pages/outbound.adoc
```diff
@@ -52,6 +52,22 @@ By default, the written record contains:
...
+The connector provides the {javadoc-base}/apidocs/io/smallrye/reactive/messaging/kafka/Record.html[`io.smallrye.reactive.messaging.kafka.Record`] class that you can use to send a pair:
+
+[source, java]
+----
+include::example$outbound/KafkaRecordExample.java[tag=code]
+----
+
+When the connector receives a message with a `Record` payload, it extracts the key and value from it.
+The configured serializers for both the key and the value must be compatible with the record's key and value.
+In addition, both the `key` and the `value` can be `null`.
```
kenfinnigan, line 67: Do we mean they both can be null, or that either of them can be null but not both at the same time?

kenfinnigan (approved):
Just the one comment on the docs as to whether it needs some clarification

documentation/src/main/doc/modules/kafka/pages/outbound.adoc
cescoffier: Fixed.

PR #738, opened 2020-09-14
Introduce Message Converters

cescoffier:
This PR is a replacement for the metadata injection support (#726).
The main use case for metadata injection was to access technical metadata from method accepting payload only. 
While functional, the API was not great, and it removes a few "type" check making error detection most complicated.

MessageConverter takes the problem from the opposite direction.
It allows converting a message into another message, generally to change the type of payload. 
For example, a converter can transform a message coming from Kafka to a KafkaRecord, which means that the user can access the key directly. 

A converter is a simple bean implementing an interface.
Then, reactive messaging looks for the right converter if needed and invoke it for each transiting message. 
For the user, it's transparent, the converter being called to transform the incoming message into the expected type.

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
@@ -1,5 +1,6 @@
 package io.smallrye.reactive.messaging;
 
+import io.smallrye.common.annotation.Experimental;
 import org.eclipse.microprofile.reactive.messaging.Message;
 
 import javax.enterprise.inject.spi.Prioritized;
@@ -12,6 +13,7 @@ import java.lang.reflect.Type;
  * When multiple converters are available, implementation should override the {@link #getPriority()} method.
  * The default priority is {@link #CONVERTER_DEFAULT_PRIORITY}. Converters with higher priority are executed first.
  */
+@Experimental("SmallRye only feature")
 public interface MessageConverter extends Prioritized {
 
     /**
--- a/api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
@@ -1,0 +1,33 @@
+package io.smallrye.reactive.messaging;
+
+import java.lang.reflect.Type;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+
+/**
+ * Converter transforming {@code Message<A>} into {@code Message<B>}.
+ * To register a converter, expose a, generally {@code ApplicationScoped} bean, implementing this interface.
+ */
+public interface MessageConverter {
+
+    /**
+     * Checks whether this instance of converter can convert the given message {@code in} into a {@code Message<T>} with
+     * {@code T} being the type represented by {@code target}.
+     *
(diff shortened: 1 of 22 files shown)


cescoffier:
A follow task would be to extend connectors to expose converters simplifying the development model.

api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Checks whether this instance of converter can convert the given message {@code in} into a {@code Message<T>} with
+     * {@code T} being the type represented by {@code target}.
+     *
+     * @param in the input message, not {@code null}
+     * @param target the target type, generally the type ingested by a method
+     * @return {@code true} if the conversion is possible, {@code false} otherwise.
+     */
+    boolean accept(Message<?> in, Type target);
```
kenfinnigan, line 21: `accept` is quite overloaded.

Maybe `canConvert`? or `isConvertible`? or `handles`?

kenfinnigan (changes requested):
Do we need to define a priority for the ordering of multiple converters handling the same types?

Is it an option to have `MessageConverter` actually define the types on it, and then we "know" which converter is applicable based on the types? Or does that not make sense? I might be missing something in understanding this

cescoffier:
On Mon 14 Sep 2020 at 21:15, Ken Finnigan <notifications@github.com> wrote:

>
>
> *@kenfinnigan* requested changes on this pull request.
>
>
>
>
> Do we need to define a priority for the ordering of multiple converters
> handling the same types?
>
>
Doesn't CDI has a priority annotation? I need to check we I believe we do
some kind of sorting somewhere else.




> Is it an option to have MessageConverter actually define the types on it,
> and then we "know" which converter is applicable based on the types? Or
> does that not make sense? I might be missing something in understanding this
>
You mean having a getSupportedType method? It won't work for Kafka, as for
example, it bases the acceptance on the metadata, not on the payload type.



------------------------------

kenfinnigan:
CDI does support `@Priority`, but I'm not sure how that behaves when we're doing `Instance<MessageConverter>`.

Not sure there's any priority ordering in that case, and then we just for loop through them to find one

cescoffier:
Just found where I did it: https://github.com/quarkusio/quarkus/blob/0dfe85b0bdbc8acd7646b5aa35ef1ac35c28e618/extensions/grpc/runtime/src/main/java/io/quarkus/grpc/runtime/GrpcContainer.java#L27-L49

So you are right, I need the same type of logic. How did I forget this?

api/src/main/java/io/smallrye/reactive/messaging/MessageConverter.java
cescoffier: Switched to `canConvert`.

cescoffier:
@kenfinnigan I've introduced priority for converters.

kenfinnigan (dismissed):

zeljkot:
Are there any examples for this?

cescoffier:
@zeljkot https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2.4/model/model.html#converters

PR #745, opened 2020-09-16
Implement MutinyEmitter for Mutiny types

### LINKED ISSUE #710: Reactive messaging Emitter<T> interface is missing Mutiny integration

kenfinnigan:
Fixes #710

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
@@ -11,6 +11,7 @@ import io.smallrye.reactive.messaging.annotations.Broadcast;
  */
 public class EmitterConfiguration {
     public String name;
+    public boolean isMutinyEmitter;
     public OnOverflow.Strategy overflowBufferStrategy;
     public long overflowBufferSize;
     public boolean broadcast;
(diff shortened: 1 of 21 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
```diff
@@ -11,6 +11,7 @@
  */
 public class EmitterConfiguration {
     public String name;
+    public Boolean isMutinyEmitter;
```
cescoffier, line 14: why Object? Isn't `boolean` enough?

cescoffier (changes requested):
LGTM, Can you update the documentation?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/EmitterConfiguration.java
kenfinnigan: Nuts, missed that one

kenfinnigan:
Sorry blanked on the docs!

kenfinnigan:
Should be sorted now, sorry about that!

cescoffier (approved):
Thanks!

PR #753, opened 2020-09-17
[752] Adding beans.xml so that jms jar is found if implicit scan is t…

malair:
Issue 752: Adding a beans.xml file so that the smallrye reactive message jms jar is picked up if implicit scanning is turned off.


kenfinnigan:
I think it needs to have content similar to https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/resources/META-INF/beans.xml

PR #760, opened 2020-09-19
Kafka test reorg and disable auto commit

cescoffier:
This PR disables the Kafka auto commit (if not explicitely set by the user) and refactor the tests.

This PR:

- Removes a dependency not used anymore since AMQP tests uses junit 5
- Fixes typos in logged messages
- Removes non-kafka properties from configurations
- Fixes typo in exception class name
- Apply a major refactoring of the Kafka tests
- Uses test container instead of debezium
- Disables auto commit by default


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -86,9 +86,14 @@ public KafkaSource(Vertx vertx,
...
 
+        if (!kafkaConfiguration.containsKey("enable.auto.commit")) {
+            log.disableAutoCommit(config.getChannel());
+            kafkaConfiguration.put("enable.auto.commit", "false");
+        }
+
         String commitStrategy = config
                 .getCommitStrategy()
-                .orElse(Boolean.parseBoolean(kafkaConfiguration.getOrDefault(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "true"))
+                .orElse(Boolean.parseBoolean(kafkaConfiguration.getOrDefault(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false"))
```
pcasaes, line 96: It looks like **enable.auto.commit** will always have a value now, no need for get**OrDefault** on line 96.

Also , why not use the defined constant ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG instead of the literal **enable.auto.commit** on line 91?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -57,6 +57,7 @@
 @ConnectorAttribute(name = "value.deserializer", type = "string", direction = Direction.INCOMING, description = "The deserializer classname used to deserialize the record's value", mandatory = true)
 @ConnectorAttribute(name = "fetch.min.bytes", type = "int", direction = Direction.INCOMING, description = "The minimum amount of data the server should return for a fetch request. The default setting of 1 byte means that fetch requests are answered as soon as a single byte of data is available or the fetch request times out waiting for data to arrive.", defaultValue = "1")
 @ConnectorAttribute(name = "group.id", type = "string", direction = Direction.INCOMING, description = "A unique string that identifies the consumer group the application belongs to. If not set, a unique, generated id is used")
+@ConnectorAttribute(name = "enable.auto.commit", type = "boolean", direction = Direction.INCOMING, description = "Whether or not auto-commit is enabled. It's recommended to NOT use auto-commit and let's the connector handle commit", defaultValue = "false")
```
pcasaes, line 60: Might be worded a bit better:

> It's recommended to NOT use auto-commit and let the connector handle commits

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
cescoffier: Done!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
cescoffier: Fixed !

PR #762, opened 2020-09-21
Use timer on throttled strategy and improve throttled health check

pcasaes:
@cescoffier 

As we discussed on #672 here I added the use of a vert.x periodic timer to perform the commit and run the health check by testing how long the oldest received message has gone unprocessed. In the case of replaying a topic from the beginning it can still trigger so it can be disabled.


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -41,37 +42,27 @@
...
     private final KafkaConsumer<?, ?> consumer;
     private final KafkaSource<?, ?> source;
-    private final int maxReceivedWithoutAckAllowed;
+    private final int unprocessedRecordMaxAge;
+    private final int autoCommitInterval;
 
     private volatile Context context;
-    private long nextCommitTime;
+
+    private Long timerId = null;
```
cescoffier, line 56: while an object, a long with an initial value of -1 works well.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -82,10 +73,16 @@ public static KafkaThrottledLatestProcessedCommit create(KafkaConsumer<?, ?> con
             Map<String, String> config,
             KafkaSource<?, ?> source) {
 
-        int maxPollRecords = Integer.parseInt(config.getOrDefault(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, "500"));
-        int maxReceivedWithoutAckAllowed = getNextPowerOfTwoEqualOrGreater(maxPollRecords * 2);
-        log.settingMaxReceivedWithoutAckAllowed(config.get(ConsumerConfig.GROUP_ID_CONFIG), maxReceivedWithoutAckAllowed);
-        return new KafkaThrottledLatestProcessedCommit(consumer, source, maxReceivedWithoutAckAllowed);
+        int unprocessedRecordMaxAge = Integer.parseInt(config.getOrDefault("throttled.unprocessed-record-max-age", "60000"));
```
cescoffier, line 76: you should store the IncomingKAfkaConfiguration and use this object instead of accessing the config. Typically, alias and default values are managed for you.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -106,25 +103,26 @@ public void partitionsAssigned(Context context, Set<TopicPartition> partitions)
...
-        resetNextCommitTime();
+        if (timerId != null) {
+            context.owner().cancelTimer(timerId);
+            timerId = null;
+        }
+
+        if (!partitions.isEmpty()) {
+            timerId = context
+                    .owner()
+                    .setPeriodic(autoCommitInterval, this::flushAndCheckHealth);
```
cescoffier, line 114: instead of using a periodic task, we may want to do a timed task starting a new timed task. So the gap between the end of the task and the beginning of the next one will be the configured time. Without it's the time between 2 beginning which might be too close.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -150,6 +150,7 @@ public KafkaSource(Vertx vertx,
                     + 11_000L; // it's possible that it might expire 10 seconds before when we need it to
 
             kafkaConsumer.partitionsAssignedHandler(set -> {
+                commitHandler.partitionsAssigned(vertx.getOrCreateContext(), set);
```
cescoffier, line 153: shouldn't we pause first?

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: Sure thing.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: I wavered between the two options. I'll go with your suggestions.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
pcasaes: Works either way since we are in the context's event thread. I'll move it if it makes it clearer.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: I see it. thanks.

pcasaes:
@cescoffier review requests done.

cescoffier (approved):
LGTM!

PR #764, opened 2020-09-21
Implement Record and ConsumerRecord converter in the Kafka connector

cescoffier:
This PR:

- Implements Record and ConsumerRecord converters
- Allows converters to be called on null payloads

The message converters allow the user's method to ingest Record and ConsumerRecord directly, without having to implement a custom conversion. 
Note that the user can also provide a specific converter is needed.


cescoffier:
CC @pcasaes - that's the PR I was referring too on the mailing list.

kenfinnigan (approved):

pcasaes:
Very nice! Looks good.

PR #771, opened 2020-09-22
Provide a flag to disable the readiness health check  in the kafka connector

cescoffier:


kenfinnigan (dismissed):

andreas-eberle:
@cescoffier: I tried setting `health-readiness-enabled` and `health-enabled` to `false` for a topic like this

```mp.messaging.incoming.my-events.health-enabled=false``` 

Unfortunately, the health check still fails with a `TimeoutException`:

```
Caused by: org.apache.kafka.common.errors.TimeoutException: Timed out waiting for a node assignment.
```

Did I configure the property wrong?

cescoffier:
Readiness or liveness? Did you try to disable all health checks?

andreas-eberle:
Ah, sorry, I just found an old `quarkus.kafka.health.enabled=true`  flag in my application.properties. After removing that, it works.

PR #773, opened 2020-09-23
Explain how applications without connectors work

cescoffier:
Just a small paragraph explaining a bit more how applications without connectors work﻿.


xstefank (approved):
LGTM

PR #774, opened 2020-09-23
Improve the latest commit strategy

cescoffier:
Instead of committing the last polled offset, commit for each partition the higher offset. 

Also, improve the tests and fix unknown properties in the admin client.

@pcasaes can you have a look?

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
@@ -1,0 +1,39 @@
+package io.smallrye.reactive.messaging.kafka.commit;
+
+import io.vertx.mutiny.core.Context;
+import io.vertx.mutiny.core.Vertx;
+
+/**
+ * A class holding a vert.x context to make sure methods are always run from the same one.
+ */
+public class ContextHolder {
+
+    protected final Vertx vertx;
+    protected volatile Context context;
+
+    public ContextHolder(Vertx vertx) {
+        this.vertx = vertx;
+    }
+
+    public Context getContext() {
+        Context ctx = this.context;
+        if (ctx == null) {
+            synchronized (this) {
+                ctx = this.context;
+                if (ctx == null) {
+                    this.context = ctx = vertx.getOrCreateContext();
+                }
+            }
+        }
+        return ctx;
+    }
+
+    public void runOnContext(Runnable runnable) {
+        if (Vertx.currentContext() == getContext()) {
+            runnable.run();
+        } else {
+            getContext().runOnContext(x -> runnable.run());
+        }
+    }
+
+}
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
@@ -1,30 +1,74 @@
 package io.smallrye.reactive.messaging.kafka.commit;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
 import io.smallrye.reactive.messaging.kafka.IncomingKafkaRecord;
-import io.vertx.mutiny.kafka.client.consumer.KafkaConsumer;
+import io.vertx.kafka.client.common.TopicPartition;
+import io.vertx.kafka.client.consumer.KafkaConsumer;
+import io.vertx.kafka.client.consumer.OffsetAndMetadata;
+import io.vertx.mutiny.core.Context;
+import io.vertx.mutiny.core.Vertx;
 
 /**
- * Will commit to the latest offset received by the Kafka consumer.
+ * Will commit the record offset received by the Kafka consumer (if higher than the previously committed offset).
  * This offset may be greater than the currently ACKed message.
  *
  * This handler is the default when `enable.auto.commit` is `false`.
+ * This strategy provides at-least-once delivery if the channel processes the message without performing
+ * any asynchronous processing.
  *
- * This strategy does not guarantee at-least-once delivery.
+ * This strategy should not be used on high-load as offset commit is expensive.
  *
  * To use set `commit-strategy` to `latest`.
  */
 public class KafkaLatestCommit implements KafkaCommitHandler {
 
     private final KafkaConsumer<?, ?> consumer;
+    private final Map<TopicPartition, Long> offsets = new HashMap<>();
+    private final Vertx vertx;
+    private Context context;
 
-    public KafkaLatestCommit(KafkaConsumer<?, ?> consumer) {
-        this.consumer = consumer;
+    public KafkaLatestCommit(Vertx vertx, io.vertx.mutiny.kafka.client.consumer.KafkaConsumer<?, ?> consumer) {
+        this.consumer = (KafkaConsumer<?, ?>) consumer.getDelegate();
+        this.vertx = vertx;
+    }
+
+    private synchronized Context getContext() {
+        if (context == null) {
+            context = vertx.getOrCreateContext();
+        }
+        return context;
     }
 
     @Override
     public <K, V> CompletionStage<Void> handle(IncomingKafkaRecord<K, V> record) {
-        return consumer.commit().subscribeAsCompletionStage();
+        Context ctxt = getContext();
+
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        ctxt.runOnContext((x) -> {
+            Map<TopicPartition, OffsetAndMetadata> map = new HashMap<>();
+            TopicPartition key = new TopicPartition(record.getTopic(), record.getPartition());
+            Long last = offsets.get(key);
+            // Verify that the latest committed offset before this one.
+            if (last == null || last < record.getOffset() + 1) {
+                offsets.put(key, record.getOffset() + 1);
+                map.put(key, new OffsetAndMetadata(record.getOffset() + 1, null));
+
+                consumer.commit(map, ar -> {
+                    if (ar.failed()) {
+                        future.completeExceptionally(ar.cause());
+                    } else {
+                        future.complete(null);
+                    }
+                });
+            } else {
+                future.complete(null);
+            }
+        });
+
+        return future;
     }
 }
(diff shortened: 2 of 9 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
```diff
@@ -18,13 +23,33 @@
...
     @Override
     public <K, V> CompletionStage<Void> handle(IncomingKafkaRecord<K, V> record) {
-        return consumer.commit().subscribeAsCompletionStage();
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        Map<TopicPartition, OffsetAndMetadata> map = new HashMap<>();
+        TopicPartition key = new TopicPartition(record.getTopic(), record.getPartition());
+        Long last = offsets.get(key);
+        // Verify that the latest committed offset before this one.
+        if (last == null || last < record.getOffset() + 1) {
+            offsets.put(key, record.getOffset() + 1);
```
pcasaes, line 40: **offsets** isn't threadsafe. The user might call the **Message#ack** method inside  a worker thread. It's the reason why on the Throttled strategy we run the handle code inside the context's event thread.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
```diff
@@ -18,13 +23,33 @@
...
+            map.put(key, new OffsetAndMetadata(record.getOffset() + 1, null));
+
+            consumer.commit(map, ar -> {
+                if (ar.failed()) {
+                    future.completeExceptionally(ar.cause());
+                } else {
+                    future.complete(null);
+                }
+            });
```
pcasaes, line 49: This has always bothered me about this strategy. Committing on every message causes High load on the cluster. Maybe we should add a warning to the documentation.

Also, what' the implication of completing exceptionally here?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/commit/CommitStrategiesTest.java
```diff
@@ -0,0 +1,310 @@
...
+import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
+import io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener;
+import io.smallrye.reactive.messaging.kafka.base.MapBasedConfig;
+import io.smallrye.reactive.messaging.kafka.base.WeldTestBase;
+import io.smallrye.reactive.messaging.kafka.impl.KafkaSource;
+import io.vertx.kafka.client.consumer.KafkaReadStream;
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.kafka.client.consumer.KafkaConsumer;
+
+public class CommitStrategiesTest extends WeldTestBase {
```
pcasaes, line 38: Very nice!

pcasaes (changes requested):

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/commit/CommitStrategiesTest.java
cescoffier: Very happy with this. Will allow much fine-grain testing, but Man... that's definitely undocumented.

Also, rebalance do not calls the callbacks... It's a TODO in their code :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
cescoffier: it's actually better than the previous one that was committing on every message, every subscribed topic/partition. At least this one tracks the message more carefully. I've updated the doc.

Before switching to the throttled by default (which is the objective), we must do a lot more tests. Typically, it does not commit on application shutdown. 

About the completeExceptionally, if the commit fails, the failure is propagated downstream and mark the application as failed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
cescoffier: Fixed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
cescoffier: Fixed.

pcasaes (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -158,27 +160,37 @@ private void startFlushAndCheckHealthTimer() {
...
-                            e -> new OffsetAndMetadata().setOffset(e.getValue() + 1L)));
-            consumer.getDelegate().commit(offsets, a -> this.startFlushAndCheckHealthTimer());
-        } else {
-            this.startFlushAndCheckHealthTimer();
+    private synchronized Context getContext() {
+        if (context == null) {
+            context = vertx.getOrCreateContext();
         }
+        return context;
+    }
```
pcasaes, line 168: I ran a test locally on a project and it works perfectly!

I do have one final suggestion. **context** here is loaded only once, so using a lock every time we acquire seems heavy handed. What do you think about using a double checked lock? We would have to change context to be volatile on line 51 

```
private volatile Context context;
```

And then rewrite this method

```
    private Context getContext() {
        Context ctx = this.context;
        if (ctx == null) {
            synchronized (this) {
                ctx = this.context;
                if (ctx == null) {
                    this.context = ctx = vertx.getOrCreateContext();
                }
            }
        }
        return ctx;
    }
```

We could do the same with the _latest_ strategy - maybe even abstract it to reduce duplicate code.

Theoretically we would see a 100x performance gain.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -158,27 +160,37 @@ private void startFlushAndCheckHealthTimer() {
...
+
+    private void flushAndCheckHealth(long timerId) {
+        // The timer may be on another context, so make sure we are on the right one to access the store.
```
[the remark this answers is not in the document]
cescoffier: It only ensures it's the same event loop if the timer is created from an event loop, which here is not necessarily the case (if the timer is created from another event loop or from a regular thread). So, it's safer to just store a context and enforce we always run on it.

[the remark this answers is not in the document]
cescoffier: A possible optimization would be a trampoline: check if we are on the context already and if so run immediately, if not, run on context. Let me add that.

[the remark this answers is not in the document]
cescoffier: Actually, you are right, got confused with my multiple branches. In this one, no need to check as the partitionAssigned method is still called from the event loop (which is something we MUST change - and I've another branch where I started this).

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: Fixed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
```diff
@@ -0,0 +1,39 @@
+package io.smallrye.reactive.messaging.kafka.commit;
+
+import io.vertx.mutiny.core.Context;
+import io.vertx.mutiny.core.Vertx;
+
+/**
+ * A class holding a vert.x context to make sure methods are always run from the same one.
+ */
+public class ContextHolder {
```
pcasaes, line 9: Very nice!!!!

pcasaes (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
```diff
@@ -0,0 +1,39 @@
...
+        if (ctx == null) {
+            synchronized (this) {
+                ctx = this.context;
+                if (ctx == null) {
+                    this.context = ctx = vertx.getOrCreateContext();
+                }
+            }
+        }
+        return ctx;
+    }
```
pcasaes, line 29: This pattern comes from
https://www.oracle.com/technical-resources/articles/javase/bloch-effective-08-qa.html

Under: Best Practices for Lazy Initialization

It's also referenced in Java Concurrency in Practice by Goetz et al in section 16.2.4

> Subsequent changes in the JMM... have enabled DLC to work if resource is made volatile.

cescoffier: Oh, I know the pattern, it was just late at night :-D 
I isolated it into its own class, so we can reuse the pattern.

Going to be fun with the rebalance listener not called on the event loop anymore...

pcasaes: I have no doubt you knew :D

But I assume others might read this.

pcasaes: > Going to be fun with the rebalance listener not called on the event loop anymore...

Yes! Haven't looked at your draft PR yet.

pcasaes: I also admit to having used the DLC before Java 5. I was younger and naive 🤣

cescoffier:
Rebased.

PR #775, opened 2020-09-24
Cloud Event fixes & Resource cleanup

cescoffier:
Fix typos in Cloud Events attributes and date-time parsing.

Also clean up some resources in the connector (a timer, DLQ producer) and make sure tests don't keep producers/consumers open after the test execution.


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaFailureHandler.java
```diff
@@ -29,4 +29,8 @@ public static Strategy from(String s) {
 
     <K, V> CompletionStage<Void> handle(IncomingKafkaRecord<K, V> record, Throwable reason);
 
+    default void terminate() {
```
kenfinnigan, line 32: As the default is do nothing is this needed?

Or was it mostly to make explicit that this case doesn't require anything?

kenfinnigan (approved):
Retriggered CI as it seemed stuck

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaFailureHandler.java
cescoffier: The DLQ implementation overrides it, and close the producer.
Other strategies don't need to do anything.

cescoffier:
@kenfinnigan Oupsy... forgot to update the documentation files. Should be better now.

kenfinnigan (approved):

PR #777, opened 2020-09-25
Add kafka source concurrency experiment

pcasaes:
@cescoffier This isn't meant to be merged. 

It's an experiment showing what we were discussing about concurrent record processing. Here there are two tests that fail because with the default Acknowledgment strategy records are only delivered by the channel in series irrespective of consumer/client/partition count.

One of the experiments is interesting because it shows we can process records from different partitions concurrently without having to have one consumer/client per partition.

The experiment parameters and result are commented in the code.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
+++ b/smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
@@ -1,0 +1,315 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.serialization.IntegerDeserializer;
+import org.eclipse.microprofile.reactive.messaging.Acknowledgment;
+import org.eclipse.microprofile.reactive.messaging.Incoming;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+import io.smallrye.reactive.messaging.kafka.base.KafkaTestBase;
+import io.smallrye.reactive.messaging.kafka.base.MapBasedConfig;
+import io.smallrye.reactive.messaging.kafka.impl.KafkaSource;
+
+/**
+ * Kafka Source Concurrency Experiment:
+ * <p>
+ * Hypothesis: If we match a topic's partition count with consumer instances
+ * then we should see concurrent record processing.
+ * <p>
+ * We will use topics with 2 partitions.
+ * We will send 2 records with keys "1" and "2".
+ * We will test that both records are processed and done so through separate partitions within 25 seconds.
+ * <p>
+ * For control we will provide a bean that will process both records as quickly as possible.
+ * For the experiment we will provide beans that purposefully sleep for 30 seconds after processing (prevent ack)
+ * <p>
+ * The idea here is that a force sleep should only delay the processing of records from 1 partition and not all.
+ */
+class KafkaSourceConcurrencyTest extends KafkaTestBase {
+
+    private static final int RECORD_COUNT = 2;
+
+    private interface ExperimentBean {
+        List<IncomingKafkaRecord<String, Integer>> getResults();
+    }
+
+    private static void waitFor30SecondsOrAllRecordsProcessed(ExperimentBean bean) {
+        await()
+                .atMost(30, TimeUnit.SECONDS)
+                .until(() -> bean.getResults().size() >= RECORD_COUNT);
+    }
+
+    /**
+     * Control bean
+     * <p>
+     * processes the record and ack as quickly as possible
+     */
+    @ApplicationScoped
+    public static class ControlBean implements ExperimentBean {
+
+        private final List<IncomingKafkaRecord<String, Integer>> list = new CopyOnWriteArrayList<>();
+
+        @Override
+        public List<IncomingKafkaRecord<String, Integer>> getResults() {
+            return list;
+        }
+
+        @Incoming("data")
+        public CompletionStage<Void> process(IncomingKafkaRecord<String, Integer> input) {
+            list.add(input);
+            return input.ack();
+        }
+    }
+
+    /**
+     * Experiment bean that uses {@link Acknowledgment.Strategy.POST_PROCESSING}
+     * <p>
+     * This tests an incoming channel with defaults
+     */
+    @ApplicationScoped
+    public static class PostAckWith30SecondSleepBean implements ExperimentBean {
+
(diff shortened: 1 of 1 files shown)


smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
```diff
@@ -0,0 +1,322 @@
...
+class KafkaSourceConcurrencyTest extends KafkaTestBase {
+
+    private static final int RECORD_COUNT = 2;
+
+    private interface ExperimentBean {
+        List<IncomingKafkaRecord<String, Integer>> getResults();
+    }
+
+    private static void waitFor30SecondsOrAllRecordsProcessed(ExperimentBean bean) {
+        for (int i = 0; i < 30; i++) {
```
cescoffier, line 59: Awaitility is more robust for this.

cescoffier (commented):
Interesting. 

The tests fail with a timeout. Is it what you expected?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+        @Override
+        public List<IncomingKafkaRecord<String, Integer>> getResults() {
+            return list;
+        }
+
+        @Incoming("data")
+        public CompletionStage<Void> process(IncomingKafkaRecord<String, Integer> input) {
+            list.add(input);
+            waitFor30SecondsOrAllRecordsProcessed(this);
```
cescoffier, line 111: you are blocking the event loop, it's not legal.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
pcasaes: done

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceConcurrencyTest.java
pcasaes: Right, but this us just an experiment. One of the attempts does not block the event loop (process in a worker thread) and the result is the same.

pcasaes:
I expect all the tests to pass. 

```
@ConnectorAttribute(name = "partitions", type = "int", direction = Direction.INCOMING, description = "The number of partitions to be consumed concurrently. The connector creates the specified amount of Kafka consumers. It should match the number of partition of the targeted topic", defaultValue = "1")
```

> The number of partitions to be consumed concurrently.

So even though we have two consumers set up for a topic for experiments `testPostAck30SecondSleep` and `testPostAckWorkerThread30SecondSleep` we are not seeing concurrent consumption of records. If they were being consumed concurrently the blocking (30 second sleep) of one record wouldn't prevent the second (on a different partition) from being processed.

I believe that in the case of **Acknowledgment.Strategy.POST_PROCESSING** even if we have two or more consumers the records are being joined and then delivered in a single event thread.

pcasaes:
> I believe that in the case of Acknowledgment.Strategy.POST_PROCESSING even if we have two or more consumers the records are being joined and then delivered in a single event thread.

I was wrong about this! So I added a consumer rebalance to check that both consumers are live before producing and I added a check to the test to see if the records were produced in the correct event thread. For those tests that configure the connector with **partitions=2** the records are consumed in 2 event threads as expected. But those two tests still fail which means that something is still preventing records from being processed concurrently.

cescoffier:
Make sense, you block the processing waiting for acknowledgments. This is what you do when you want to enforce the order, but in your case you want concurrency. 
Check this version: https://github.com/smallrye/smallrye-reactive-messaging/compare/pc/kafka-consurrency-experiment?expand=1.

pcasaes:
I see. So what you did is very similar to what was done with the **NONE** Ack strategy. It does work but with **POST_PROCESSING** we'll lose at-least once semantics because the **CompletableFuture.completedFuture(null)** will inform the framework to ack the message right?

This makes me question the validity of setting **partitions** to a value greater than 1. The framework already decouples processing from consumption (on the wire). Setting it to a value greater than 1 will cost us the resource of an extra connection and give us marginal network throughput. Kafka already does a good job of delivering records to the consumer, especially if you set [fetch.max.wait.ms](https://docs.confluent.io/current/installation/configuration/consumer-configs.html#fetch.max.wait.ms) to something reasonable (100 to 500ms).

I suppose if you care about latency then it would be valuable to raise **partitions**.

cescoffier:
> I see. So what you did is very similar to what was done with the **NONE** Ack strategy. It does work but with **POST_PROCESSING** we'll lose at-least once semantics because the **CompletableFuture.completedFuture(null)** will inform the framework to ack the message right?

With post-processing yes. In this case, you can't use this strategy and should use the manual strategy which lets you ack/nack when the real processing is done. 

> 
> This makes me question the validity of setting **partitions** to a value greater than 1. The framework already decouples processing from consumption (on the wire). Setting it to a value greater than 1 will cost us the resource of an extra connection and give us marginal network throughput. Kafka already does a good job of delivering records to the consumer, especially if you set [fetch.max.wait.ms](https://docs.confluent.io/current/installation/configuration/consumer-configs.html#fetch.max.wait.ms) to something reasonable (100 to 500ms).
> 
> I suppose if you care about latency then it would be valuable to raise **partitions**.

Yes. I would like to find a better way by actually having 1 stream par partition. However, it fairly complex and I'm not even sure it's possible when we have components having state as it may be called concurrently.

pcasaes:
> I would like to find a better way by actually having 1 stream par partition. However, it fairly complex and I'm not even sure it's possible when we have components having state as it may be called concurrently.

Yeah, I can see it would be pretty tough. For now I'm thinking of using an Interceptor that acts a lot like Blocking but lets me have 1 worker thread per partition like i've done on this experiment. Using on MANUAL/NONE I can get it to give me at-least once delivery, ordered processing per partition, and concurrency between partitions but not within a partition.

cescoffier:
@pcasaes How to you want to proceed with this PR? Should we merge it? Should it update the documentation?

pcasaes:
Unless you think these tests are interesting I would close it. It would definitely be worth adding a note to the documentation that with partitions being greater that 1 we still only have one stream.

Adding an example on how to achieve at-least once delivery, ordered processing per partition, and concurrency between partitions would be nice.

cescoffier:
Actually, while I was working on something else, I may have find a more elegant _way_. Something like:

```java
@Incoming("in")
@Acknowledgement(MANUAL)
public void partitions(Multi<Message<T>> multi) {  
    multi.groupItems().asMulti(item -> getPartition(item)) // Return a Multi<Multi> which each nested multi being associated with one partition
       .onItem().transformToMultiAndMerge(streamForPartition -> 
            streamForPartition
               .emitOn(executor) // thread pool, should match the number of partition
               .onItem().call(m -> process(m)) // Run with n concurrent thread (nb of partitions)
               .onItem().call(m -> ack(m))
               .onFailure().call(m -> nack(m))
      ).subscribe().with(() -> /* ignored */)
);
}
```

The idea is to generate a multi per partition (group) and for each stream emits the items on a dedicated thread. These items are processed (`process` and ack/nack accordingly. It means that the items are processed in order for each partition.

pcasaes:
I see one problems with it. 

_thread pool, should match the number of partition_: This requires us to know before hand the number of partitions which might not always be the case. Also, not sure how this would impact a possible need to increase the number of partitions. I'm assuming it would incur downtime or at least a rolling restart as we would have to reconfigure affected services.

cescoffier:
I meant the `partitions` attribute.

pcasaes:
What I did doesn't offer any back pressure control. Does using a multi merge do that?

cescoffier:
Yes, even if in this case, it will continue re-requesting. The `.onItem().call(m -> process(m))` is the stage that will handle the back pressure (but that's Mutiny business)

pcasaes:
So if it get's backed up the kafka stream will also stop polling?

cescoffier:
Yes, but only from the stream attached to that partition.

pcasaes:
Oh that is nice. I'll try your solution out.

pcasaes:
Not seeing a **asMulti** in **MultiGroup**

cescoffier:
Yes `throttled` is what you need. It's becoming the default in the next version.

PR #782, opened 2020-09-29
lookup for CamelReactiveStreamsService before creating an instance

### LINKED ISSUE #781: CamelConnector init should check for existing CamelReactiveStreamsService configuration
It's possible that the `CamelReactiveStreamsService` is already configured on the `CamelContext` before `CamelConnector` attempts its own configuration in the `@PostConstruct` annotation `init` method.

This is the case in Camel Quarkus where its reactive streams extension has already pre-configured `CamelReactiveStreamsService`.

It would be good if `CamelConnector.init()` had a check like:

```java
@PostConstruct
public void init() {
    reactive = CamelReactiveStreams.get(camel);
    if (reactive == null) {
        // Configuration logic here...
    }
}
```

cescoffier:
Also, register the created instance in the Camel service registry

Fix #781


cescoffier:
Code checked by reporter.

PR #783, opened 2020-09-30
CDI Events for Consumer/Producer

ebullient:
Skipping qualifier for create/delete for now.. so the event is only for create. 
Some other detection (or maybe a separate event entirely) would be needed for destroy.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIEvents.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIEvents.java
@@ -1,0 +1,28 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Event;
+import javax.inject.Inject;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.producer.Producer;
+
+@ApplicationScoped
+public class KafkaCDIEvents {
+    @Inject
+    Event<Consumer<?, ?>> consumerEvent;
+
+    @Inject
+    Event<Producer<?, ?>> producerEvent;
+
+    public Event<Consumer<?, ?>> consumer() {
+        return consumerEvent;
+    }
+
+    public Event<Producer<?, ?>> producer() {
+        return producerEvent;
+    }
+
+    public KafkaCDIEvents() {
+    }
+}
--- a/smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
+++ b/smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
@@ -1,0 +1,89 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import java.lang.annotation.Annotation;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.LongAdder;
+
+import javax.enterprise.event.Event;
+import javax.enterprise.event.NotificationOptions;
+import javax.enterprise.util.TypeLiteral;
+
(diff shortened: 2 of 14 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIEvents.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public Event<Producer<?, ?>> producer() {
+        return producerEvent;
+    }
+
+    public KafkaCDIEvents() {
+    }
+
+    private KafkaCDIEvents(boolean isNoop) {
```
cescoffier, line 34: Can you add a comment about this constructor that is not used during "normal" operations.

cescoffier: Is it even used?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.enterprise.event.Event;
+import javax.enterprise.event.NotificationOptions;
+import javax.enterprise.util.TypeLiteral;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.producer.Producer;
+
+public class CountKafkaCdiEvents extends KafkaCDIEvents {
+    public static final KafkaCDIEvents noCdiEvents = new CountKafkaCdiEvents();
```
cescoffier, line 15: Should we reset the counter at the end of each test?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public final LongAdder firedConsumerEvents = new LongAdder();
```
cescoffier, line 17: Nice, I always used AtomicLong! @jponge might be interesting for the spies.

cescoffier (changes requested):
That's looks great! 

Two comments:

* not sure the private constructor in the KafkaCDIEvents class is used.
* wondering if we need to add a line in the doc, or if we should keep this under the radar until the metrics side is implemented.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
jponge: Yep

ebullient:
Yea.. I thought I removed it? May not have pushed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIEvents.java
ebullient: Not anymore. I thought I removed it.. but I may not have pushed the change. Will double check.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/CountKafkaCdiEvents.java
ebullient: For this one.. it's saying "I don't care" .. so I didn't bother resetting the counter on this one.

cescoffier: ok

cescoffier (approved):

PR #790, opened 2020-10-04
Improvement of the throttled commit strategy (and make it the default)

cescoffier:
This PR fixes a few issues with the throttled commit strategy:

* commit on partition revocation 
* handle partition assignation when called after the reception of messages
* improve context handling (for performance purpose)


cescoffier:
CC @pcasaes

cescoffier:
This PR does not switch the rebalance listener to be blocking. That will be done in another PR.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -105,11 +104,50 @@ private OffsetStore getOffsetStore(TopicPartition topicPartition) {
      */
     @Override
     public void partitionsAssigned(Set<TopicPartition> partitions) {
-        offsetStores.clear();
+        stopFlushAndCheckHealthTimer();
+
+        // Remove all handled partitions that are not in the given list of partitions
+        for (TopicPartition partition : new HashSet<>(offsetStores.keySet())) {
```
pcasaes, line 110: Why is tuning the keySet into a HashSet necessary?

cescoffier: Because the key set is a live view, and removing from it may trigger a ConcurrentModificationException.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -105,11 +104,50 @@ private OffsetStore getOffsetStore(TopicPartition topicPartition) {
      */
     @Override
     public void partitionsAssigned(Set<TopicPartition> partitions) {
-        offsetStores.clear();
+        stopFlushAndCheckHealthTimer();
+
+        // Remove all handled partitions that are not in the given list of partitions
```
pcasaes, line 109: This is something I got wrong. From the documentation:
> The list of partitions that are now assigned to the consumer (previously owned partitions will NOT be included, i.e. this list will only include newly added partitions)

https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html#onPartitionsAssigned-java.util.Collection-

This needs to be used in conjunction with revoked to get the full picture
> partitions - The list of partitions that were assigned to the consumer and now have been reassigned to other consumers (may not include all currently assigned partitions, i.e. there may still be some partitions left)

https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html#onPartitionsRevoked-java.util.Collection-


If the idea is to remove partitions from the store then that should be done in partitionsRevoked.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -105,11 +104,50 @@ private OffsetStore getOffsetStore(TopicPartition topicPartition) {
...
+                if (largestOffset > -1) {
+                    toCommit.put(partition, new OffsetAndMetadata(largestOffset + 1L, null));
+                }
+            }
+        }
+
+        if (!toCommit.isEmpty()) {
+            // Commit the offsets
+            consumer.getDelegate().commit(toCommit);
+        }
```
pcasaes, line 148: Concerning the comment left on the partitionsAssigned method. After committing we should remove partitions from the store.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -105,11 +104,50 @@ private OffsetStore getOffsetStore(TopicPartition topicPartition) {
...
+            if (!partitions.contains(partition)) {
+                offsetStores.remove(partition);
+            }
+        }
+
+        // We cannot commit the removed offsets as we are not assigned to this partition anymore.
+
+        if (!offsetStores.isEmpty()) {
+            startFlushAndCheckHealthTimer();
+        }
```
pcasaes, line 120: If we are not adding partitions to the store in this method then we need to start the time if either that store or partitions aren't empty.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -282,13 +320,54 @@ boolean hasTooManyMessagesWithoutAck() {
...
-            super("Too Many Messages without Ack");
+            super("Too Many Messages without acknowledgement");
         }
     }
 
     @Override
     public void terminate() {
-        // TODO Force commit sync.
+        commitAllAndAwait();
+        offsetStores.clear();
```
pcasaes, line 330: Nice!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
```diff
@@ -15,24 +19,29 @@ public ContextHolder(Vertx vertx) {
...
+            throw new IllegalArgumentException("Cannot capture the context - not a KafkaReadStreamImpl");
+        } else {
+            try {
+                Field field = KafkaReadStreamImpl.class.getDeclaredField("context");
+                field.setAccessible(true);
+                context = new Context((io.vertx.core.Context) field.get(stream));
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Cannot capture the context", e);
             }
         }
```
pcasaes, line 33: Will we be living with this as afinal solution or are we going to work the Vert.X to get access to the context properly?

pcasaes (changes requested):
Check comments.

The reason why I didn't commit the offsets originally on revoked is because it was tough to reason what would happen with Vert.X async API in respect to any consumer re-balance listener that the user might configure.

It really would help if the consumer re-balance were sync as the original kafka consumer api intends.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: But it's all running in the same Vert.X context, no?

pcasaes: It's my understanding the the Vert.X Kafka Client calls revoked and assigned within the context.

pcasaes:
From the original kafka documentation:

> Under normal conditions, if a partition is reassigned from one consumer to another, then the old consumer will always invoke onPartitionsRevoked for that partition prior to the new consumer invoking onPartitionsAssigned for the same partition.

https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html

But by using the Vert.X async API we lose this guarantee. The new consumer might start up before the old one has committed its offsets. I that happens the new consumer will ignore and override the offsets effectively re-processing the same record.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: No, it can come from the `remove` just below. It's always safer to create a copy if you want to remove. Another solution would be to create a temporary `Set`, collect the keys to remove and remove outside of the loop. Both require an additional set, but this variant avoids another loop.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: Hum... ok, I should not remove the partitions.
However, before we were _Clearing_ the full store was wrong. 

The problem is that in some cases I received records BEFORE this callback is called. This is because of the scheduling in the event loop.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: They have been already removed in `OffsetStore store = offsetStores.remove(partition);`
The commit happens after the removal.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: Yes!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: Yeah, it was wrong.

> This is something I got wrong

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
cescoffier: That's a good question.
We are going to go with this for now.

There are multiple _things_:

* We need to switch to a blocking rebalance listener - this is done using a hack too (see my other branch - that need to be reworked completely BTW). 
* We only use the Vert.x client to get a `Multi` of Record. That part is not that complicated, so starting to see if we need to remove it completely or not. 
* This code is not changed in Vert.x 3.9 nor Vert.x 4.x (for now). 

I've done a list of improvements to applied in the upstream Vert.x client. Exposing the context needs to be added. Generally, we don't do this in Vert.x as we expect the client creation to be called form an event loop (and so you can capture the context from a creation site), but in our case, it's not the case, and the client creates its own context.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: Gotcha!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: I see. nice!

cescoffier:
> From the original kafka documentation:
> 
> > Under normal conditions, if a partition is reassigned from one consumer to another, then the old consumer will always invoke onPartitionsRevoked for that partition prior to the new consumer invoking onPartitionsAssigned for the same partition.
> 
> [kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html](https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html)
> 
> But by using the Vert.X async API we lose this guarantee. The new consumer might start up before the old one has committed its offsets. If that happens the new consumer will ignore and override the offsets, effectively re-processing the same record.

Exactly! I had to change some test to relax assertions as this case happened more often than anticipated. It also has another issue. I started receiving records before receiving the assigned partition. 

Once this one is merged, I will start working on the blocking listeners.

cescoffier:
> The reason why I didn't commit the offsets originally on revoked is because it was tough to reason what would happen with Vert.X async API in respect to any consumer re-balance listener that the user might configure.
> 
> It really would help if the consumer re-balance were sync as the original kafka consumer api intends.

Do you mean that in this case it's delegated to the user's code? 
We can do the commit only if there is no custom rebalance listener (and check that the doc clearly indicate that the listener must commit).

pcasaes:
>  I started receiving records before receiving the assigned partition.

Yeah! Bane of my existence! That's why on the consumer re-balance work we call the pause and fetch methods. But that' won't work here since we really do need to block the poll call.

pcasaes:
> > The reason why I didn't commit the offsets originally on revoked is because it was tough to reason what would happen with Vert.X async API in respect to any consumer re-balance listener that the user might configure.
> > It really would help if the consumer re-balance were sync as the original kafka consumer api intends.
> 
> Do you mean that in this case it's delegated to the user's code?
> We can do the commit only if there is no custom rebalance listener (and check that the doc clearly indicates that the listener must commit).

We'd just have to give it a good think. We are running the commit handler's re-balance before the user. If the user performs any async operations on the kafka consumer it should happen after what we do since it will be in the same context. So in this case if the user fetches the offsets it should be what we committed (**maybe we can test for this**).

I'm not really seeing any issues at this moment. If the user is using re-balance to self manage offsets then they really should be using the **ignore** strategy (it's what I do). Maybe we should update the docs, in the example where we set up TTL based Kafka backed cache we should inform the user that they should configure **ignore**
documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRebalancedConsumerRebalanceListener.java

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
pcasaes: > The problem is that in some cases I received records BEFORE this callback is called. 

Perhaps we should use the pause/fetch strategy we use for the user supplied consumer re-balance as well. In this case we'd have to return a Uni.

cescoffier: That would not always work, I've seen this issue when the application starts, where we are not in a pause/resume cycle. 

Let's revisit this when we have the blocking callbacks. We are trying to work around an issue that should not be an issue...

cescoffier:
I've updated the javadoc of `io.smallrye.reactive.messaging.kafka.KafkaConsumerRebalanceListener` to be more specific.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: Done!

pcasaes (approved):

cescoffier:
Merging this one! The next one about blocking rebalance listeners is going to be a lot of... Not!

pcasaes:
I'm excited to see it. The draft or is looking good

PR #772, opened 2020-09-23
Throttling Commit Strategy - toward being the default

cescoffier:
This PR is not intended to be merged for now. It includes breaking changes

- Make the rebalance listeners blocking.

Next step: extend tests.

**Breaking Changes:**

* Change the `KafkaRebalanceListener` interface to be blocking

**Hacks:** 

* Replace the built-in rebalance listener used by Vert.x by ours as the Vert.x one does not follow the rebalance protocol correctly. 

CC @pcasaes

PR #797, opened 2020-10-08
Make the rebalance listeners blocking 

cescoffier:
- change the method signatures to make rebalance listener blocking (**breaking change**)
- substitute the internal rebalance listener used by the Vert.x client by our own
- invoke the rebalance listener from the Kafka polling threading thread.

~~The synchronization protocol in the throttled commit strategy is not completed yet.~~

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
@@ -1,6 +1,12 @@
 package io.smallrye.reactive.messaging.kafka.commit;
 
 import java.lang.reflect.Field;
+import java.time.Duration;
+import java.util.concurrent.*;
+
+import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.common.errors.WakeupException;
 
 import io.vertx.kafka.client.consumer.KafkaReadStream;
 import io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl;
@@ -13,10 +19,13 @@ import io.vertx.mutiny.core.Vertx;
 public class ContextHolder {
 
     protected final Vertx vertx;
+    private final int timeout;
     protected volatile Context context;
 
-    public ContextHolder(Vertx vertx) {
+    public ContextHolder(Vertx vertx, KafkaConnectorIncomingConfiguration configuration) {
         this.vertx = vertx;
+        this.timeout = configuration.config().getOptionalValue(ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG, Integer.class)
+            .orElse(60000);
     }
 
     public void capture(KafkaReadStream<?, ?> stream) {
@@ -42,6 +51,20 @@ public class ContextHolder {
             runnable.run();
         } else {
             context.runOnContext(x -> runnable.run());
+        }
+    }
+
+    public <T> T runOnContextAndAwait(Callable<T> action) {
+        FutureTask<T> task = new FutureTask<>(action);
+        runOnContext(task);
+
+        try {
+            return task.get(timeout, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new WakeupException();
+        } catch (ExecutionException | TimeoutException e) {
+            throw new CompletionException(e);
         }
     }
 
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
@@ -1,11 +1,8 @@
 package io.smallrye.reactive.messaging.kafka.commit;
 
 import java.lang.reflect.Field;
-import java.time.Duration;
 import java.util.concurrent.*;
 
-import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.errors.WakeupException;
 
 import io.vertx.kafka.client.consumer.KafkaReadStream;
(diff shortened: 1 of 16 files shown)


cescoffier:
CC @pcasaes (WIP)

Supersedes #772

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -38,7 +38,7 @@
 
     private static final Map<String, Map<Integer, TopicPartition>> TOPIC_PARTITIONS_CACHE = new ConcurrentHashMap<>();
 
-    private final Map<TopicPartition, OffsetStore> offsetStores = new HashMap<>();
+    private final Map<TopicPartition, OffsetStore> offsetStores = new ConcurrentHashMap<>();
```
pcasaes, line 41: This could be a problem as OffsetStore is not thread safe and was designed to be called within the context thread.

In `partitionsRevoked` we call `clearLesserSequentiallyProcessedOffsetsAndReturnLargestOffset` which mutates the OffsetStore. Is it not possible to make a blocking call in the context thread from the poll thread?

pcasaes: I ran the tests locally and they passed. What I did was to revert the `offsetStores` map back to a HashMap and ran the relevant parts of `partitionsRevoked` within the context. I'll post my changes later.

cescoffier: no partitionRevoked must run completely from the polling thread. It cannot run from the event loop. When the method returns, everything must have been completed or failed. 

One thing I was thinking is to block the polling thread and use a countdown latch decreased from the event loop. However, it can introduce deadlocks.

pcasaes: I just used a FutureTask to manipulate the OffsetStore and fed it into the context and waited for a response on the poll thread. Seemed to work. Going to do some more tests.

pcasaes: Here what I tested
https://github.com/pcasaes/smallrye-reactive-messaging/commit/3db01661479eb0f3404adb921a5cb8d7c8c66488

Notice there are a couple of FIXMEs.

I did the run the tests as well as manual testing. Worked well.

I did pickup a small bug which I'll open a separate PR for. During revoke we should be committing the offsets of those revoked partitions. There's a `not contains` which should `contains`.

cescoffier: I've rebased the branch and updated to fit the synchronization protocol.
The FutureTask is a great idea!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
```diff
@@ -45,4 +50,18 @@ public void runOnContext(Runnable runnable) {
...
 
+    public <T> T runOnContextAndAwait(Callable<T> action) {
+        FutureTask<T> task = new FutureTask<>(action);
+        runOnContext(task);
+
+        try {
+            return task.get(timeout, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new WakeupException();
```
pcasaes, line 61: Are we sure about throwing a WakeupException? It will stop the poll thread dead cold. I used it without giving it much thought on my tests. The other option is to wrap it in a CompletionException along with the others (besides marking the current thread as interrupted). 

Dealing with InterruptedException is always tricky because you never know where they can come from.

pcasaes:
Looking really good! I just had a question about how to handle the InterruptedException.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/ContextHolder.java
cescoffier: So, I checked the Apache Kafka client, and it's how it's called:

```java
 private Exception invokePartitionsAssigned(final Set<TopicPartition> assignedPartitions) {
        log.info("Adding newly assigned partitions: {}", Utils.join(assignedPartitions, ", "));

        ConsumerRebalanceListener listener = subscriptions.rebalanceListener();
        try {
            final long startMs = time.milliseconds();
            listener.onPartitionsAssigned(assignedPartitions);
            sensors.assignCallbackSensor.record(time.milliseconds() - startMs);
        } catch (WakeupException | InterruptException e) {
            throw e;
        } catch (Exception e) {
            log.error("User provided listener {} failed on invocation of onPartitionsAssigned for partitions {}",
                listener.getClass().getName(), assignedPartitions, e);
            return e;
        }

        return null;
    }
```

The `CompletionException` we are throwing would be handled by the second `catch` block and denotes a problem in the listener. The listener is the source of the issue. So, this case is fine. 

As you mentioned, an interruption can come from anywhere, and it should not be handled by the second block, but by the first one. I believe we should throw an `InterruptException` (note that it's a Kafka-specific unchecked variant of `InterruptedException`) and let the polling thread handle the interruption.

cescoffier: Updated the code.

cescoffier:
Finally... removing the "draft".

I guess it should be fine now. We may want to improve the documentation. I did the first pass in the last release, but the commit concept is not easy to understand, especially if you are used to regular message acknowledgment. I'm planning to write a blog post about that. We may be able to reuse some content.

cescoffier:
The rebalance listener documentation needs to be updated.

pcasaes (approved):
Lgtm. I second better explaining what's going on. In the end the user will need to have a good understanding of kafka to use it appropriately.

cescoffier:
Yes, the documentation will need an update. The same applies to the failure-strategy. This week, I will write the failure strategy one and then the commit strategy.

PR #798, opened 2020-10-09
Remove and commit revoked partitions from the throttled offset store

pcasaes:
@cescoffier Noticed this bug when testing #797 out.

We should be removing the revoked partitions from the store and committing any offset we have for those.


cescoffier (approved):
LGTM and avoid a collection copy

PR #799, opened 2020-10-13
Reduce hard dependencies on OpenTelemetry

kabir:
for environments where it is not available and tracing is disabled


kabir:
Seems I had a stale checkout. I have rebased now

PR #800, opened 2020-10-13
update AmqpCreditTest to use a test server rather than full broker

gemmellr:
Updates test of credit handling for outgoing/sending (AmqpCreditTest) to use a test server rather than a full broker. Also renames the test base class to represent its use of a broker, and adds (presents as a modification) a simpler base for tests that dont, more of which could be transitioned over time.

Using a test server allows for much more specific tests of what the connector actually does than observing its effect from the other side of a broker intermediary does, with the bonus of also being much faster due to having fewer parts and doing less work overall. In this particular case, running AmqpCreditTest in isolation the observed time taken was reduced from ~13sec to ~1.5sec.


cescoffier (approved):
LGTM! Thanks!

PR #801, opened 2020-10-13
Fix how configuration property coming from environment variables are handled

cescoffier:
Fix how are handled configuration properties defined in environment variables (uppercase with underscores). Unfortunately, how we extract the Kafka object configuration (consumer, producer, admin) is based on the property names which was filtered to only contain the properties for the specific channel. In the case of env var, the keys were filtered out as they were not matching the expected prefix. This PR fixes this. 

Fix https://github.com/quarkusio/quarkus/issues/12678

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -91,8 +91,14 @@
 
     <dependency>
       <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>compile</scope>
+      <version>${version.junit5}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
-      <version>5.7.0</version>
       <!-- used in snippets -->
       <scope>compile</scope>
     </dependency>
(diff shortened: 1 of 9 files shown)


documentation/pom.xml
```diff
@@ -89,10 +89,16 @@
...
 
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>compile</scope>
+      <version>${version.junit5}</version>
+    </dependency>
+
     <dependency>
       <groupId>org.junit.jupiter</groupId>
```
kenfinnigan, line 100: Is the API dependency not a transitive of the one below?

kenfinnigan (commented):
Is this change needed because we're doing things outside the regular config chain?

I would've expected environment variable handling to be provided from SRye Config?

documentation/pom.xml
cescoffier: Yes, because the inherited dependency is in the test scope.

cescoffier:
Yes and no. Unfortunately how property names are handled is pretty limited. If you have an end var names FOO_BAR, it could match foo.bar, boo-bar, foo_bar, FOO_BAR... so asking the keys from the source is not ideal in our case as Kafka expect its own key.

During my kafka spree, I started thinking about implementing a Map (that's how kafka gets its config) delegating on the underlying Config object. But, I'm not done as it needs caching (for perf reason) and it does not play nicely with the vertx client using JSON object and not maps.

kenfinnigan (approved):

documentation/pom.xml
cescoffier: The main pom has jupiter and jupiter-api in the test scope.

PR #805, opened 2020-10-14
Set the consumer/producer/admin client.id when not set in the config

cescoffier:
That makes consuming the metrics much simpler.


ebullient (approved):
While I am not going to claim to understand the -1 bits, clearing up the declared client ids and other attributes makes a lot of sense. Cool. ;)

PR #811, opened 2020-10-15
Improve health check reporting for application's failure

cescoffier:
Improve the health check reporting for application failure. 
In addition, close the Kafka consumer if the failure is fatal.﻿

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
@@ -41,6 +41,10 @@ public class HealthCenter {
         failures.add(new ReportedFailure(source, cause));
     }
 
+    public void reportApplicationFailure(String method, Throwable cause) {
+        failures.add(new ReportedFailure("application-" + method, cause));
+    }
+
     public static class ReportedFailure {
         final String source;
         final Throwable failure;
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
```diff
@@ -41,6 +41,10 @@ public void report(String source, Throwable cause) {
         failures.add(new ReportedFailure(source, cause));
     }
 
+    public void reportApplicationFailure(String method, Throwable cause) {
+        failures.add(new ReportedFailure("application-" + method, cause));
```
kenfinnigan, line 45: Does "application" as a prefix make sense given it's combined with the method name? Is that going to be confusing?

kenfinnigan (approved):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
cescoffier: Yes and no, I tried with the method name only and found it even more confusing. The `application` prefix indicates it's a code issue.

cescoffier: Anyway, let's see and we can revisit it later. The important part is that it's reported.

PR #817, opened 2020-10-18
Fix log message parameter order

pcasaes:
@cescoffier Sorry about that. A last minute change to the log message broke the parameter order. This caused an IllegalArgumentException.

Now it's fine. Message example

```
2020-10-18 16:51:15,325 WARN  [io.sma.rea.mes.kafka] (vert.x-eventloop-thread-0) trace_id  span_id  SRMSG18239: Acked record 1018 on group 'group1' was ignored because the topic partition 'TopicPartition{topic=TEST_TOPIC, partition=0}' was revoked for this instance. Record will likely be processed again.
```


cescoffier:
I need to invest some time to cover this with some tests. Unfortunately, it's not that simple and the mock consumer does not call the rebalance listener callbacks.

cescoffier (approved):

cescoffier (approved):

PR #822, opened 2020-10-20
Use throwable.toString on reason and cause if message is null

pcasaes:
@cescoffier 

Getting an unchecked exception when producing to DLQ if the reason or cause messages are null.

```
Caused by: java.lang.NullPointerException
	at io.vertx.core.buffer.impl.BufferImpl.<init>(BufferImpl.java:52) ~[vertx-core-3.9.4.jar:3.9.4]
	at io.vertx.core.buffer.impl.BufferImpl.<init>(BufferImpl.java:56) ~[vertx-core-3.9.4.jar:3.9.4]
	at io.vertx.core.buffer.impl.BufferFactoryImpl.buffer(BufferFactoryImpl.java:39) ~[vertx-core-3.9.4.jar:3.9.4]
	at io.vertx.core.buffer.Buffer.buffer(Buffer.java:72) ~[vertx-core-3.9.4.jar:3.9.4]
	at io.vertx.kafka.client.producer.impl.KafkaHeaderImpl.<init>(KafkaHeaderImpl.java:36) ~[vertx-kafka-client-3.9.4.jar:3.9.4]
	at io.vertx.kafka.client.producer.impl.KafkaProducerRecordImpl.addHeader(KafkaProducerRecordImpl.java:139) ~[vertx-kafka-client-3.9.4.jar:3.9.4]
	at io.vertx.mutiny.kafka.client.producer.KafkaProducerRecord.addHeader(KafkaProducerRecord.java:148) ~[smallrye-mutiny-vertx-kafka-client-1.2.2.jar:?]
	at io.smallrye.reactive.messaging.kafka.fault.KafkaDeadLetterQueue.handle(KafkaDeadLetterQueue.java:87) ~[classes/:?]
	at io.smallrye.reactive.messaging.kafka.IncomingKafkaRecord.nack(IncomingKafkaRecord.java:132) ~[classes/:?]
```

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
@@ -80,13 +80,21 @@ public class KafkaDeadLetterQueue implements KafkaFailureHandler {
         }
     }
 
+    private String getThrowableMessage(Throwable throwable) {
+        String text = throwable.getMessage();
+        if (text == null) {
+            text = throwable.toString();
+        }
+        return text;
+    }
+
     @Override
     public <K, V> CompletionStage<Void> handle(
             IncomingKafkaRecord<K, V> record, Throwable reason) {
         KafkaProducerRecord<K, V> dead = KafkaProducerRecord.create(topic, record.getKey(), record.getPayload());
-        dead.addHeader(DEAD_LETTER_REASON, reason.getMessage());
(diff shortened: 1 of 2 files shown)


cescoffier:
How did you end up with a `null` reason? `nack` should get a reason.

cescoffier (approved):

cescoffier:
Just got it... `NullPointerException` don't have messages.

cescoffier:
Thanks!

pcasaes:
It got a reason. But the reason had no message. In this case an npe.

pcasaes:
Actually. In the test case the reason is an illegal argument exception with no message. This causes an npe trying to add the kafka header. 

An npe is an exception that often has no message so it would also fail to write to the dlq.

PR #824, opened 2020-10-20
Fix Kafka producer concurrency settings

cescoffier:
Do not limit the processor request with the Kafka `max.in.flights.requests.per.connection`, as that setting is per batch or per message. 

This improves a lot (* 100) the throughput of the provider.


cescoffier:
CC @pcasaes

kenfinnigan (approved):

pcasaes:
If I understood correctly this completely decouple Kafka's `max.in.flight.requests.per.connection` from the stream producer?

pcasaes:
Wow. just performed a test here. Huge difference!

cescoffier:
@pcasaes Yes, that link was a misconception as the _request_ is actually the number of sent batches, not messages.  The situation is much better now where the `batch.size` and `linger.ms` are controlling when batches are sent. Depending on the size of messages and latency, you can achieve max throughput by configuring the `max-inflight`, `batch.size`, and `linger.ms`.

PR #832, opened 2020-10-24
cloud-events defined twice

### LINKED ISSUE #831: mp.messaging.connector.smallrye-kafka.cloud-events listed twice in application.properties
**Describe the bug**
(Describe the problem clearly and concisely.)
We noticed some tests failures in Quarkus Tools for IntelliJ after we updated our test applications to 1.9.0.Final. The problem is that the property `mp.messaging.connector.smallrye-kafka.cloud-events` is listed twice

**Expected behavior**
(Describe the expected behavior clearly and concisely.)
Property should be listed once only

**Actual behavior**
(Describe the actual behavior clearly and concisely.)
Property is listed twice

**To Reproduce**

Link to a small reproducer (preferably a Maven project if the issue is not Gradle-specific).
https://code.quarkus.io/d?g=org.acme&a=code-with-quarkus&v=1.0.0-SNAPSHOT&b=MAVEN&s=9qf&cn=code.quarkus.io

Or attach an archive containing the reproducer to the issue.

1. Create a Quarkus application with SmallRye Reactive Messaging - Kafka Connector extension
2. open application in IntelliJ or VSCode
3. open application.properties
4. enter Ctrl+Space to activate code assist

**Configuration**

# Add your application.properties here, if applicable.

**Screenshots**
(If applicable, add screenshots to help explain your problem.)

quarkus-kafka

jeffmaury:
Fixes #831

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -84,7 +83,7 @@
...
 @ConnectorAttribute(name = "max-inflight-messages", type = "long", direction = Direction.OUTGOING, description = "The maximum number of messages to be written to Kafka concurrently. It limits the number of messages waiting to be written and acknowledged by the broker. You can set this attribute to `0` remove the limit", defaultValue = "1024")
-@ConnectorAttribute(name = "cloud-events", type = "boolean", direction = Direction.OUTGOING, description = "Enables (default) or disables the Cloud Event support. If enabled, the connector sends the outgoing messages as Cloud Event if the message to be sent includes Cloud Event Metadata.", defaultValue = "true")
+@ConnectorAttribute(name = "cloud-events", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Enables (default) or disables the Cloud Event support. If enabled, the connector sends the outgoing messages as Cloud Event if the message to be sent includes Cloud Event Metadata and analyzes the incoming records and try to create Cloud Event metadata.", defaultValue = "true")
```
kenfinnigan, line 86: It's probably a good idea to move the attribute definition to the top where the other Incoming and Outgoing attributes are defined

gsmet: And maybe a test should check that there are no duplicates? It shouldn't be too hard to implement.

cescoffier: This is not a duplicate (duplicate would cause compilation errors):


We have:
```
@ConnectorAttribute(name = "cloud-events", type = "boolean", direction = Direction.INCOMING, description = "Enables (default) or disables the Cloud Event support. If enabled, the connector analyzes the incoming records and try to create Cloud Event metadata.", defaultValue = "true")


@ConnectorAttribute(name = "cloud-events", type = "boolean", direction = Direction.OUTGOING, description = "Enables (default) or disables the Cloud Event support. If enabled, the connector sends the outgoing messages as Cloud Event if the message to be sent includes Cloud Event Metadata.", defaultValue = "true")
```

The directions are different. The first one is for INCOMING, the second one is for OUTGOING. It's not the same attribute. They share the same name but are inherently different attributes.

cescoffier (changes requested):
What's the reason to merge two different attributes? 
The description would need to be extended to explain that the behavior for `incoming` and `outgoing` is different.

cescoffier:
About the CI: you need to do a full run (you can skip tests, don't worry):

```
mvn clean install -DskipTest
```

This will update the documentation with the new description.

jeffmaury:
A little clarification for this modification: as explained in the animated gif, the microprofile tooling (Eclipse, IntelliJ, VSCode) is performing reflection on the connector class to retrieve attributes and map them to property names. Having a single attribute for both direction will prevent tooling to generate 2 properties with the same name.
Also adding @angelozerr to the discussion

kenfinnigan:
@jeffmaury I appreciate multiple named configuration can clutter the IDE.

But how can it distinguish the meaning for the two usages?

@cescoffier is it also possible a user would only want it enabled in one direction? ie. Cloud Event coming in but not out?

If that's a valid use case, then it prevents it from being used as a single attribute. Maybe each attribute needs to be named slightly differently?

cescoffier:
@kenfinnigan yes, there are use cases: gateway receiving raw events and transforming them into Cloud event where the methods are adding the Cloud event metadata. The opposite is also possible (even if I don't see a use case for that).

That being said, even if you enable cloud events globally, we could still handle that use case. It will just get some performance penalties as it will try for every event to check if it's a Cloud event or not. Unfortunately, the Cloud event spec is made in a way that the only way to detect is pretty expensive (basically: look inside and if it looks like a cloud event, it's a cloud event...).

cescoffier:
superseded by #906

PR #867, opened 2020-11-12
Improve reporting when too many records are not committed

### LINKED ISSUE #865: Report topic / partition / offset in Exceptions
E.g. in exceptions like `io.smallrye.reactive.messaging.kafka.commit.KafkaThrottledLatestProcessedCommit$TooManyMessagesWithoutAckException`, it would be very helpful to get the topic, partition and offset that hasn't been acked for too long. This helps investigating the issues.

Also see https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Reactive.20Messaging.20Kafka.3A.20Speed.20up.20event.20receiving/near/216371150 and above.

cescoffier:
Fix #865

﻿Improve reporting when too many message have not been acked or nacked.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
@@ -13,6 +13,7 @@ import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
 import io.smallrye.reactive.messaging.kafka.impl.KafkaSource;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
+import io.vertx.core.impl.NoStackTraceThrowable;
 import io.vertx.kafka.client.common.TopicPartition;
 import io.vertx.kafka.client.consumer.OffsetAndMetadata;
 import io.vertx.mutiny.core.Vertx;
@@ -292,11 +293,16 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
         }
 
         if (this.unprocessedRecordMaxAge > 0) {
-            offsetStores
-                    .values()
-                    .stream()
-                    .filter(OffsetStore::hasTooManyMessagesWithoutAck)
-                    .forEach(o -> this.source.reportFailure(new TooManyMessagesWithoutAckException(), true));
+            for (OffsetStore store : offsetStores.values()) {
+                if (store.hasTooManyMessagesWithoutAck()) {
+                    long lastOffset = store.getLastCommittedOffset();
+                    String topic = store.topicPartition.getTopic();
+                    int partition = store.topicPartition.getPartition();
+                    TooManyMessagesWithoutAckException exception = new TooManyMessagesWithoutAckException(topic, partition,
+                            lastOffset);
+                    this.source.reportFailure(exception, true);
+                }
+            }
         }
 
     }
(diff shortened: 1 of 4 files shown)


cescoffier:
@andreas-eberle Can you have a look?

andreas-eberle:
@cescoffier : I did build it locally and tried to include it my Quarkus project. Unfortunately, I get some weired errors about `java.lang.ClassNotFoundException: io.quarkus.agroal.runtime.LegacyDataSourcesJdbcBuildTimeConfig`. So I haven't managed to test it in my project.

Is there a recommended way to exchange a single dependency with a snapshot version?

cescoffier:
you need to build quarkus completely or use the 1.10.1.CR1 and update the connector dependency explicitly.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
```diff
@@ -139,8 +139,11 @@
     void configuredPattern(String channel, String pattern);
 
     @LogMessage(level = Logger.Level.WARN)
-    @Message(id = 18231, value = "The amount of received messages without acking is too high for topic partition '%s', amount %d.")
-    void receivedTooManyMessagesWithoutAcking(String topicPartition, long amount);
+    @Message(id = 18231, value = "The amount of received messages without acking is too high for topic partition '%s', "
+            + "amount %d. The last committed offset was %d. It means that the Kafka connector received Kafka Records that "
+            + "have neither be acked nor nacked in a timely fashion. The connector accumulates records in memory, but that "
+            + "buffer reached its maximum capacity. The connector cannot commit as a record processing has not completed.")
```
andreas-eberle, line 145: I'm not sure if the remark about the maximum capacity is always true. In my case, only the timeout hit (I have 89 messages waiting in batching). Not sure if you should change the wording.

andreas-eberle:
@cescoffier: I can confirm the message shows the last offset.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -355,10 +366,14 @@ long clearLesserSequentiallyProcessedOffsetsAndReturnLargestOffset() {
                         break;
                     }
                     unProcessedTotal--;
-                    largestSequentialProcessedOffset = receivedOffsets.poll().getOffset();
+                    OffsetReceivedAt poll = receivedOffsets.poll();
+                    if (poll != null) {
+                        largestSequentialProcessedOffset = poll.getOffset();
+                    }
```
pcasaes, line 372: This null check isn't strictly necessary since the code block is in a while received offset is not empty. But this is fine.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
cescoffier: You are right, there is also a deadline.
I fixed it.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
cescoffier: Yes, some weird issues have been reported. So, better to be safe...

andreas-eberle:
Looks good to me. Thanks for the quick implementation @cescoffier.

PR #870, opened 2020-11-16
Prevent JsonHelper from swallowing Booleans

lorenzleutgeb:
I was trying to pass a `Boolean` value as Kafka configuration. Very unexpectedly it disappeared after `asJsonObject`. Since JSON has the usual notion of booleans, I think it is only natural for `asJsonObject` to support that.

If you decide that you do not want to accept this MR, I think you should instead have `asJsonObject` log the skipped properties.

### DIFF
--- a/smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/JsonHelper.java
+++ b/smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/JsonHelper.java
@@ -27,6 +27,14 @@ public class JsonHelper {
             }
 
             try {
+                boolean b = config.getValue(key, Boolean.class);
+                json.put(key, b);
+                continue;
+            } catch (ClassCastException | IllegalArgumentException e) {
+                // Ignore me
+            }
+
+            try {
                 String value = config.getValue(key, String.class);
                 if (value.trim().equalsIgnoreCase("false")) {
                     json.put(key, false);
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/JsonHelper.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/JsonHelper.java
@@ -39,6 +39,16 @@ public class JsonHelper {
             }
 
             try {
+                Optional<Boolean> b = config.getOptionalValue(originalKey, Boolean.class);
+                if (b.isPresent()) {
+                    json.put(key, b.get());
+                    continue;
+                }
+            } catch (ClassCastException | IllegalArgumentException e) {
+                // Ignore me
+            }
+
+            try {
                 String value = config.getOptionalValue(originalKey, String.class).orElse("").trim();
                 if (value.equalsIgnoreCase("false")) {
                     json.put(key, false);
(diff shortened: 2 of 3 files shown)


lorenzleutgeb:
Erm... it [looks like this breaks things](https://github.com/smallrye/smallrye-reactive-messaging/runs/1408758316?check_suite_focus=true#step:5:59555). That's unexpected again...

smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/JsonHelper.java
```diff
@@ -26,6 +26,14 @@ public static JsonObject asJsonObject(Config config) {
                 // Ignore me
             }
 
+            try {
+                boolean b = config.getValue(key, Boolean.class);
```
cescoffier, line 30: This returns "false" for whatever is not "true". 
That explains the next branch where we compare it with "false" (line 39)

cescoffier (changes requested):

smallrye-reactive-messaging-http/src/main/java/io/smallrye/reactive/messaging/http/JsonHelper.java
lorenzleutgeb: Yes, I thought so, judging from the failed test. I am not that familiar with the MP Config API, so I would expect this to return empty in case the value is not a boolean. Turns out my intuition about this API was wrong (too?), because I did not read about "Automatic Converters".

So, how to proceed? Would the following work?

```java
Object o = config.getValue(key, Object.class);
if (o instanceof Boolean) {
  ...
}
else if (o instanceof Integer) {
  ...
}
else if (o instanceof String) {
  ...
}
...
```

Also, feel free to close the PR if you like.

cescoffier:
Good question. I don’t know, worth a try.

What was missing in the configuration? You mentioned Boolean, do you have
an example?

lorenzleutgeb:
> What was missing in the configuration? You mentioned Boolean, do you have an example?

I am using Protocol Buffers (deserializer implementation by Confluent), and wanted to set [`derive.type`](https://github.com/confluentinc/schema-registry/blob/ea9d630457701cbb63c89ebc87c315e210f7ac96/protobuf-serializer/src/main/java/io/confluent/kafka/serializers/protobuf/KafkaProtobufDeserializerConfig.java#L34) to `true`.

cescoffier:
Why wouldn't your case be handle by:


```
String value = config.getOptionalValue(originalKey, String.class).orElse("").trim();
if (value.equalsIgnoreCase("false")) {
  json.put(key, false);
} else if (value.equalsIgnoreCase("true")) {
  json.put(key, true);
} else {
 json.put(key, value);
}
```

lorenzleutgeb:
Because I did:

```java
  @Produces
  @ApplicationScoped
  @Named("default-kafka-broker")
  public static Map<String, Object> produce(Config config) {
    return Map.of(
        SCHEMA_REGISTRY_URL_CONFIG,
        config.getValue(PREFIX + ".kafka." + SCHEMA_REGISTRY_URL_CONFIG, String.class),
        VALUE_SUBJECT_NAME_STRATEGY,
        ConfigurableNameStrategy.class.getName(),
        DERIVE_TYPE_CONFIG,
        true,
        REFERENCE_SUBJECT_NAME_STRATEGY_CONFIG,
        ConfigurableNameStrategy.class.getName());
  }
```

which does not work. Because the value for key `DERIVE_TYPE_CONFIG` gets lost. To work around the problem, I did

```diff
        DERIVE_TYPE_CONFIG,
-        true,
+        "true", // Booleans don't work
```

But I decided to raise this PR because my guess is that I will not be the only to ever try putting a `Boolean`...

cescoffier:
Thanks, I’m want to see what we get in this case. I should be able to fix
it.

cescoffier:
Can you look at #https://github.com/smallrye/smallrye-reactive-messaging/pull/872?

PR #880, opened 2020-11-30
Copy header from the origin header in dead letter queue

loicmathieu:


loicmathieu:
I didn't add any test because I'm not sure where to put it.
Also, I'm not fan of the new String creation, if you have better solution (maybe via a Buffer) please provide some advice.

cescoffier (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
```diff
@@ -102,6 +102,7 @@ private String getThrowableMessage(Throwable throwable) {
         dead.addHeader(DEAD_LETTER_TOPIC, record.getTopic());
         dead.addHeader(DEAD_LETTER_PARTITION, Integer.toString(record.getPartition()));
         dead.addHeader(DEAD_LETTER_OFFSET, Long.toString(record.getOffset()));
+        record.getHeaders().forEach(header -> dead.addHeader(header.key(), new String(header.value())));
```
cescoffier, line 105: We can introduce huge string creation and even invalid strings
Can you switch to: 

```
record.getHeaders().forEach(header -> dead.addHeader(KafkaHeader.header(header.key(), Buffer.buffer(header.value()))));
```

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
loicmathieu: done

cescoffier (approved):

cescoffier:
Thanks!

PR #885, opened 2020-12-04
Implement getNack for IncomingKafkaRecord

pcasaes:
When using a **RecordConverter**  the incoming message is converted to a new message but maintaining the same **ack** and **nack** functions by calling the incoming message's **getAck()** and **getNack(Throwable)** methods:

```java
    default <P> Message<P> withPayload(P payload) {
        return of(payload, Metadata.from(this.getMetadata()), this.getAck(), this.getNack());
    }
```

IncomingKafkaRecord does not override **getNack(Throwable)** so its failure strategies aren't applied.

This can be easily verified by using the tuple class **io.smallrye.reactive.messaging.kafka.Record** as the message payload. This class invokes the use of **RecordConverter**.


cescoffier (approved):

cescoffier:
Thanks!

PR #887, opened 2020-12-04
Ignore failure during rebalance

### LINKED ISSUE #866: Kafka: Liveness probe fails due to a rebalance causing the commit to fail
Shortly after the start of my service, I received the following log statement:

``` 
2020-11-12 12:05:56,025 INFO  [io.sma.health] (vert.x-worker-thread-12) SRHCK01001: Reporting health down status: {"status":"DOWN","checks":[{"name":"SmallRye Reactive Messaging - liveness check","status":"DOWN","data":{"my-events":"[KO] - Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation."}}]}
```

Processing continued fine after this until the liveness probe failed two more times (same log statement as shown above) and thus kubernetes took down the pod and restarted it. 

I have no other consumers in the group of this consumer. 

Also see [zulip the conversation](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Reactive.20Messaging.20Kafka.3A.20Speed.20up.20event.20receiving/near/216459364) and below.

cescoffier:
Fix #866

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
@@ -130,6 +130,10 @@ public interface KafkaLogging extends BasicLogger {
     @Message(id = 18228, value = "A failure has been reported for Kafka topics '%s'")
     void failureReported(Set<String> topics, @Cause Throwable t);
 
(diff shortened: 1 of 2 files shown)


cescoffier:
@loicmathieu by any chance, can you try with this PR?

loicmathieu:
It would be difficult to test it as the issue only occurs under load inside my k8s cluster. 
And the build of the container is automated so I cannot used locally built library

Issue #879, opened 2020-11-30
Mediator gets discovered twice, leading to WeavingException


lorenzleutgeb:
This is on version 2.0.4. I am not sure whether what I am seeing is a bug or I am doing something wrong. I am getting

```
io.smallrye.reactive.messaging.WeavingException: Synchronous error caught during the subscription of [...]
	at io.smallrye.reactive.messaging.SubscriberMediator.run(SubscriberMediator.java:125)
	...
Caused by: java.lang.IllegalStateException: Already subscribed
	...
```

So I started investigating. By adding a breakpoint to the lambda in `MediatorManager#analyze` that adds objects to `#collected`. It is called twice, the two methods being:

```
[BackedAnnotatedMethod] @Incoming @Blocking @Merge public volatile com.example.Foo.onMessage(Object)
[BackedAnnotatedMethod] @Incoming @Blocking @Merge public          com.example.Foo.onMessage(MyMessageType)
```

I have no clue where that `volatile` method that takes an `Object` comes from. The declaring type is the same (reference) for both. Any clues?

cescoffier:
It looks like you have an overloaded method (same name, different signature). Can you try to change one of the method names to check if that's the issue?

lorenzleutgeb:
Hmm, my `Foo` class implements an interface. If I remove that, it works. It goes like this:

```
@ApplicationScoped
class Foo implements MyMessageTypeListener {
  @Override
  @Multi
  @Blocking
  @Incoming(...)
  public void onMessage(MyMessageType message) { ... }
}

interface MyMessageTypeListener implements SimpleListener<MyMessageType> {}

@FunctionalInterface
interface SimpleListener<T> extends Listener<T> {
  void onMessage(T message);
}
```

cescoffier:
Can you try removing the type <T> from your listener. That may be the real issue. I need to check how we find and identify the method.

lorenzleutgeb:
> Can you try removing the type from your listener. That may be the real issue. I need to check how we find and identify the method.

You mean the generic type argument `T` in `Listener`? No, not really. We use this as a "marker" interface. We have many beans that have `@Incoming` methods (all with different responsibilities). By using that interface we force them all to have a uniform name (`onMessage`) and if we want to list them we can just look up all implementations of that interface. If I remove the generic type argument, then we cannot do all of that.

cescoffier:
That was just to pinpoint the issue - not a workaround.

[...]

PR #892, opened 2020-12-07
Implementation of a new wiring algorithm

### LINKED ISSUE #414: NPE when broadcasting emitted messages to SSE
The following example:

```java
  @Inject
  @Channel("issues")
  Emitter<Status> emitter;

  @Inject
  @Channel("issues-out")
  Publisher<Status> status;

  @Incoming("issues")
  @Outgoing("issues-out")
  @Broadcast
  public Status publish(Status status) {
     return status;
  }

   @GET
   @Path("/issues/status")
   @Produces(MediaType.SERVER_SENT_EVENTS)
   @SseElementType(MediaType.APPLICATION_JSON)
   public Publisher<Status> streamIssues() {
        return status;
    }

   @GET
   @Path("/issues/collect")
   public void collectIssues() {
        fetchIssues().stream()
                     .map(i -> { emitter.send(new Status(i); return i})
                     .forEach(i.persist());
    }
```

Results in:

```
Caused by: java.lang.NullPointerException: Processor must not be null
at java.util.Objects.requireNonNull(Objects.java:228)
at org.eclipse.microprofile.reactive.streams.operators.core.Stages$ProcessorStage.<init>(Stages.java:175)
at org.eclipse.microprofile.reactive.streams.operators.core.PublisherBuilderImpl.via(PublisherBuilderImpl.java:186)
at io.smallrye.reactive.messaging.ProcessorMediator.connectToUpstream(ProcessorMediator.java:31)
at io.smallrye.reactive.messaging.extension.MediatorManager.lambda$null$5(MediatorManager.java:210)
at java.util.Optional.ifPresent(Optional.java:159)
at io.smallrye.reactive.messaging.extension.MediatorManager.lambda$weaving$6(MediatorManager.java:209)
at java.util.ArrayList.forEach(ArrayList.java:1257)
at io.smallrye.reactive.messaging.extension.MediatorManager.weaving(MediatorManager.java:204)
at io.smallrye.reactive.messaging.extension.MediatorManager.initializeAndRun(MediatorManager.java:188)
at io.smallrye.reactive.messaging.extension.MediatorManager_ClientProxy.initializeAndRun(MediatorManager_ClientProxy.zig:100)
at io.quarkus.smallrye.reactivemessaging.runtime.SmallRyeReactiveMessagingLifecycle.onApplicationStart(SmallRyeReactiveMessagingLifecycle.java:20)
... 14 more
```

cescoffier:
The previous algorithm led to chicken-and-egg issues as the traversal of the graph and the instantiation of the component was made in a single step. 

This new algorithm distinguishes the two phases.
First, the resolution phase looks at the metadata and components (emitter, channels, connectors, and mediators) and assembles the graph. Once built, the graph is validated to detect wiring issues (missing broadcast, missing upstream/downstream...). These issues are reported to the user before starting anything.  

After the validation, the materialization happens. It actually creates the managed objects and connects everything. 

In comparison to the first algorithm, this new one:

* improves error reporting - error messages have been clarified
* supports some topologies not handled before (having a channel, a processor, and an emitter in the same class)
* detect more issues before the application is even started. 

It fixes #414, #879, and https://github.com/quarkusio/quarkus/issues/12820

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
@@ -1,6 +1,7 @@
 package io.smallrye.reactive.messaging;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.microprofile.reactive.messaging.Emitter;
@@ -10,7 +11,12 @@ import org.eclipse.microprofile.reactive.streams.operators.SubscriberBuilder;
 
 public interface ChannelRegistry {
 
-    PublisherBuilder<? extends Message<?>> register(String name, PublisherBuilder<? extends Message<?>> stream);
+    default PublisherBuilder<? extends Message<?>> register(String name, PublisherBuilder<? extends Message<?>> stream) {
+        return register(name, stream, false);
+    }
+
+    PublisherBuilder<? extends Message<?>> register(String name, PublisherBuilder<? extends Message<?>> stream,
+            boolean broadcast);
 
     SubscriberBuilder<? extends Message<?>, Void> register(String name,
             SubscriberBuilder<? extends Message<?>, Void> subscriber);
@@ -32,4 +38,6 @@ public interface ChannelRegistry {
     Set<String> getOutgoingNames();
 
     Set<String> getEmitterNames();
+
+    Map<String, Boolean> getIncomingChannels();
 }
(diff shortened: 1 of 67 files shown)


kenfinnigan (commented):
Should all the exception classes you've added use defined exception message codes as we do for other exceptions?

cescoffier:
@kenfinnigan These exceptions used tailored messages. I don't believe our current approach provides that flexibility (but I may be wrong). It needs iterations on lists, and for some, actually, type checks. Having a class doing to work to create a "naked" exception that just has the error message does not bring much, because anyway, it won't be able to translate it.

cescoffier:
@kenfinnigan I just tried, and yes, it requires a _message_, and I can't pass a message as it's tailored in the exceptions themselves.

kenfinnigan:
No worries, thanks for verifying

cescoffier:
@kenfinnigan Did you review this PR?

kenfinnigan (approved):
Sorry, forgot to come back to it after the exception questions

PR #898, opened 2020-12-09
Put the MapBasedConfigs into a common module

kabir:
This will make it easier to move to MP Config 2.0 when that happens


cescoffier:
Thanks! I've merged it manually, as there were a _few_ conflicts.

PR #904, opened 2020-12-13
Fix missing getAck /  getNack implementation in Message implementations

### LINKED ISSUE #903: AMQP messages are not acknowledged when using a MessageConverter
https://github.com/languitar/amqp-ack-test-kotin contains a test quarkus kotlin project that exhibits missing ACKs for AMQP messages. In essence, the important code is:

```kotlin
class DummyData

@ApplicationScoped
class DummyDataConverter : MessageConverter {
    override fun canConvert(`in`: Message<*>, target: Type): Boolean = true

    override fun convert(`in`: Message<*>, target: Type): Message<*> {
        println("Converting $`in`")
        return `in`.withPayload(DummyData())
    }
}

@ApplicationScoped
@Path("/hello-resteasy")
class GreetingResource {

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    fun hello() = "Hello RESTEasy"

    @Incoming("test")
    fun process(data: DummyData) {
        println("Processing $data")
    }
}
```

Without the message converter incoming messages are acknowledged as expected. With this message converter, the ack is not sent, which I could also verify when tracing the communication with the broker (ActiveMQ Artemis) using wireshark. The `disposition` messages are not sent from the client to the broker once the message converter is installed. So, restarting the project simply reprocesses old messages. This doesn't happen when the converter is removed and the handler method is changed to

```kotlin
    @Incoming("test")
    fun process(data: ByteArray) {
        println("Processing $data")
    }
```

cescoffier:
Some implementations of Message didn't override the getAck and getNack method, breaking the acknowledgement chain.

- Fix #903
- Fix missing getAck / getNack override in Message implementations

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessage.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessage.java
@@ -2,9 +2,14 @@ package io.smallrye.reactive.messaging.amqp;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 import org.apache.qpid.proton.amqp.Binary;
-import org.apache.qpid.proton.amqp.messaging.*;
+import org.apache.qpid.proton.amqp.messaging.AmqpSequence;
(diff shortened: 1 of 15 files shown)


languitar:
To understand the implementation pattern behind this thing: why can't the `getAck` and `getNack` methods be abstract to avoid forgetting to implement them?

cescoffier:
Because that would create an infinite recursion in Message (coming from the spec)

kenfinnigan (approved):

PR #923, opened 2021-01-08
Improve injected channel support

cescoffier:
The injected channels will now merge multiple upstreams.
In the case of injecting a stream of payload, it broadcasts the payload to multiple subscribers. It also acknowledges the message.
In the case of a stream of messages, acknowledgment and broadcast are under the user's responsibility. 
﻿


cescoffier:
CC @gunnarmorling - we discussed that yesterday
CC @jponge - we also discussed that yesterday 

Basically, you don't need an intermediate "merging" method.
You don't need a method to handle the ack and broadcast anymore either.

jponge:
Great!

kenfinnigan (approved):

PR #929, opened 2021-01-12
Detect and throw a CDI DefinitionException when trying to inject an Emitter<Message<?>>

cescoffier:
- Detect the injection of `Emitter<Message<?>>`﻿
- Detect the injection of raw emitter (`Emitter`)
- Detect the injection of `Emitter<Message>>`
- Improve error reporting when the `@Channel` qualifier is missing


cescoffier:
CC @loicmathieu

loicmathieu (approved):
Maybe not an official review but this looks like a very good addition to detect early those development issue !

kenfinnigan (approved):

Issue #806, opened 2020-10-14
superflous AMQP message sections being sent


gemmellr:
Currently, the AMQP connector sends superflous empty sections for message areas it isnt really even using. For example sending a Message created from only a body payload, causes emission of an AMQP message including empty delivery-annotations, message-annotations, application-properties and footer sections, which for small payloads can rather significantly bloat the overall message. It does appear to omit including the unused/all-defaults header section though.

If a message section isnt being used it can just omit the section, saving the work needlessly done encoding and/or decoding at the sender, reciever, and intermediaries, plus the bandwidth transferring them.

gemmellr:
For later reference, this is an example of the payload previously being sent for an integer body (with the 'properties.to' address field populated):

>\x00Sq\xc1\x01\x00\x00Sr\xc1\x01\x00\x00Ss\xd0\x00\x00\x00,\x00\x00\x00\x03@@\xa1$65783e8e-d85f-4edf-9d8c-c3e1ff6228a1\x00St\xc1\x01\x00\x00Swq\x00\x00\x03\x8d\x00Sx\xc1\x01\x00

(format uses \xAB formatted hex for non printables)

gemmellr:
The encoded payload of a comparable message after #812 now being:
> \x00Ss\xd0\x00\x00\x00,\x00\x00\x00\x03@@\xa1$98bf06a4-e2bc-429e-817e-008b7efeb4ee\x00Swq\x00\x00\x01\x7c

PR #931, opened 2021-01-13
various AMQP test improvements

gemmellr:
Update various tests to use a test server rather than full broker, more closely validates the client behaviour of interest directly rather than through loose observation from the other side of an intermediary. Follows on from #800. Also adds a better test for #806. Knocks about 60sec off the time of the tests updated (plus maybe another ~20sec off a local build by default due to the test-containers profile addition; not precisely sure since it doesnt work in my local env, another reason to put it in a profile).


cescoffier (approved):

cescoffier:
Thanks!

PR #936, opened 2021-01-14
Set strict via MP Config rather than direct system property

kabir:


cescoffier (changes requested):
Documentation must be updated.

Are we sure this is backward compatible?

kabir:
> Documentation must be updated.

@cescoffier I added a commit

> Are we sure this is backward compatible?

Yes, there are tests that still use the system property and it ends up in the MP Config. Also, my WildFly WIP still uses the property

cescoffier (approved):

PR #943, opened 2021-01-18
Vertx4 mp4

kabir:
Would you mind merging this? If this doesn't become the 4.x branch, it can always be readjusted later. Having this and microprofile-mp4 share the same snapshot version causes some issues locally. Alternatively, I can just apply this commit in my local checkout.


cescoffier (approved):

cescoffier:
I hope it will be version 3, but let's see later.

PR #947, opened 2021-01-21
Fix Kafka producer close that can block forever

cescoffier:
This issue happens when the broker is not available anymore.


cescoffier:
\CC @mkouba that's the issue you mentioned to me yesterday.

PR #953, opened 2021-01-24
The AMQP incoming `durable` attribute should be false by default

cescoffier:


cescoffier:
@gemmellr as discussed.

gemmellr:
Looks good to me @cescoffier.

We should probably add a more specific test of the behaviour of the config change, I'll look to add that separately during the week (I think I recall one that at least somewhat tests this currently and could be expanded).

PR #963, opened 2021-01-29
expand test to cover change in default behaviour from #953

gemmellr:
Test for #953 as previously mentioned there.


gemmellr:
@cescoffier ^

cescoffier (approved):
Thanks!

PR #966, opened 2021-02-01
Make execution holder lazy

cescoffier:
@Azquelt Can you try with this one?


Azquelt:
Yeah, that works for me. I still need a few vertex classes included so that the classes can be loaded by CDI.

cescoffier:
Yes, but. No more runtime part.

Emily-Jiang:
Thank you @cescoffier @Azquelt ! This is a great news!

cescoffier:
Thanks for the feedback! Let's click on the "green" button.

cescoffier:
@kabir it should not change anything, so more a FYI.

PR #976, opened 2021-02-07
Kafka - Pause / Resume underlying consumer when the downstream does not request items

### LINKED ISSUE #974: Pause the Kafka consumer when request reaches 0
And resume it when new requests are made. 

* When the amount of request reaches 0, create a timer tracking the requests. If no request is made during a given time, pause the consumer, avoiding the poll thread to break
* If during the grace period, requests are made, cancel the timer
* If the consumer is paused and new requests arrive, resume the consumer (and make sure the amount of request is passed).

cescoffier:
Pause / Resume the underlying Kafka consumer following the downstream request rate.
When there are no requests for a period of time the consumer is paused. The consumer is resumed when the next request is received.

This PR also offers a facade service to retrieve the managed Kafka consumer and producer.

Fix #974:

- Provide a facade service to retrieve the managed Kafka consumer and producers
- Implement pause/resume cycle following the downstream request rate


cescoffier:
@loicmathieu FYI.

cescoffier:
This PR does not work as expected.
The problem is that the underlying Vert.x Kafka client stops polling which is not correct. Even paused or if you have no requests, you MUST continue polling regularly.

loicmathieu:
@cescoffier maybe the part of the PR to allow to access the underlying producer/consumer can be extracted and merged as it serves a different purpose ?

cescoffier:
There is another (draft) PR with a new client (fixing the pause/resume) and providing the access to the underlying client.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -205,10 +226,21 @@ public KafkaSource(Vertx vertx,
...
-                .map(rec -> commitHandler
-                        .received(
-                                new IncomingKafkaRecord<>(rec, commitHandler, failureHandler, isCloudEventEnabled,
-                                        isTracingEnabled)));
+                .onRequest().call(l -> {
+                    lastRequest.set(System.currentTimeMillis());
+                    if (paused.compareAndSet(true, false)) {
+                        log.resumingChannel(channel);
+                        return consumer.paused()
+                                .flatMap(consumer::resume);
```
pcasaes, line 234: Not sure if this is related to the issues you saw with this PR.

Back when we first tried to allow for a user provided consumer re-balance listener we used **pause** and **resume** to get around the Vert.x client's non blocking interface. Using **resume** broke the consumer because it switched it from fetch mode to flow mode: 
> io.smallrye.mutiny.subscription.BackPressureFailure: Buffer full, cannot emit item

See Streaming mode:
https://vertx.io/docs/apidocs/io/vertx/core/streams/ReadStream.html

references:
https://github.com/vert-x3/vertx-kafka-client/pull/180
https://github.com/smallrye/smallrye-reactive-messaging/pull/672

cescoffier: No, it's way worse than this. When paused, it stops polling, to after the heartbeat period, so the broker considers you "dead". 

In addition, the pause/resume methods from the client are unrelated to the pause/resume methods from Kafka. 

I have made progress on a new client in this PR: #995

Issue #1002, opened 2021-02-16
Support MP Config 2.0 for MP Reactive Messaging 1.0


Azquelt:
I'd like to make Reactive Messaging 1.0 work with MP Config 2.0. I think `ConnectorConfig` would be the only class which would need updating to be compatible since it implements the `Config` interface.

Ideally I'd have liked to make a new version which works with both MP Config 2.0 and 1.x, but I don't think that's easily done since 2.0 adds methods to the `Config` interface which reference classes which didn't exist in 1.x.

cescoffier:
One of the issues I faced in the implementation of these new methods. Not necessarily easy or happy with what I have in SR RM 3.x.

Azquelt:
There were a few bits I wondered about. If I end up making changes to `ConnectorConfig`, I'll be sure to open a PR for those changes in the 3.x branch as well.

Azquelt:
Upon further investigation, I don't think this is actually possible.

The `1.1.0` release has changes which make it incompatible with the Reactive Messaging 1.0 API (in particular, `Message` implementations need to override the default `getAck()` for acknowledgement to work correctly.)

We can't branch off from the `1.0.8` release either because we can't support both MP Config 1.x and 2.0 with one implementation of `ConnectorConfig`. (The 2.0 interface references `ConfigValue` which is a new interface not present in 1.x. A valid 2.0 implementation of `ConnectorConfig` therefore can't be loaded or compiled against the 1.x API). We couldn't put out a `1.0.9` release which required users to update to MP Config 2.0.

PR #1017, opened 2021-02-18
Improve ConnectorConfig

Azquelt:
* Do correct conversion for the channel and connector name properties
* Avoid unnecessarily throwing and catching exceptions when we need to
  check several different names for one lookup
* Corrections to the way we create the list of property names, taking
  into account the rules for matching environment variable names
* Create a proper ConfigValue object when looking up the channel or
  connector name as a ConfigValue.
* Don't return the overall config from .unwrap(). If a connector needs
  to access the overall config, they can inject it.

@cescoffier You mentioned in #1002 that you weren't entirely happy with the `ConnectorConfig`. I had a look through and here are my suggested changes.

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConnectorConfig.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConnectorConfig.java
@@ -5,6 +5,7 @@ import static io.smallrye.reactive.messaging.i18n.ProviderMessages.msg;
 import static org.eclipse.microprofile.reactive.messaging.spi.ConnectorFactory.*;
 
 import java.util.*;
+import java.util.regex.Pattern;
 
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigValue;
@@ -64,38 +65,37 @@ public class ConnectorConfig implements Config {
         return CONNECTOR_PREFIX + connector + "." + keyName;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public <T> T getValue(String propertyName, Class<T> propertyType) {
         if (CHANNEL_NAME_ATTRIBUTE.equalsIgnoreCase(propertyName)) {
-            return (T) name;
+            return convert(name, propertyType);
         }
         if (CONNECTOR_ATTRIBUTE.equalsIgnoreCase(propertyName) || "type".equalsIgnoreCase(propertyName)) {
-            return (T) connector;
+            return convert(connector, propertyType);
         }
 
         // First check if the channel configuration contains the desired attribute.
-        try {
-            return overall.getValue(channelKey(propertyName), propertyType);
-        } catch (NoSuchElementException e) {
-            // If not, check the connector configuration
-            try {
-                return overall.getValue(connectorKey(propertyName), propertyType);
-            } catch (NoSuchElementException e2) {
-                // Catch the exception to provide a more meaningful error messages.
-                throw ex.noSuchElementForAttribute(propertyName, name, channelKey(propertyName),
-                        connectorKey(propertyName));
-            }
+        Optional<T> maybeResult = overall.getOptionalValue(channelKey(propertyName), propertyType);
+        if (maybeResult.isPresent()) {
+            return maybeResult.get();
         }
+
+        // Then check if the connector configuration contains the desired attribute.
+        maybeResult = overall.getOptionalValue(connectorKey(propertyName), propertyType);
+        if (maybeResult.isPresent()) {
+            return maybeResult.get();
+        }
+
+        throw ex.noSuchElementForAttribute(propertyName, name, channelKey(propertyName), connectorKey(propertyName));
     }
 
     @Override
     public ConfigValue getConfigValue(String propertyName) {
         if (CHANNEL_NAME_ATTRIBUTE.equalsIgnoreCase(propertyName)) {
-            return overall.getConfigValue(channelKey(CHANNEL_NAME_ATTRIBUTE));
+            return new ConfigValueImpl(CHANNEL_NAME_ATTRIBUTE, name);
         }
         if (CONNECTOR_ATTRIBUTE.equalsIgnoreCase(propertyName) || "type".equalsIgnoreCase(propertyName)) {
-            return overall.getConfigValue(channelKey(CONNECTOR_ATTRIBUTE));
+            return new ConfigValueImpl(CONNECTOR_ATTRIBUTE, connector);
         }
         // First check if the channel configuration contains the desired attribute.
         ConfigValue value = overall.getConfigValue(channelKey(propertyName));
@@ -106,20 +106,48 @@ public class ConnectorConfig implements Config {
         return value;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public <T> Optional<T> getOptionalValue(String propertyName, Class<T> propertyType) {
         if (CHANNEL_NAME_ATTRIBUTE.equalsIgnoreCase(propertyName)) {
-            return Optional.of((T) name);
+            return convertOptional(name, propertyType);
         }
         if (CONNECTOR_ATTRIBUTE.equalsIgnoreCase(propertyName) || "type".equalsIgnoreCase(propertyName)) {
-            return Optional.of((T) connector);
+            return convertOptional(connector, propertyType);
         }
 
         // First check if the channel configuration contains the desired attribute.
         Optional<T> maybe = overall.getOptionalValue(channelKey(propertyName), propertyType);
         return maybe.isPresent() ? maybe
                 : overall.getOptionalValue(connectorKey(propertyName), propertyType);
+    }
+
+    private <T> T convert(String rawValue, Class<T> propertyType) {
+        Optional<Converter<T>> converter = overall.getConverter(propertyType);
+
+        if (!converter.isPresent()) {
+            if (propertyType.isAssignableFrom(String.class)) {
+                return propertyType.cast(rawValue);
+            }
+            throw ex.noConverterForType(propertyType);
+        }
+
+        T result = converter.get().convert(rawValue);
+
+        if (result == null) {
+            throw ex.converterReturnedNull(converter.get(), rawValue);
+        }
+
+        return result;
+    }
+
+    private <T> Optional<T> convertOptional(String rawValue, Class<T> propertyType) {
+        Optional<Converter<T>> converter = overall.getConverter(propertyType);
+
+        if (!converter.isPresent() && propertyType.isAssignableFrom(String.class)) {
+            return Optional.of(propertyType.cast(rawValue));
+        }
+
+        return converter.map(c -> c.convert(rawValue));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/impl/ConnectorConfig.java
```diff
@@ -131,47 +159,56 @@ public ConfigValue getConfigValue(String propertyName) {
...
+                String computed = name.substring(prefixAlphaUpper.length());
+                if (nameExists(prefix + computed)) {
                     names.add(computed);
                 }
-            } else if (name.startsWith(prefixFromEnv)) {
-                String computed = name.substring(prefixFromEnv.length());
-                // Remove the potential existing key
-                names.removeIf(s -> toEnv(s).equalsIgnoreCase(computed));
-                names.add(computed);
             }
```
Azquelt, line 197: I don't really like this code but I couldn't find a better way to do it.

The problem is that config properties derived from environment variables should be returned if they match the confit name exactly, match the config name with non-alphanumeric characters converted to `_` or match the config name with non-alphanumber caracters converted to `_` and the rest converted to upper case.

However, for config properties from other sources, they usually have to match exactly.

The only way I could find to do this was to first try all the combinations which are allowed by the environment variable config source, then check if the final name we've created is actually present in the config. If it was from an environment variable we should find it, but if it wasn't then we won't.

Azquelt: I also removed the behaviour where you tried to avoid adding a name twice if it only differred by case, since the smallrye config implementation would return both names.

(e.g. if I have `foo` in a microprofile-config.properties file, and `FOO` in an environement variable, `getPropertyNames()` will return both `foo` and `FOO`, even though I'll get the same value when I look either of them up.)

cescoffier: Unfortunately, if we did that it's because we got a bug where the "foo" attribute was unable to be overridden by "FOO".
If I remember correctly it was:

`kafka.bootstrap.servers` vs. `KAFKA_BOOTSTRAP_SERVERS`. 

So we must be sure that this works (as well as when `kafka.bootstrap.servers` is not set and uses the default)

cescoffier (changes requested):
It looks good. I've made a comment where we need to be sure we don't introduce a regression when properties are overridden. 

The problem occurs in the Kafka connector that, to satisfy the Kafka client configuration, needs to do some extract handling on the properties.

Azquelt:
> Unfortunately, if we did that it's because we got a bug where the "foo" attribute was unable to be overridden by "FOO".
> If I remember correctly it was:
> 
> `kafka.bootstrap.servers` vs. `KAFKA_BOOTSTRAP_SERVERS`.
> 
> So we must be sure that this works (as well as when kafka.bootstrap.servers is not set and uses the default)

Hmm, that should be covered by the `at-tr` case in the unit test.

Let me outline a couple of general cases and then I can add them to the unit test if it would let us feel more confident about the changes:

First case (which I think is the one you're talking about):
* TestConfigSource [ordinal 100] contains `mp.messaging.incoming.mychannel.foo=bar`
* Environment variable source [ordinal 300] contains `MP_MESSAGING_INCOMING_MYCHANNEL_FOO=baz`
* `config.getValue("foo", String.class)` returns `"baz"`
* `config.getValue("FOO", String.class)` returns `"baz"`
* `config.getPropertyNames()` returns `["foo", "FOO"]`

Second case:
* TestConfigSource [ordinal 100] contains `mp.messaging.incoming.mychannel.foo=bar`
* Environment variable source [ordinal 300] contains `MP_MESSAGING_INCOMING_MYCHANNEL_FOO=baz`
* TestConfigSource2 [ordinal 500] contains `mp.messaging.incoming.mychannel.foo=qux`
* TestConfigSource3 [ordinal 700] contains `MP_MESSAGING_INCOMING_MYCHANNEL_FOO=quuz`
* `config.getValue("foo", String.class)` returns `"qux"`
  * Not `"quuz"` because the uppercase matching is only done for environment variables, so the value from `TestConfigSource3` is not considered
* `config.getValue("FOO", String.class)` returns `"baz"`
  * Not `"quuz"` because we'll search the overall config for `"mp.messaging.incoming.mychannel.FOO"` and uppercase matching is only done for environment variables
  * Not `"qux"` because
    1. Uppercase matching is only done for environement variables
    1. Even when we do uppercase matching, we only uppercase the thing we're searching for, not the value from the config source. Searching `foo` may find a key named `FOO`, but searching `FOO` will never find `foo`.
* `config.getPropertyNames()` returns `["foo", "FOO"]`

There is another case which I hit in our kafka connector:
* Environment variable source [ordinal 300] contains `MP_MESSAGING_INCOMING_MYCHANNEL_FOO=baz`
* `config.getValue("foo", String.class)` returns `"baz"`
* `config.getValue("FOO", String.class)` returns `"baz"`
* `config.getPropertyNames()` returns `["FOO"]`
* If our connector looks explicitly for `"foo"` then it finds it. However, if it iterates through the result of `getPropertyNames()`, it doesn't see `"foo"` there at all.
  * This means that if the connector has a list of properties to look for to pass through to the Kafka client then it works
  * However, if it tries to gather up all the property names it doesn't recognise and pass them to the Kafka client, it ends up passing a property named `"FOO"` instead of `"foo"`
* I can see two possible workarounds for this:
  * The connector has to have a list of property names it looks for and can't rely on `getPropertyNames()`
  * The user can define the desired property with the correct casing in `microprofile-config.properties` and then override it in an environment variable. (This is the same as the First case above.)

cescoffier:
Unfortunately, in the Kafka world, you don't know the set of keys, as it depends on the various component used (serde, assignator...). 

That being said, the code you removed can be restored in the Kafka connector to eliminate duplicated keys.

Azquelt:
I had a look and I couldn't actually see where the Kafka connector is using the set of property names. I did find this example in the docs which would need updating slightly: https://github.com/smallrye/smallrye-reactive-messaging/blob/master/documentation/src/main/doc/modules/kafka/pages/default-configuration.adoc

cescoffier:
It's this class that does not gymnastics with the property names: https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/JsonHelper.java

Azquelt:
I'm happy to add some de-duplication to `JsonHelper`, but I'm a bit confused at what the old code was trying to accomplish.

As far as I can work out, the old code said "If you have an uppercase version and a lowercase version of the same property, prefer the uppercase version". This is supported by the [old test looking for](https://github.com/smallrye/smallrye-reactive-messaging/pull/1017/files#diff-9826df505660d404d63487cb621c216c80b55e45ad50f25a3f93a944fc634565L70-L71) `AT_TR` but not `at-tr`.

However, this would then mean that environment variables would override everything else (including from sources with a higher ordinal, like system properties) because we would then look up `AT_TR` which would only find the property named `AT_TR` in the environment variables and would not find a property named `at-tr` in another config source.

I think the correct behaviour would be to discard the uppercase version if we have an uppercase and lowercase version, but it would probably be a good idea to add a test first to check if there is a bug and assert the correct behaviour.

The way that `JsonHelper` then does _further_ processing to convert to lowercase and convert underscores to dots also gives another opportunity for name collisions so we should probably warn if collisions do occur.

cescoffier:
I agree that the uppercase/lower case is not handled correctly. However, MP Config does not really provide the right API for this (at least last time I checked). Accessing all config sources and doing the work on our side sounds a bit convoluted, but I guess, we don't have much choice. 

The problem with Kafka is that the set of attributes is unknown. So, we need to iterates over the property names and get the right value. Let's imagine the underlying client needs the property `foo.bar` (Kafka properties use dots as separator):

- Case 1 - neither `foo.bar` not `FOO_BAR` is set - ok, easy
- Case 2 - `foo.bar` is set but not `FOO_BAR` - easy again
- Case 3 - `FOO_BAR` is set but not `foo.bar` - easy again
- Case 4 - both are set. In this case, it may be a collision, but not necessarily. It may be an overridden property (most of the time it is). So, in this case, we should pick the one coming from the highest config source (highest priority). In the current code, it considers that ENV is the highest, as it's generally the "last resort to override values". If we can improve this and check which config sources (plural) provide the value and gets the right one it would be great.

Azquelt:
> If we can improve this and check which config sources (plural) provide the value and gets the right one it would be great.

With MP Config 2.0, we can actually do that using `getPropertyValue()` which tells you what config source and ordinal were used to get the value.

However, there's another problem:
* `mp.messaging.incoming.my-channel.foo.bar` is set in a low ordinal config source
* `mp.messaging.connector.kafka.FOO_BAR` is set in a high ordinal config source
* The first value should be used, even though it's from a lower ordinal config source, because it's specific to the channel

At this point, the `Config` object provided by reactive messaging is hindering rather than helping us since it's hiding whether a given property was defined for the connector or for the channel.

I think we have two options:
1. Require that configuration properties for the kafka client must be set in lower case (unless they're set as environment variables).
    - This would allow us to take the list of property names, lowercase them and replace `_` with `.` and then look them up through the `Config` and take the result it gives us.
    - If the user set `FOO_BAR` in the system properties, it would be ignored since we'd look up `foo.bar` and wouldn't find anything.
1. Inject the overall `Config` into the Kafka connector and implement separate name-processing logic to allow wrongly-cased config properties from any config source.
    1. This would allow us to create two lists of properties, those for the channel and those for the connector
    2. We could then look up each name and get the ordinal of the source that provided it
    3. We could then lowercase every name and where there are collisions use the ordinal to decide which one to use
        - You could still have a tie at this point since `foo.bar` and `FOO_BAR` could be defined in the same config source.

I think I'd argue that the first option is ok, there's no reason why someone can't write `mp.messaging.incoming.my-channel.bootstrap.properties` rather than `mp.messaging.incoming.mychannel.Bootstrap.Properties` in most config sources and it could still be declared as `MP_MESSAGING_INCOMING_MYCHANNEL_BOOTSTRAP_PROPERTIES` in an environment variable.

My only concern is that existing users could have put `mp.messaging.incoming.mychannel.Bootstrap.Properties` in their system properties and that would stop working after this change.

cescoffier:
Thanks for this detailed comment. I like 2. As it would be only done for 3.x (MP Config 2), breaking is ok (even if we don't want to break everything:-)).

Azquelt:
Ok, I'll make the following test cases for option 2:
* Case 1
  * Env: `MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = a`
  * Sys props: `mp.messaging.incoming.mychannel.foo.bar = b`
  * Expect `foo.bar = b`
    * Sys props is higher ordinal
* Case 2
  * Env: MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = "a"
  * Sys props: mp.messaging.incoming.mychannel.FOO_BAR = "b"
  * Expect `foo.bar = b`
    * Matching on the name is case insensitive
* Case 3
   * Env: `MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = a`
   * microprofile-config.properties: `mp.messaging.incoming.mychannel.foo.bar = b`
   * Expect `foo.bar = a`
     * Env is a higher ordinal
* Case 4
   * Env: `MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = a`
   * Sys props: `MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = b`
   * Expect `foo.bar = a`
     * Prefix matching is still case sensitive, so the property from sys props is ignored
* Case 5
  * Env: `MP_MESSAGING_CONNECTOR_KAFKA_FOO_BAR = a`
  * Sys props: `mp.messaging.connector.kafka.foo.bar = b`
  * Expect `foo.bar = b`
    * Same as case 1, but for connector properties rather than channel properties
* Case 6
  * Env: `MP_MESSAGING_INCOMING_MYCHANNEL_FOO_BAR = a`
  * Sys props: `mp.messaging.connector.kafka.foo.bar = b`
  * Expect `foo.bar = a`
    * Channel properties override connector properties, regardless of ordinal

Note on case 4: I wanted to ensure that the Kafka connector has the same requirements as other connectors. It would be confusing if the prefix was case sensitive on other connectors but case insensitive for the Kafka connector.

cescoffier:
* Case 8
   * Env property: MP_MESSAGING_CONNECTOR_KAFKA_BOOTSTRAP_SERVERS=....
   * Expect `bootstrap.servers=...`
   
* Case 9
   * Env property: MP_MESSAGING_INCOMING_KAFKA_BOOTSTRAP_SERVERS=....
   * Expect `bootstrap.servers=...`

Azquelt:
Hmm, I noticed that you have some documented properties with dashes rather than dots.

Should we allow `_` instead of `-` as well?

* Case 10
  * Env property: `MP_MESSAGING_CONNECTOR_KAFKA_RETRY_ATTEMPTS = 5`
  * Expected `getRetryAttempts() = 5`
* Case 11
  * Sys property: `mp.messaging.connector.kafka.RETRY_ATTEMPTS = 5`
  * Expected `getRetryAttempts() = 5 or -1?`
    * Note: the documented name is `retry-attempts`
    * Should we take a value from `RETRY_ATTEMPTS`?
    * What about `RETRY-ATTEMPTS`? (see also case 7 above)
    * What about `RETRY.ATTEMPTS`? (should we treat all punctuation as equivalent?)

Azquelt:
> Should we allow `_` instead of `-` as well?

The more I think about this, the more I dislike it. We're moving further and further away from the standard MP Config behaviour and relying more and more on processing the entire set of config properties and iterating over and filtering them ourselves.

I definitely think it would be better if we were able to look up the config properties known in advance with `config.getOptionalValue()`.

Where we have to iterate over the set of names to create the Kafka config, the behaviour should be consistent with knowing the list of names in advance and looking them up with `config.getOptionalValue()`. I.e. if we see a property named `FOO_BAR`, we can intuit that there could be a property named `foo.bar`, but we would then actually call `config.getOptionalValue("foo.bar")` and take the result. (This is pretty close to Option 1 above.)

Still, I guess it depends how much of a burden the change would be for users and whether avoiding that is worth the inconsistency and additional maintenance effort for this code.

cescoffier:
Configuration properties using `-` are defined by the connector and known. The problem is the ones consumed by Kafka which are unknown (because a serializer can expect any attribute).

Azquelt:
Right, my question was around whether we should do any additional processing on those known properties.

If we do no further processing, then the lookup of those properties will be case sensitive (unless they're environment variables).

This contrasts with test case 7 above (`topic` is a known configuration property) and creates an inconsistency between how "known" and "unknown" properties are looked up since known properties would be case sensitive and unknown properties would not.

I would suggest that all properties (known and unknown) should be case sensitive (unless they're defined using environment variables). This would change the result of Case 2 and Case 7.

Azquelt:
@cescoffier I've made the change to the way kafka properties are handled and added the tests. The actual code change ended up being pretty small, but it should now all be consistent between the known and unknown properties.

I ended up using custom config sources where I can directly control the ordinal rather than system properties as I ran into junit-pioneer/junit-pioneer#432

cescoffier (approved):
Thanks! That's awesome work.

PR #1057, opened 2021-02-24
Fix#1053: Use passedConfig over defaultConfig in KafkaConnector

flansen:
This fixes the issue that the default config took precedence over the more specific configuration


flansen:
I'm not sure why the Java11 pipeline would fail, is it flaky?

cescoffier (approved):

cescoffier:
Thanks!

PR #1075, opened 2021-03-01
AMQP - Expose virtual host attribute

cescoffier:
Related to https://github.com/quarkusio/quarkus/issues/15229

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -52,7 +52,7 @@ import io.vertx.mutiny.core.Vertx;
 @ConnectorAttribute(name = "host", direction = INCOMING_AND_OUTGOING, description = "The broker hostname", type = "string", alias = "amqp-host", defaultValue = "localhost")
 @ConnectorAttribute(name = "port", direction = INCOMING_AND_OUTGOING, description = "The broker port", type = "int", alias = "amqp-port", defaultValue = "5672")
 @ConnectorAttribute(name = "use-ssl", direction = INCOMING_AND_OUTGOING, description = "Whether the AMQP connection uses SSL/TLS", type = "boolean", alias = "amqp-use-ssl", defaultValue = "false")
-@ConnectorAttribute(name = "virtual-host", direction = INCOMING_AND_OUTGOING, description = "If set, configure the virtual host name for the connection AMQP Open frame and TLS SNI server name (if TLS is in use)", type = "string", alias = "amqp-virtual-host")
+@ConnectorAttribute(name = "virtual-host", direction = INCOMING_AND_OUTGOING, description = "If set, configure the hostname value used for the connection AMQP Open frame and TLS SNI server name (if TLS is in use)", type = "string", alias = "amqp-virtual-host")
 @ConnectorAttribute(name = "sni-server-name", direction = INCOMING_AND_OUTGOING, description = "If set, explicitly override the hostname to use for the TLS SNI server name", type = "string", alias = "amqp-sni-server-name")
 @ConnectorAttribute(name = "reconnect-attempts", direction = INCOMING_AND_OUTGOING, description = "The number of reconnection attempts", type = "int", alias = "amqp-reconnect-attempts", defaultValue = "100")
 @ConnectorAttribute(name = "reconnect-interval", direction = INCOMING_AND_OUTGOING, description = "The interval in second between two reconnection attempts", type = "int", alias = "amqp-reconnect-interval", defaultValue = "10")
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -52,6 +52,8 @@ import io.vertx.mutiny.core.Vertx;
 @ConnectorAttribute(name = "host", direction = INCOMING_AND_OUTGOING, description = "The broker hostname", type = "string", alias = "amqp-host", defaultValue = "localhost")
 @ConnectorAttribute(name = "port", direction = INCOMING_AND_OUTGOING, description = "The broker port", type = "int", alias = "amqp-port", defaultValue = "5672")
 @ConnectorAttribute(name = "use-ssl", direction = INCOMING_AND_OUTGOING, description = "Whether the AMQP connection uses SSL/TLS", type = "boolean", alias = "amqp-use-ssl", defaultValue = "false")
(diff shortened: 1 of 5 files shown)


cescoffier:
FYI  @gemmellr

kenfinnigan (approved):

gemmellr:
As I said on quarkus/issues/15229, vertx-protons virtualhost client option explicitly sets *both* the TLS SNI value and the AMQP Open frame hostname values to the value provided, overriding the DNS hostname connected to. This PR only exposes usage of the vertx-amqp-client(/vertx-proton) virtualhost option, but seemingly not the related SniServerName option. Doesnt look like it is exposed otherwise? That may be important if in the case being asked about they may not want the SNI to match their broker-internal virtualhost, but some other value, and if so might need to also explicitly configure that too (e.g back to the DNS hostname connected to, or some other common name the server cert uses). 

Usage of vertx-protons virtualhost ClientOption was expected to be rare. For the Open frame the main interaction if wanting something other than the DNS name connected to is use of the setHostname method on the connection object before opening it. As vertx-amqp-client opens the connection itself it obviously cant expose a similar setter, it would have to either manipulate the options given to vetx-proton and introduce its own behaviour, or it might be worth vertx-amqp-client having its own e.g 'amqp-hostname' option to allow it to set the Open hostname using the vertx-proton connection.setHostname(..) method before it then opens it.

Actually, looking at the vertx-amqp-client code, it seems to already both pass the virtualhost option straight through to vertx-proton *and* then use the same option value to explicitly set the hostname on the connection object before opening it, which doesnt really make sense as it should already be done by then. That could be changed to make use of such a new option.

cescoffier:
@gemmellr so we need a new attribute to configure the hostname? 
As you mentioned, the `virtual-host` option is passed as hostname, so if that's not totally correct, what would need to be set?

gemmellr:
Im saying that vertx-amqp-client using the virtualhost option value to call setHostname on the vertx-proton connection should be redundant. Doing that is one of the two things that vertx-proton does with the virtualhost option, and so it should already be done by then. The value is fine, it should just already be set

I'm also saying that reactive-messaging either needs to also expose the SniServerName option, so it can be used to either the SNI value, perhaps to undo what the virtualhost option does to the SNI value. That, or a further option be added to allow solely controlling the AMQP connection hostname (e.g amqp-connection-hostname?),  as that cant be done in isolation with vertx-amqp-client currently (as it lacks the setHostname ability and it cant be added), but can be done with vertx-proton (as it has the setHostname ability).

cescoffier:
Ok, so for the SNI part we would need some fix in the various upstreams.

gemmellr:
Mm, no?

cescoffier:
Oh, I just found the SNI options! So, yes, we could expose it too as `amqp-connection-hostname`. 

So, in the end, we would have 2 new attributes:

* amqp-connection-hostname
* amqp-virtual-host

Both passed, if set, to the AMQP Client Options.

gemmellr:
The `amqp-connection-hostname` was a suggestion for a new option that only controlls the Open hostname (which would indeed need a vertx-amqp-client change to be made). There is already an SniServerName option as youve found for controlling the SNI, just not exposed by reactive-messaging.

The virtualhost option this PR exposes, controls both the Open hostname and the SNI hostname, unless the SNI is already overriden with its option which takes precedence. Using both may be needed for the requesting user (to set SNI back to the DNS value), or the new option above could be added instead.

cescoffier:
Yes, `amqp-connection-hostname` should be `amqp-sni-server-name`

gemmellr:
I dont really follow that. The SNI stuff already has its own explicit option, and is also implicit via the existing virtualhost option. Why wouldnt the explicit SNI option just the existing explicit SNI option?

The `amqp-connection-hostname` suggestion was to add another separate option to vertx-amqp-client and here, that specifically controls only the Open frame hostname value, independent of the SNI. Thats something you can already do with vertx-proton but not with vertx-amqp-client, because it opens the underlying connection before returning it. The new option could be used to set the hostname value [here](https://github.com/vert-x3/vertx-amqp-client/blob/3.9.6/src/main/java/io/vertx/amqp/impl/AmqpConnectionImpl.java#L85-L87), rather than the redundant re-use of the virtualhost option that will already be in effect.

cescoffier:
I don't believe SNI has its own exposed option, at least it's not exposed in the connector.

Here is the configuration code:

```
AmqpClientOptions options = new AmqpClientOptions()
                    .setUsername(username)
                    .setPassword(password)
                    .setHost(host)
                    .setPort(port)
                    .setContainerId(containerId)
                    .setSsl(useSsl)
                    .setReconnectAttempts(reconnectAttempts)
                    .setReconnectInterval(reconnectInterval)
                    .setConnectTimeout(connectTimeout);
```

cescoffier:
I agree with changing the vert.x AMQP client, but it won't be made before the next release.

gemmellr:
It isnt exposed here yet, that was what my original comment on this PR was, that its missing and the user prompting this may need it. It is definitely there in vertx-amqp-client since its there in the options in vertx-proton: https://github.com/vert-x3/vertx-amqp-client/blob/3.9.6/src/main/java/io/vertx/amqp/AmqpClientOptions.java#L499-L502

cescoffier:
Ok, I updated the PR.

It exposes the two options we discussed:

* virtual host
* sni server name

When set, they are passed to the Vert.x AMQP Client Options.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -52,6 +52,8 @@
 @ConnectorAttribute(name = "host", direction = INCOMING_AND_OUTGOING, description = "The broker hostname", type = "string", alias = "amqp-host", defaultValue = "localhost")
 @ConnectorAttribute(name = "port", direction = INCOMING_AND_OUTGOING, description = "The broker port", type = "int", alias = "amqp-port", defaultValue = "5672")
 @ConnectorAttribute(name = "use-ssl", direction = INCOMING_AND_OUTGOING, description = "Whether the AMQP connection uses SSL/TLS", type = "boolean", alias = "amqp-use-ssl", defaultValue = "false")
+@ConnectorAttribute(name = "virtual-host", direction = INCOMING_AND_OUTGOING, description = "If set, configure the virtual host name for the connection AMQP Open frame and TLS SNI server name (if TLS is in use)", type = "string", alias = "amqp-virtual-host")
```
gemmellr, line 55:
```suggestion
@ConnectorAttribute(name = "virtual-host", direction = INCOMING_AND_OUTGOING, description = "If set, configure the hostname value used for the connection AMQP Open frame and TLS SNI server name (if TLS is in use)", type = "string", alias = "amqp-virtual-host")
```

gemmellr (commented):
Looks good, one small tweak suggestion for the descriptions (would need a rebuild for the others) to better indicate what its doing.

A test would be good to ensure it is at least affecting the Open hostname value, I'm putting one together.

gemmellr:
I created a test as referenced just above, you can either pick the commit in or I can do a PR after yours lands.

I also raised a PR for vertx-amqp-client to add the new option that could be used in a future update.

cescoffier:
Thanks @gemmellr 

I've applied your suggestion and cherry-picked the test

PR #1081, opened 2021-03-07
HiveMQ version proposal

masini:
Hi Clement, as discussed this is a new version with both client inside a single module, two connectors and tests that are able to run on both kind of Source and Sink.


cescoffier:
Can you squash your commits?

masini:
Hi Clement, sorry for asking but how can I squash them now that the PR is ongoing ? 

Never did that before and I don't want to create a mess on the git log.

Thanks

masini:
Hi Clement, check if that works for you.

cescoffier:
I still see 2 commits with similar messages. But I should be able to squash during the merge.

One other question: does the connector work in native?

cescoffier:
Hum, that's a bit problematic, because basically, we need an extension in addition to the connector. 
In addition, it uses a shaded version of Netty, which means we would need to duplicate the substitutions we have in Quarkus and adapt them to use the shaded package name.

cescoffier:
> In the meanwhile can we mark the connector "not ready" for native build ?

not really, because it would mean we need to have an extension somewhere (quarkiverse probably). 

All the maintained connectors from SR Reactive Messaging works in native. They have been designed for that. Some do need "extensions" (like Kafka), some do not (like AMQP, even if we did a very simple extension).

masini:
No problem Clement, I can start cloning the Kafka Extension and work that way to compile in native mode.

Also if you think is a good idea I can switch back to the standard dependency of HiveMQ.

Are you the mantainer of the extension ?

cescoffier:
For Hive, yes. In native mode, you don't want anything shaded (or we would need to duplicate substitutions).
For the Kafka extension, yes, I'm the maintainer, but I don't it will be the same, as each extension is tied to the _supported_ library.

masini:
Hi Clement, adding this method:
```java
    @BuildStep
    public void build(BuildProducer<RuntimeInitializedPackageBuildItem> initializedPackageBuildItemBuildProducer) {
        initializedPackageBuildItemBuildProducer.produce(new RuntimeInitializedPackageBuildItem("com.hivemq"));
    }
```
to the SmallRyeReactiveMessagingMqttProcessor I was able to remove this from the pom.xml:
```xml
        <quarkus.native.additional-build-args>
          --initialize-at-run-time=com.hivemq
        </quarkus.native.additional-build-args>
```
Now I only need to know how to configure Arc to exclude that jar, something that in the application I did with the two properties:
```java
quarkus.arc.exclude-dependency.hivemq.group-id=com.hivemq
quarkus.arc.exclude-dependency.hivemq.artifact-id=hivemq-mqtt-client

```
Any hints on how to add this to the extention ? I can't find anything on the net, I also asked on quarkus-dev.

Thanks

cescoffier:
is there anything we can do (Substitutions)  to avoid the runtime initialization? It has important costs.

masini:
I can put class-by-class instead of the entire package.

I cannot avoid all of them because some of them does socket connection to the broker, something we cannot avoid.

Any idea also on how can I avoid the scanning of the jar containing beans.xml by ARC ?

cescoffier:
It's not a question of packages or classes, we try to avoid as much as possible runtime initialization as it has a direct impact on the startup time and may have an impact on the memory too. To avoid that, we use substitutions, basically, we modify the code of the class to avoid requiring runtime initialization. We trim the need for runtime initialization to the strict minimum. 

About Arc, I don't know, @mkouba may know.
Another approach would be to veto the defined beans.

mkouba:
> About Arc, I don't know, @mkouba may know.

https://github.com/quarkusio/quarkus/issues/15713

masini:
@mkouba I commented on quarkusio/quarkus#15713, I did what you said and it works perfectly.

@cescoffier I would like to avoid as much runtime-init as I can, but from what I saw and from my incomplete comprehension of HiveMQ, it creates socket connection inside the constructor so during image generation I don't know how to modify the classes not to do that. I don't want to change the way the clients works because now is tested very well, if I modify it then is like having another hivemq for native apps. Or I miss a piace of the puzzle ?

masini:
Hi @cescoffier the PR for the new BuildItem has been merged yesterday and it works, now the hivemq.jar archive is excluded during native build.

I also synched with the new 2.10-SNAPSHOT, and everything still works great.

If you agree I can create a PR for the SR MQTT Extension, where I've used the new Build Item to exclude it.

Let me know your thoughts...

cescoffier:
Great! 

We would need a separate extension. The current MQTT extension should still use the Vert.x based client. We can create the extension in Quarkiverse. 

We are able to reduce the number of classes requiring runtime definition. If the extension is in quarkiverse, it's not a problem, but in "quarkus core", we want to reduce it to the bare minimum.

masini:
I think the Quarkiverse is the correct place to go.

The workflow there is always the same ? I fork, write code and then PR for you ?

cescoffier:
@gastaldi can create the repository and do the onboarding (CI, release process...).

gastaldi:
@cescoffier @masini yes, we have the onboard steps documented in https://github.com/quarkiverse/quarkiverse/wiki#joining-quarkiverse

If you create the issue there I can set it up for you 😉

masini:
Here is the issue @gastaldi 

https://github.com/quarkusio/quarkus/issues/15944

What can I use for the artifactId and groupdId ?

masini:
Hi @cescoffier, I'm sorry to bother you, but I would like to understand how to use the new extension inside the quarkiverse. I tried and it works, but I neet to add that to the pom when I use mqtt-connector-hivemq inside SR.

How can we document that in the right way ? 

Regarding this PR, what do you think I still need to do ?

Thanks

cescoffier:
The extension should depend on the new connector (not the opposite) and your application should depend on the extension. 

I want to have another look at this connector, but I've, unfortunately, have other priorities at the moment (Quarkus 2.x and Kafka-related stuff).  

If you need something quick, you can just copy the code of the connector into the "runtime"  part of your extension.

masini:
Hi @cescoffier do you think we can merge this PR in the next minor release of Quarkus 1.13 ? 

I worked on the extension for the native part and all seems fine, when you have time (I know you're working hard on Quarkus 2) we can create a release of the extension and I can document that.

Or the only path to release is to copy the connector into the extension?

cescoffier:
This connector should be merged in the 3.x branch and so, would be available in the Quarkus 2.x. 
We had extended the other MQTT connector with certs support which would need to be backported in this one too.

cescoffier:
Closing - reconnect has been added to the other MQTT connector.

PR #1092, opened 2021-03-13
Implement default and named config lookup for the kafka connector

cescoffier:
Implement default and named configuration lookup for the Kafka connector.

This allows the connector to resolve the configuration from named `Map` (exposed using CDI).﻿


cescoffier:
@geoand this is related to service binding with multiple brokers.

geoand:
Cool!

Ladicek:
I wanted to file an issue eventually, but it seems I should do it sooner rather than later. We should not use `@Named`, it doesn't work as all other normal qualifiers do. (I'm sure @mkouba could swamp you with details :-) )

I'll file an issue and add the link here, before we spread `@Named` too much.

Ladicek:
Filed https://github.com/smallrye/smallrye-reactive-messaging/issues/1096

cescoffier:
Good point. In that case, we would need another qualifier. Something we could add to smallrye-annotations?

Ladicek:
Would be happy to review this when rebased and using `@Identifier`.

cescoffier:
Ah yes, good point, this need to be revisited.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
```diff
@@ -37,6 +37,10 @@ Type: _boolean_ | false | `true`
 
 Type: _boolean_ | false | `true`
 
+| *kafka-configuration-name* | The name set in `javax.inject.Named` of a bean that provides the default Kafka consumer/producer (`Map<String, Object>`) configuration to use for this channel. The channel configuration can still override any attribute.
```
wtrocki, line 40: So we can use for this name of the service binding that will somehow work or actually do provide extra field that will be uniquely naming the service:

I have proposed something to the spec (but we can do it ourselves)
https://github.com/k8s-service-bindings/spec/issues/187

We can do it for kafka and any future services.

[Note] this is just though - binding name is ok for now

cescoffier:
I've recreated the PR:

1. From the main branch
2. Using @Identifier 
3. Cleaning up how we deal with the config in the connector

Ladicek:
Looks fine otherwise. I briefly considered whether we want to add support for `@Named` for these additional config `Map`s, but probably no.

cescoffier:
Unlike the default lookup (which still uses @Named to not break), this feature should use @Identifier. The pending work is to generate the Maps in Quarkus - which uses @Identifier already.

I was wondering if we wanted to remove the @Named support for the default strategy. I kept it, for now, I need to check with the consumers if they use that or not.

Ladicek:
We should remove support for `@Named` one day, but I agree today is not that day.

ozangunalp:
Can we avoid this `@Named` / `@Identifier` issue altogether by removing `-name` suffix off the config property? Same applies for the rebalance listener.

cescoffier:
@Ladicek I've applied your suggestion (was a left-over of the very old PR)
@ozangunalp I've removed the `-name` suffix.

documentation/src/main/doc/modules/kafka/pages/default-configuration.adoc
```diff
@@ -37,3 +37,24 @@ This previous example would extract all the configuration keys from MicroProfile
 ====
 Starting with Quarkus 1.5, a map corresponding to the previous example is automatically provided.
 ====
+
+In addition to this default configuration, you can configure the name of the `Map` producer using the `kafka-configuration-name` attribute:
```
Ladicek, line 41: This part of documentation still uses the `-name` suffix :-)

ozangunalp:
@cescoffier this should be rebased/reviewed after merging #1406.

cescoffier:
If I’m not mistaken, there are 3 Pr updating the same files. I need to see
the merge order to simplify the rebase.

cescoffier:
I removed all the remaining reference to kafka-configuration-name.

ozangunalp (approved):

PR #1108, opened 2021-03-17
Use @Identifier instead of @Named

### LINKED ISSUE #1096: stop using `@Named`
While I understand the desire to use stringly-typed injection from time to time (and I think it's OK), the `@Named` qualifier is a wrong way to do it. In CDI, `@Named` doesn't work as all other normal qualifiers do.

Imagine this:

```java
@ApplicationScoped
public class MapProducer {
    @Produces
    @Named("my-name")
    Map<String, Object> map() {
        Map<String, Object> result = new HashMap<>();
        result.put("foo", "bar");
        result.put("baz", "quux");
        return result;
    }
}

@ApplicationScoped
public class MapConsumer {
    @Inject
    Map<String, Object> map;

    void init(@Observes @Initialized(ApplicationScoped.class) Object event) {
        System.out.println(map);
    }
}
```

Note that the injection point in `MapConsumer` doesn't declare a qualifier -- yet the `init` method prints `{foo=bar, baz=quux}`.

I'm pretty sure we (in SRye RM) didn't intend this, but that's what happens.

I'm currently aware of these usages of `@Named`:

- if you create a bean of type `KafkaConsumerRebalanceListener` qualified with `@Named("some-name")`, it's used as a consumer rebalance listener for consumers of channels that declare `mp.messaging.incoming.*.consumer-rebalance- listener.name=some-name`, or alternatively, for consumers in the consumer group `some-name`;
- if you create a bean of type `Map<String, Object>` qualified with `@Named("default-kafka-broker")`, it's used as a default config for all Kafka-based channels;
- this [PR](https://github.com/smallrye/smallrye-reactive-messaging/pull/1092) proposes that if you create a bean of type `Map<String, Object>` qualified with `@Named("some-name")`, it's used as a default config for those Kafka-based channels that declare `mp.messaging.incoming.*.kafka-configuration-name=some-name`.

There might be more that I'm not aware of, but I hope not :-)

My suggestion would be to create dedicated qualifier annotations. For the 3 cases listed above, respectively:

- `@[Kafka]ConsumerRebalanceListenerName("...")`
- `@DefaultKafkaConfiguration` (or `@KafkaDefaultConfiguration`)
- `@KafkaConfiguration("...")`

WDYT?

CC @mkouba as the tireless fighter against `@Named` misuse :-)

Ladicek:
Support for `@Named` is retained, as well as a couple of tests,
but `@Identifier` is preferred now. At some point, support for
`@Named` should be removed altogether.

Resolves #1096

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
@@ -9,6 +9,7 @@ import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.literal.NamedLiteral;
 
 import io.smallrye.common.annotation.Identifier;
+import io.smallrye.reactive.messaging.i18n.ProviderLogging;
 import io.vertx.amqp.AmqpClientOptions;
 import io.vertx.mutiny.amqp.AmqpClient;
 import io.vertx.mutiny.core.Vertx;
@@ -39,6 +40,9 @@ public class AmqpClientHelper {
         if (options.isUnsatisfied()) {
             // this `if` block should be removed when support for the `@Named` annotation is removed
             options = instance.select(NamedLiteral.of(optionsBeanName));
+            if (!options.isUnsatisfied()) {
+                ProviderLogging.log.deprecatedNamed();
+            }
         }
         if (options.isUnsatisfied()) {
             throw ex.illegalStateFindingBean(AmqpClientOptions.class.getName(), optionsBeanName);
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.literal.NamedLiteral;
 
+import io.smallrye.common.annotation.Identifier;
 import io.vertx.amqp.AmqpClientOptions;
 import io.vertx.mutiny.amqp.AmqpClient;
 import io.vertx.mutiny.core.Vertx;
@@ -34,7 +35,11 @@ public class AmqpClientHelper {
 
     static AmqpClient createClientFromClientOptionsBean(Vertx vertx, Instance<AmqpClientOptions> instance,
             String optionsBeanName) {
-        Instance<AmqpClientOptions> options = instance.select(NamedLiteral.of(optionsBeanName));
+        Instance<AmqpClientOptions> options = instance.select(Identifier.Literal.of(optionsBeanName));
+        if (options.isUnsatisfied()) {
+            // this `if` block should be removed when support for the `@Named` annotation is removed
+            options = instance.select(NamedLiteral.of(optionsBeanName));
+        }
         if (options.isUnsatisfied()) {
             throw ex.illegalStateFindingBean(AmqpClientOptions.class.getName(), optionsBeanName);
         }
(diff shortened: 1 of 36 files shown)


Ladicek:
Draft for now, as there's no release of SmallRye Common with `@Identifier` yet.

radcortez:
I'm planning to do one today.

radcortez:
Done :)

Ladicek:
Thanks!

Ladicek:
Rebased and updated to SmallRye Common 1.6.0.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
```diff
@@ -34,7 +35,11 @@ static AmqpClient createClient(AmqpConnector connector, AmqpConnectorCommonConfi
 
     static AmqpClient createClientFromClientOptionsBean(Vertx vertx, Instance<AmqpClientOptions> instance,
             String optionsBeanName) {
-        Instance<AmqpClientOptions> options = instance.select(NamedLiteral.of(optionsBeanName));
+        Instance<AmqpClientOptions> options = instance.select(Identifier.Literal.of(optionsBeanName));
+        if (options.isUnsatisfied()) {
+            // this `if` block should be removed when support for the `@Named` annotation is removed
+            options = instance.select(NamedLiteral.of(optionsBeanName));
```
cescoffier, line 41: Should we add a warning for the user to invite him to switch to the new annotation?

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsConnector.java
```diff
@@ -157,7 +160,12 @@ private ConnectionFactory pickTheFactory(String factoryName) {
         if (factoryName == null) {
             iterator = factories.iterator();
         } else {
-            iterator = factories.select(NamedLiteral.of(factoryName)).iterator();
+            Instance<ConnectionFactory> matching = factories.select(Identifier.Literal.of(factoryName));
+            if (matching.isUnsatisfied()) {
+                // this `if` block should be removed when support for the `@Named` annotation is removed
+                matching = factories.select(NamedLiteral.of(factoryName));
```
cescoffier, line 166: Should we add a warning to invite the user to switch to the new annotation?

cescoffier (approved):
LGTM! Thanks!

I've made a few comments about adding a warning to invite the user to switch to the new annotation. WDYT?

Ladicek:
Added a commit with this (`@Once` is really nice!). I had to promote the test-scoped dependency on `smallrye-reactive-messaging-provider` in `smallrye-reactive-messaging-jms` to compile-scoped -- I guess that should be fine, as all other modules have it in compile scope as well.

cescoffier:
Great! I didn't know `@Once`.

cescoffier (approved):

cescoffier:
Cannot backport to 3.x due to merge conflicts. Please backport manually:
Run:
```
git clone https://github.com/smallrye/smallrye-reactive-messaging.git
git fetch origin pull/1108/head:pr-1108
git checkout -b 3.x origin/3.x
git checkout -b backport-#1108-to-3.x
# One or more of the following command will fail, you will need to fix the conflict manually
git cherry-pick 3fea5e3c3d0e9e7d1313dfcce734dd0866772ab0
git cherry-pick 4267584974999731a402f093db5af2b8dbff718d
# Once all commits have been cherry-picked:
git push --set-upstream origin backport-#1108-to-3.x
```
To fix the conflict, first check which file is impacted using: `git status`
For each file with a resolved conflict, execute: `git add $FILE`
Then, commit the files using the same commit message as the original commit: `git commit -m "..."`

Once done and pushed, open the pull request.

* Title: [3.x] Backport Use @Identifier instead of @Named
* Message: Backport #1108 to 3.x.
* ⚡ **Set the target branch to 3.x** 
* Set the milestone and the labels if needed

Ladicek:
Do you want me to submit a PR against `3.x`? I can do that, just don't want to conflict with you.

cescoffier:
I just did and added you as reviewer :-D

PR #1114, opened 2021-03-18
[3.x] Backport Use @Identifier instead of @Named

cescoffier:
Backport #1108 to 3.x.


Ladicek (approved):

Ladicek:
Too bad you had to change the log message code :-(

cescoffier:
Yes, the codes are always a problem...

PR #1132, opened 2021-03-25
Fix MQTT SSL Helpers to better handle misconfigurations

cescoffier:


cescoffier:
Created Backports: 
- #1133 to [backport-#1132-to-3.x](https://github.com/smallrye/smallrye-reactive-messaging/tree/3.x)

PR #1154, opened 2021-04-14
improve error handling in the AMQP connector

### LINKED ISSUE #1125: AMQP producer requests credits asynchronously, leading to possible subscription error
I'm playing with the AMQP connector and found an interesting thing that I'd personally call a bug.

I have a textbook producer configured to produce messages into a queue in an Artemis broker:

```java
@Outgoing("ticks-producer")
public Multi<Message<String>> produce() {
    return Multi.createFrom()
            .ticks().every(Duration.ofMillis(1000))
            .map(it -> Message.of("" + it));
}
```

There's nothing else in the application -- it just pushes messages into a queue. However, when I run it, I get nothing -- no messages appear in the queue, no error message is printed to the log, nothing.

I added a few debug prints and found that `AmqpCreditBasedSender.onError` is called early on with a backpressure error: the ticks stream can't produce anything, because noone has requested any items. I can easily make the whole thing work by the following little change:

```java
@Outgoing("ticks-producer")
public Multi<Message<String>> produce() {
    return Multi.createFrom()
            .ticks().startingAfter(Duration.ofMillis(2000)).every(Duration.ofMillis(1000))
            .map(it -> Message.of("" + it));
}
```

Here, the producer takes a little break at the beginning, which is enough for the async process started in `AmqpCreditBasedSender.getSenderAndCredits` to finish. That, in turn, requests items from the source stream, and all goes well.

I'm not exactly sure where the problem is -- I'm not well-versed in the Reactive Streams request protocol, so I'm not sure if `Subscription.request` is supposed to work synchronously or if asynchrony is permitted. I first thought that it must be synchronous, so I rewrote the `AmqpCreditBasedSender.request` method to this:

```java
public void request(long l) {
    if (!once.getAndSet(true)) {
        try {
            getSenderAndCredits().await().indefinitely();
        } catch (Exception e) {
            downstream.get().onError(e);
        }
    }
}
```

and indeed that also fixed the problem. But it sounds more like a workaround, and seems to not comply with the Reactive Streams specification.

Full reproducer is here: https://github.com/Ladicek/artemis-playground Run the Artemis broker per the README, then run the `ProducerMain` class from the `amqp/producer` module.

Ladicek:
The outgoing part of the AMQP connector used to silently mark
the channel as closed in case of an error. Now, the error is
also properly logged.

Resolves #1125

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -219,7 +219,10 @@ public class AmqpConnector implements IncomingConnectorFactory, OutgoingConnecto
 
         return ReactiveStreams.<Message<?>> builder()
                 .via(processor)
-                .onError(t -> opened.put(oc.getChannel(), false))
+                .onError(t -> {
+                    log.failureReported(oc.getChannel(), t);
+                    opened.put(oc.getChannel(), false);
+                })
                 .ignore();
     }
 


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -219,7 +219,10 @@ void setup(ExecutionHolder executionHolder) {
 
         return ReactiveStreams.<Message<?>> builder()
                 .via(processor)
-                .onError(t -> opened.put(oc.getChannel(), false))
+                .onError(t -> {
+                    log.failureReported(oc.getChannel(), t);
+                    opened.put(oc.getChannel(), false);
+                })
```
Ladicek, line 225: I have contemplated calling `reportFailure(oc.getChannel(), t)`, which is defined as:

```java
public void reportFailure(String channel, Throwable reason) {
    log.failureReported(channel, reason);
    opened.put(channel, false);
    terminate(null);
}
```

However, the `terminate` method seems to close all channels managed by this connector. I can only see one reason to do that: application shutdown (and the `terminate` method already observes the corresponding CDI event). There's one caller of `reportFailure`, which I don't pretend to understand.

So I left the `reportFailure` method alone and did just this :-)

cescoffier: In Reactive Streams, failures are terminal. 
But I would add a "fatal" flag (as in the Kafka connector), to avoid stopping everything if we can recover.

cescoffier (approved):

cescoffier:
Created Backports: 
- #1157 to [backport-#1154-to-3.x](https://github.com/smallrye/smallrye-reactive-messaging/tree/3.x)

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
Ladicek: Yea I understand that reactive streams are always `onItem* (onComplete|onError)`. What I find weird is that we have a piece of code (`AmqpFailStop.handle`) calling `terminate` (indirectly, through `AmqpConnector.reportFailure`), which terminates all running streams in the connector (if I understand it correctly). I'm sure there was a reason, I just don't see it :-)

PR #1166, opened 2021-04-19
[sr-rm-kafka] Add a `bean-discovery-mode` attribute into beans.xml

jstourac:
To comply with the XSD definition file of beans.xml [1].

This change is motivated with a warning message present in WildFly
from the Weld subsystem after incorporating SmallRye ReactiveMessaging
project, see [2].

[1] http://xmlns.jcp.org/xml/ns/javaee/beans_2_0.xsd
[2] https://issues.redhat.com/browse/WFLY-14643

Note: there are much more `beans.xml` files that would deserve such update. My intention was to make this as small change as possible. Please, let me know whether you feel this is okay for you.


cescoffier (approved):
Thanks

jstourac:
Thank you, @cescoffier. I went on and raised also a PR against `main` branch, since this is what dependable bot does too. Feel free to reject it in case this is not expected - https://github.com/smallrye/smallrye-reactive-messaging/pull/1167.

PR #1167, opened 2021-04-20
[sr-rm-kafka] Add a `bean-discovery-mode` attribute into beans.xml

jstourac:
To comply with the XSD definition file of beans.xml [1].

This change is motivated with a warning message present in WildFly
from Weld subsystem after incorporating SmallRye ReactiveMessaging
project, see [2].

[1] http://xmlns.jcp.org/xml/ns/javaee/beans_2_0.xsd
[2] https://issues.redhat.com/browse/WFLY-14643

For reference - [PR into 3.x](https://github.com/smallrye/smallrye-reactive-messaging/pull/1166)


cescoffier:
Thanks!

PR #1168, opened 2021-04-20
fix lookup of Kafka default configuration

Ladicek:


Ladicek:
Motivation is https://github.com/quarkusio/quarkus/issues/16386#issuecomment-823179215

Ladicek:
I'll submit a PR for the 3.x branch in a bit, but I wonder, do we still use the 2.x stream somewhere? Maybe we should rename `main` to `2.x` and then rename `3.x` to `main`.

cescoffier (approved):
Oh nice catch!

We would need to backport this to 3.x (the 'main' development stream).

cescoffier:
So 2.x (current main) is Quarkus 1.x
3.x is Quarkus 2.x and EAP.

I will rename main into 2.x.

Ladicek:
I've also submitted a PR for 3.x: https://github.com/smallrye/smallrye-reactive-messaging/pull/1169

PR #1169, opened 2021-04-20
fix lookup of Kafka default configuration

Ladicek:


Ladicek:
Motivation is https://github.com/quarkusio/quarkus/issues/16386#issuecomment-823179215

PR #1177, opened 2021-04-26
Provide access to the low-level Kafka client

### LINKED ISSUE #955: Access the low level Kafka consumer / producer from the Kafka connector
For some low level stuff, we need to have access from the channel the corresponding Kafka Consumer / Producer.

Idealy, those needs to be recoverable from CDI (directly of via the connector).

This allow for example to manyally pause a consumer.

cescoffier:
Fix #955

This is done using the `io.smallrye.reactive.messaging.kafka.KafkaClientService` (experimental API for now, so not documented on purpose).

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaClientService.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaClientService.java
@@ -1,5 +1,7 @@
 package io.smallrye.reactive.messaging.kafka;
 
+import org.apache.kafka.clients.producer.Producer;
+
 import io.smallrye.common.annotation.Experimental;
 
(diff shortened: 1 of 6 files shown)


Ladicek (approved):

cescoffier:
Using a producer in the middle of an `@Incoming` / `@Outgoing` method is fine.
Using a consumer would work more of the time as you will schedule an action that will happen after (Kafka consumer are single-threaded). 
But I agree we need to see how it's used in practice, that's why I made it experimental. 

The main use case we had was about pausing/resuming (which can also be a recipe for disaster if you don't understand what it means for Kafka). Fortunately, that use case is less prominent now that the connector does this for you automatically.

Ladicek:
Hmm, yea, I guess if your `@Incoming`/`@Outgoing` methods have the "invoked once per message" signature, then pausing/resuming may make sense and do what you wish...

PR #1199, opened 2021-05-06
allow modifying Kafka metadata when sending a failed incoming message to a dead letter topic

### LINKED ISSUE #1103: Add custom header to message before it is sent to dead letter queue
Hi! 

I'm trying to add a custom header to the message I fail to process, but as far as I can see the header is not there in the new event in the dead letter queue. 

Here's a snippet of my application that adds the custom header before I send the `nack`

```java 
    protected CompletionStage<Void> handleMessage(final Message<String> message) {
        try {
            CloudEventUtils.decode(message.getPayload()).ifPresent(this::handleCloudEvent);
        } catch (Exception e) {
            Optional<IncomingKafkaRecordMetadata> metadata = message.getMetadata(IncomingKafkaRecordMetadata.class);
            int retryCounter = 1;
            if (metadata.isPresent()) {
                LOG.info("metadata present");
                Header header = metadata.get().getHeaders().lastHeader("my-header");
                if (header != null) {
                    LOG.info("header present");
                    retryCounter = intFromByteArray(header.value());
                }
            } else {
                LOG.info("metadata not there");
            }
            if (retryCounter >= 10) {
                LOG.error("retry counter = 10");
                return message.ack();
            }
            message.withMetadata(Metadata.of(buildFailedCounterMetadata(retryCounter + 1)));
            return message.nack(e);
        }
        return message.ack();
    }

    private OutgoingKafkaRecordMetadata<String> buildFailedCounterMetadata(int value) {
        return OutgoingKafkaRecordMetadata.<String> builder()
                .withKey("failure-count-key")
                .withHeaders(new RecordHeaders()
                        .add("my-header", intToByteArray(value))
                        .add("nextTimestamp", intToByteArray((int) System.currentTimeMillis() / 1000 + 60)))
                .build();
    }
```

Is this scenario not supported or am I doing something wrong? Or am I facing a bug? If needed, I can create a separated reproducer. 

Thank you very much!

Ladicek:
Fixes #1103

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/OutgoingKafkaRecord.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/OutgoingKafkaRecord.java
@@ -198,12 +198,14 @@ public class OutgoingKafkaRecord<K, T> implements KafkaRecord<K, T> {
 
     @Override
     public OutgoingKafkaRecord<K, T> withMetadata(Iterable<Object> metadata) {
+        // TODO this adds the entire provided Iterable<Object> as a single datum in the existing Metadata
         Metadata newMetadata = getMetadata().with(metadata);
         return OutgoingKafkaRecord.from(Message.of(getPayload(), newMetadata, getAck(), getNack()));
     }
 
     @Override
     public OutgoingKafkaRecord<K, T> withMetadata(Metadata metadata) {
+        // TODO this adds the entire provided Metadata as a single datum in the existing Metadata
         Metadata newMetadata = getMetadata().with(metadata);
         return OutgoingKafkaRecord.from(Message.of(getPayload(), newMetadata, getAck(), getNack()));
     }
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
@@ -11,10 +11,12 @@ import java.util.Map;
 import java.util.concurrent.CompletionStage;
 
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.eclipse.microprofile.reactive.messaging.Metadata;
 
 import io.smallrye.reactive.messaging.kafka.IncomingKafkaRecord;
 import io.smallrye.reactive.messaging.kafka.KafkaCDIEvents;
 import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
+import io.smallrye.reactive.messaging.kafka.OutgoingKafkaRecordMetadata;
 import io.smallrye.reactive.messaging.kafka.impl.ConfigurationCleaner;
 import io.smallrye.reactive.messaging.kafka.impl.KafkaSource;
 import io.vertx.mutiny.core.Vertx;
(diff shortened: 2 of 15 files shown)


Ladicek:
The failure is in a Reactive Streams TCK (`KafkaClientReactiveStreamsPublisherTest>PublisherVerification.required_spec313_cancelMustMakeThePublisherEventuallyDropAllReferencesToTheSubscriber`), and I have no idea how anything in this PR could affect that.

Is that perhaps a known flaky test?

cescoffier:
Yes, it's a flaky test due to a broken behavior in mutiny (keeping a reference on the subscriber). I've fixed mutiny already. We should release and update. 

CC @jponge

jponge:
> Yes, it's a flaky test due to a broken behavior in mutiny (keeping a reference on the subscriber). I've fixed mutiny already. We should release and update.
> 
> CC @jponge

Yep, there are other nice bugs being fixed. Let's have the deprecations bundle in then release early next week

Ladicek:
Alright, thanks! Will rerun the tests then.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/OutgoingKafkaRecord.java
```diff
@@ -198,12 +198,14 @@ public String key() {
 
     @Override
     public OutgoingKafkaRecord<K, T> withMetadata(Iterable<Object> metadata) {
+        // TODO this adds the entire provided Iterable<Object> as a single datum in the existing Metadata
```
cescoffier, line 201: is this something to handle in this PR?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
```diff
@@ -96,8 +98,36 @@ private String getThrowableMessage(Throwable throwable) {
...
-            IncomingKafkaRecord<K, V> record, Throwable reason) {
-        KafkaProducerRecord<K, V> dead = KafkaProducerRecord.create(topic, record.getKey(), record.getPayload());
+            IncomingKafkaRecord<K, V> record, Throwable reason, Metadata metadata) {
+
+        OutgoingKafkaRecordMetadata<K> outgoing = null;
+        if (metadata != null) {
+            for (Object item : metadata) {
+                if (item instanceof OutgoingKafkaRecordMetadata) {
+                    outgoing = (OutgoingKafkaRecordMetadata<K>) item;
+                    // more than 1 instance => undefined behavior
```
cescoffier, line 108: actually, not supported by the metadata model.

cescoffier: We can add an accessor by type in `Metadata`:
```
Optional<OutgoingKafkaRecordMetadata> maybe = metadata.get(OutgoingKafkaRecordMetadata.class)
```

cescoffier (commented):
Looks good to me.

I've added a few comments, let me know what you think?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/OutgoingKafkaRecord.java
Ladicek: I noticed this while writing the test, but couldn't quite figure out a proper solution, so I decided to at least leave a TODO. I still need to wrap my head around the entire `Metadata` thing.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
Ladicek: Yea I noticed that `Message` has an accessor like this. Frankly, I don't see why `Metadata` are not backed by a `Map`, because what we're currently doing (see e.g. `replaceOrAdd`) is weird and inefficient (and, in case of `replaceOrAdd`, also wrong :-) ).

Let me take a crack at this.

Ladicek: Added a commit with refactored `Metadata` implementation (backing store is now a `Map<Class<?>, Object>`), and a `get(Class)` method.

Note that the signature of the `get` method I added is `<T> Optional<T> get(Class<T> clazz)`, so it looks up an object that is precisely of given class. The `Message.getMetadata(Class)` method is `<M> Optional<M> getMetadata(Class<? extends M> clazz)`, so it looks up an object that is of a subtype of given class. It looks strange to me, was that even intentional?

cescoffier (approved):
Thanks!

Ladicek:
Test failure was very much related, had to fix a NPE :-)

PR #1223, opened 2021-05-24
Fixes #1222

### LINKED ISSUE #1222: Disable Vertx Tracing when producing Kafka Messages
To prevent duplication of spans between Reactive Message and Vert.x Tracer, we need to disable tracing from Vert.x when we're producing Kafka messages and we've created the span ourselves

kenfinnigan:
Disable Vert.x Tracing policy when we have tracing enabled for producing messages to Kafka


kenfinnigan:
Would be great if we can get a 3.4.0 with this fix on Tuesday for inclusion into Quarkus 2.0 CR1!

cescoffier (approved):

PR #1227, opened 2021-05-26
manage Kafka sending thread manually instead of using Vert.x Kafka client

Ladicek:
When we have moved from using the Vert.x Kafka client for consumers
to using our own managed thread ("polling thread"), we kept using
the Vert.x Kafka client for producers.

With this commit, we no longer use the Vert.x Kafka client even for
producers. Similarly to consumers, we now manage our own thread
("sending thread") and delegate all sending operations to it.

This entails one breaking change: where we previously exposed
Kafka's `Producer`, we now expose our own `KafkaProducer`.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaClientServiceImpl.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaClientServiceImpl.java
@@ -5,12 +5,12 @@ import java.util.Objects;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
-import org.apache.kafka.clients.producer.Producer;
 import org.eclipse.microprofile.reactive.messaging.spi.Connector;
 
 import io.smallrye.reactive.messaging.kafka.KafkaClientService;
 import io.smallrye.reactive.messaging.kafka.KafkaConnector;
 import io.smallrye.reactive.messaging.kafka.KafkaConsumer;
+import io.smallrye.reactive.messaging.kafka.KafkaProducer;
 
 @ApplicationScoped
 public class KafkaClientServiceImpl implements KafkaClientService {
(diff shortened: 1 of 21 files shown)


Ladicek:
I spent some time investigating the performance impact. The Vert.x Kafka client based producer is slightly more performant than this [1], but that is more than compensated by the optimization of `Message.getMetadata(Class)`.

I need to write tests and documentation before this is actually ready (but I guess it's worth reviewing already). The breaking change is something to discuss (but I'd say we should do it).

[1] I currently guess it's caused by how the Vert.x Kafka client handles threading. The `send` operation is offloaded to a worker pool, and when the operation finishes, result handling is moved back to the event loop. The present implementation doesn't move back after sending completes, so the sending thread probably does a lot more work than the worker thread.

Ladicek:
Added tests and documentation, marking ready for review.

I also did some tests locally where I either shutdown the Kafka broker in the middle of the test, or induce high latency on the localhost interface (using `tc`). In both cases, I didn't see a behavioral change: the messages were properly nacked (when the broker was down / behind high latency), and when I removed the error condition, everything healed and messages were again properly acked. I didn't observe a single lost message.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaClientServiceImpl.java
```diff
@@ -25,7 +25,7 @@
     }
 
     @Override
-    public <K, V> Producer<K, V> getProducer(String channel) {
+    public <K, V> KafkaProducer<K, V> getProducer(String channel) {
```
cescoffier, line 28: That is a breaking change, so we need to remember to mention it to the migration guide (in Quarkus)

cescoffier (approved):

cescoffier:
Can we remove the dependency on vertx-kafka-client?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaClientServiceImpl.java
Ladicek: +1

Ladicek:
The Vert.x Kafka client is still used for health or something like that.

cescoffier:
Right, the admin client...

Ladicek:
Yep, that's it!

PR #1233, opened 2021-05-31
Gracefully handle serialisation failure in the AMQP connector

cescoffier:
Related to https://github.com/quarkusio/quarkus/issues/17511.
Note that the issue reported in https://github.com/quarkusio/quarkus/issues/17511 was already fixed. This PR improves the reporting.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpSinkTest.java
+++ b/smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpSinkTest.java
@@ -1,9 +1,12 @@
 package io.smallrye.reactive.messaging.amqp;
 
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+import static org.awaitility.Awaitility.await;
 
 import java.nio.charset.StandardCharsets;
 import java.util.*;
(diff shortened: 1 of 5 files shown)


smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpSinkTest.java
```diff
@@ -242,12 +253,77 @@ public void testSinkUsingObject() throws Exception {
...
+            Binary bin = ((Data) body).getValue();
+            byte[] bytes = Binary.copy(bin).getArray();
+            JsonObject json = Buffer.buffer(bytes).toJsonObject();
+            Person p = json.mapTo(Person.class);
+            assertThat(p.getName()).startsWith("hello-");
+            count.incrementAndGet();
+        });
+
+        assertThat(count.get()).isEqualTo(msgCount / 2);
+        assertThat(nack).hasValue(5);
```
Ladicek, line 317: This is useless, because we already do `await().until(() -> nack.get() == 5)` above, but not a big deal :-)

Ladicek (approved):

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpSinkTest.java
cescoffier: Ah ah, I added the await after this line (because obviously, it was racy).

PR #1260, opened 2021-06-15
Fix #1259 - Do not fail because on uncommitted records if the consumer do not own the partition anymore

### LINKED ISSUE #1259: Kafka: Liveness probe fails after rebalance took place
I'm using quarkus 2.0.0.CR3 and therefore smallrye kafka 3.3.2

Consuming a high velocity stream with multiple consumers in a group.
Using `commit-strategy: throttled`

Incoming channel is annotated like this:
```
  @Incoming("events")
  @Blocking
  @Acknowledgment(Strategy.MANUAL)
```
In here I have some async operation (Elasticsearch BulkPorcessor) etc. which ack / nacks the messages.

When a new consumer joins on the others I get

`SRMSG18239: Acked record 8041153253 on group 'group' was ignored because the topic partition 'topic-4' was revoked for this instance. Record will likely be processed again.`

I'm ok with some duplicates, so far so good.

After a few seconds on the next livenes check I get

`SRHCK01001: Reporting health down status: {"status":"DOWN","checks":[{"name":"SmallRye Reactive Messaging - liveness check","status":"DOWN","data":{"events":"[KO] - Too Many Messages without acknowledgement in topic topic (partition:4), last committed offset is 7980556233Too Many Messages without acknowledgement in topic topic (partition:7), last committed offset is 7868958352"}}]}`

I cannot see any other errors or nacked messages. It looks like the messages which are acked from the aync operation after the rebalance took place count into the number causing 'Too Many Messages without acknowledgement' and cause the liveness to get down. Eventually they are also leaking memory.

Am I doing something wrong?
Is there a way to get rid of these messages which are acked on a revoked partition?

cescoffier:
After a rebalance, ignore uncommitted records on revoked partitions, they will be cleared eventually.
The main change is the tracking of the assignments and checking if the partition is still assigned to the consumer before throwing the exception.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
@@ -47,6 +47,7 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
     private final int unprocessedRecordMaxAge;
     private final int autoCommitInterval;
     private volatile long timerId = -1;
+    private final Collection<TopicPartition> assignments = new CopyOnWriteArrayList<>();
 
     private KafkaThrottledLatestProcessedCommit(
             String groupId,
@@ -110,7 +111,7 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
     public void partitionsAssigned(Collection<TopicPartition> partitions) {
         runOnContextAndAwait(() -> {
             stopFlushAndCheckHealthTimer();
-
+            assignments.addAll(partitions);
             if (!partitions.isEmpty() || !offsetStores.isEmpty()) {
                 startFlushAndCheckHealthTimer();
             }
@@ -123,11 +124,13 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
      * Revoked partitions.
      * This method is called from the Kafka pool thread.
      *
-     * @param partitions The list of partitions that were assigned to the consumer on the last rebalance
+     * @param partitions The list of partitions that were assigned to the consumer and now need to be revoked
+     *        (may not include all currently assigned partitions).
      */
     @Override
     public void partitionsRevoked(Collection<TopicPartition> partitions) {
         runOnContextAndAwait(() -> {
+            assignments.removeAll(partitions);
             stopFlushAndCheckHealthTimer();
             return null;
         });
@@ -193,7 +196,6 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
     @Override
     public <K, V> IncomingKafkaRecord<K, V> received(IncomingKafkaRecord<K, V> record) {
         TopicPartition recordsTopicPartition = getTopicPartition(record);
-
         offsetStores
                 .computeIfAbsent(recordsTopicPartition, k -> new OffsetStore(k, unprocessedRecordMaxAge))
                 .received(record.getOffset());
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/commit/DeprecatedCommitStrategiesTest.java
```diff
@@ -323,6 +326,7 @@ void testThrottledStrategyWithTooManyUnackedMessages() throws Exception {
 
         consumer.schedulePollTask(() -> {
             consumer.rebalance(offsets.keySet());
+            source.getCommitHandler().partitionsAssigned(offsets.keySet());
```
Ladicek, line 329: I'd expect identical changes in the `CommitStrategiesTest` -- this one is a copy of that using `@Named` instead of `@Identifier`.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -378,6 +380,11 @@ boolean hasTooManyMessagesWithoutAck() {
             return false;
         }
 
+        private boolean isStillAssigned() {
+            // If the topic/partition is not assigned to us already, the store will be cleared eventually.
+            return assignments.contains(topicPartition);
```
Ladicek, line 385: Don't `offsetStores` include the same information? In other words, could this method be implemented as `return offsetStore.containsKey(topicPartition);`?

cescoffier: No, because the deletion happens asynchronously, and so there is a window where it would still be there, while not being assigned anymore.

Ladicek:
Can't say I truly understand the code, but the logic seems fine. Added just 2 small comments.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
Ladicek: Ah interesting!

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/commit/DeprecatedCommitStrategiesTest.java
cescoffier: Fixed.

FYI, the mock consumer does not handle rebalance (does not call the listener) but we need to call rebalance to set up the assigned partitions. So adding this second invocation simulates a "real" rebalance.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -47,6 +47,7 @@
     private final int unprocessedRecordMaxAge;
     private final int autoCommitInterval;
     private volatile long timerId = -1;
+    private final Collection<TopicPartition> assignments = new CopyOnWriteArrayList<>();
```
Ladicek, line 50: Just today morning, I realized I saw `CopyOnWriteArrayList` here. I think `Collections.newSetFromMap(new ConcurrentHashMap<>())` would probably be more appropriate, given that you only use `addAll`, `removeAll` and `contains`.

cescoffier: Hum, very good point! 
Fancy a PR (this one was already merged).

Ladicek: Sure, gimme a sec :-)

Ladicek: #1261#1261

PR #1263, opened 2021-06-16
Add OpenTelemetry tracing to AMQP Connector

### LINKED ISSUE #562: OpenTelemetry support for Kafka and AMQP

kenfinnigan:
Resolves #562

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-amqp-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-amqp-incoming.adoc
@@ -81,6 +81,10 @@ _(amqp-client-options-name)_ | The name of the AMQP Client Option bean used to c
 
 Type: _string_ | false | 
 
+| *tracing-enabled* | Whether tracing is enabled (default) or disabled
+
+Type: _boolean_ | false | `true`
+
 | *broadcast* | Whether the received AMQP messages must be dispatched to multiple _subscribers_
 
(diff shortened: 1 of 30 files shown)


kenfinnigan:
I need to verify this works as expected in Quarkus, but it's very similar to the Kafka approach so should be good

cescoffier (approved):

cescoffier:
Hum, that does not seem to work. The propagation tests are failing.

kenfinnigan:
Weirdly one of the four tests fails on JDK 11 with a condition timeout, and two of the four fail on JDK 16 with condition timeout.

But JDK 8 and 15 were fine.

kenfinnigan:
Yay! @cescoffier could you re-review?

cescoffier (approved):

cescoffier:
Just in time for the release :-D

kenfinnigan:
phew!

PR #1266, opened 2021-06-18
Kafka processor consuming parameter type can be message or payload

ozangunalp:
Currently supported processor signatures for single items are : 
- `Message<O> method(Message<I> msg)`
- `O method(I payload)`
- `CompletionStage<O> method(I payload)`
- `CompletionStage<Message<O>> method(Message<I> msg)`
(And `Uni<O>` variants)

A processor can support `Message` or payload object in consumption parameter type such as :
- `Message<O> method(I payload)`
- `CompletionStage<O> method(<Message<O> msg)`

Whereas currently these signatures result in ClassCastExceptionn: class [PAYLOAD] cannot be cast to class org.eclipse.microprofile.reactive.messaging.Message

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,7 @@ root = true
 charset = utf-8
 indent_style = space
 indent_size = 4
+ij_continuation_indent_size = 8
 trim_trailing_whitespace = true
 end_of_line = lf
 insert_final_newline = true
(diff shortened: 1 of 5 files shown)


Ladicek (approved):
Nice!

Ladicek:
Except there's a failing test. I _guess_ it isn't caused by this PR; instead, it seems to be a flaky test introduced in #1263.

ozangunalp:
Yes, it seems like a flaky test. I can't re-run the failing build. Can you @Ladicek ?

cescoffier (approved):

PR #1267, opened 2021-06-20
added support for headers on camel's OutgoingExchangeMetadata

mswiderski:
A very common need when working with camel is to set header that actually drive how given component is used. For example when using camel drops component to upload files there is a need to set the header `CamelDropboxPutFileName` that will configure the name of the file at the remote drive. 

This pull request adds support for setting headers on Camel's in message via camel's `OutgoingExchangeMetadata`.

It is implemented pretty much the same way as the properties are that are put on the exchange level.


cescoffier (approved):

mswiderski:
@cescoffier the failure comes from amqp I don’t think it’s related to the changes in this PR, or?

cescoffier:
No, it's not related. We have a flaky test introduced when we added OptneTelemetry to AMQP. CC @kenfinnigan

cescoffier:
Merged manually.

mswiderski:
@cescoffier thanks for quick merge. May I ask what version of quarkus I might expect this to be in?

cescoffier:
2.1, as 2.0 is being cut this week (Wednesday). 
2.1 should be available in July.

mswiderski:
excellent, thanks @cescoffier

PR #1272, opened 2021-06-22
Create Kafka API module

kabir:


smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaRecordTest.java
```diff
@@ -41,6 +47,24 @@ public void testCreationOfMessageWithKeyAndValue() {
         assertThat(metadata.getTopic()).isNull();
         assertThat(metadata.getTimestamp()).isNull();
         assertThat(metadata.getHeaders()).isNull();
+        // TODO - we don't have duplicate metadata when the Message API is used directly
+        // KafkaTestBase.tempCompareLegacyAndApiMetadata(metadata, message);
```
kabir, line 51: This is a slight hole in the current implementation. But I hope that it will not be problematic.

My thinking is that when people set this metadata using one key they can be expected to look it up with the same key. As I understand it this will only be a problem before the message is sent? 

Once KafkaSink receives it, it will make the metadata available under both keys.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaRecordTest.java
```diff
@@ -41,6 +47,24 @@ public void testCreationOfMessageWithKeyAndValue() {
...
+                OutgoingKafkaRecordMetadata.<String> builder().withKey("boo").build());
+        OutgoingKafkaRecordMetadata<?> metadata = message.getMetadata(OutgoingKafkaRecordMetadata.class)
+                .orElseThrow(() -> new AssertionError("Metadata expected"));
+        assertThat(metadata.getPartition()).isEqualTo(-1);
+        assertThat(metadata.getKey()).isEqualTo("boo");
+        assertThat(metadata.getTopic()).isNull();
+        assertThat(metadata.getTimestamp()).isNull();
+        assertThat(metadata.getHeaders()).isNull();
+        // TODO - we don't have duplicate metadata when the Message API is used directly
+        // KafkaTestBase.tempCompareLegacyAndApiMetadata(metadata, message);
```
kabir, line 67: Same consideration as above

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSinkWithLegacyMetadataTest.java
```diff
@@ -0,0 +1,112 @@
...
+import io.smallrye.mutiny.Multi;
+import io.smallrye.reactive.messaging.kafka.base.KafkaMapBasedConfig;
+import io.smallrye.reactive.messaging.kafka.base.KafkaTestBase;
+import io.smallrye.reactive.messaging.kafka.impl.KafkaSink;
+
+/**
+ * Duplicate of {@link KafkaSinkTest} - delete once we remove the legacy
+ * {@link io.smallrye.reactive.messaging.kafka.api.OutgoingKafkaRecordMetadata}
+ * <p/>
+ * Tests not using metadata have been removed
```
kabir, line 37: The honest reason for this is that when keeping all the test methods, I got problems in one of the `testABeanProducingXXXX()` methods (I forgot which). The problem is that the `expected` AtomicInteger contained 20 rather than 10 entries.

It worked fine on its own, so it is probably some test setup pollution. I didn't try to track down the issue since I realised it didn't have anything to do with metadata

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceWithLegacyMetadataTest.java
```diff
@@ -0,0 +1,101 @@
...
+import io.smallrye.reactive.messaging.kafka.api.KafkaMetadataUtil;
+import io.smallrye.reactive.messaging.kafka.base.KafkaMapBasedConfig;
+import io.smallrye.reactive.messaging.kafka.base.KafkaTestBase;
+import io.smallrye.reactive.messaging.kafka.impl.KafkaSource;
+
+/**
+ * Duplicate of {@link KafkaSourceTest} - delete once we remove the legacy
+ * {@link io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordMetadata}
+ * <p/>
+ * Tests not using metadata have been removed
```
kabir, line 27: The main reason for this is that `testABeanConsumingTheKafkaMessagesStartingOnFifthOffsetFromLatestThatFailsUntilSecondRebalance()`  ended up with 
Gets ` [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]` rather than ` [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`.

It worked fine on its own, so again I think it is due to some pollution in the test setup. But I removed all non-metadata tests rather than investigate.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaRecordTest.java
cescoffier: I agree, not a problem.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceWithLegacyMetadataTest.java
cescoffier: Did the topic get reused no?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSinkWithLegacyMetadataTest.java
cescoffier: Make sure to use a unique topic.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/KafkaSourceWithLegacyMetadataTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private KafkaMapBasedConfig myKafkaSourceConfig(int partitions, String withConsumerRebalanceListener,
+            String group) {
+        KafkaMapBasedConfig.Builder builder = KafkaMapBasedConfig.builder("mp.messaging.incoming.data");
+        if (group != null) {
+            builder.put("group.id", group);
+        }
+        builder.put("value.deserializer", IntegerDeserializer.class.getName());
+        builder.put("enable.auto.commit", "false");
+        builder.put("auto.offset.reset", "earliest");
+        builder.put("topic", "data");
```
cescoffier, line 53: Use a unique topic.

cescoffier (changes requested):
Documentation must also be updated.

smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/KafkaMetadataUtil.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param msg the message
+     * @param outgoingKafkaRecordMetadata the {@link OutgoingKafkaRecordMetadata} to write. Must not be {@code null}.
+     * @throws NullPointerException if {@code msg} or {@code outgoingKafkaRecordMetadata} are {@code null}
+     */
+    public static <T> Message<T> writeOutgoingKafkaMetadata(Message<T> msg,
+            OutgoingKafkaRecordMetadata<T> outgoingKafkaRecordMetadata) {
+        return msg.addMetadata(outgoingKafkaRecordMetadata);
+    }
+}
```
Ladicek, line 35: It seems this util class is harder to use than the underlying `Message.addMetadata|getMetadata`? :-) Especially on the `write` side -- `Message` is immutable, so adding metadata item returns a new `Message`, which should at the very least be mentioned in a `@return` section in the javadoc.

Ladicek:
I've realized several times that the public API we expose in the Kafka connector is very large, but couldn't think of a good solution. I think this is a great step in the right direction. We'll want to add more classes to the API module (`KafkaConsumerRebalanceListener`, `KafkaClientService`, `KafkaConsumer`, `KafkaProducer`, maybe more), and eventually say that the API module is the _only_ public API we have. So +1_000_000 to this!

smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/KafkaMetadataUtil.java
kabir: The main reason for this, is that in WildFly/EAP I am not allowed to expose experimental features to users in areas where there is a spec. Essentially we need to stick with the latest version of the spec (i.e. MP RM 2.0). And in areas where there is not a spec, such as the Kafka API, we need to limit the API surface and have something as stable as possible, while avoiding pulling in non-spec things.

Although we use the SmallRye API at runtime, so we have access to e.g.`Metadata`, `Message.add/getMetadata()` etc. when running in the server, that becomes more for internal framework code. We 'hide' these things from users by exposing the spec RM 2.0 API only in our BOMs.

This util class is a way to side-step the problem. It makes it possible to use `Metadata` internally, while not exposing all the other stuff from the SmallRye API module to our users. (Additionally we do some deployment time checks to make sure clever users are not bypassing our restrictions by compiling against the SmallRye API rather then the MP RM 2.0.)

Quarkus, which is less strict in this regard, can continue to use the `Message.add/getMetadata()` methods.

Thanks for pointing out the javadoc, I will make that clearer :)

kabir:
> I've realized several times that the public API we expose in the Kafka connector is very large, but couldn't think of a good solution. I think this is a great step in the right direction. We'll want to add more classes to the API module (`KafkaConsumerRebalanceListener`, `KafkaClientService`, `KafkaConsumer`, `KafkaProducer`, maybe more), and eventually say that the API module is the _only_ public API we have. So +1_000_000 to this!

I'll add a few more of these at least in a future PR, I think we need some of these in WildFly too

kabir:
I think I've updated the docs now so they reference the new API classes. If you want to show using the `KafkaMetadataUtil` to access the metadata, let me know. I am unclear on if that should replace existing `Metadata.get/setMetadata()` examples, or if I should add to them as an alternative (possibly with some 'If you are running in an environment where you only have access to the spec classes, use this...' comment)?

cescoffier (approved):
It looks like a great first step.
We will probably move some of the API we are exposing in the connector into this module too, to create a clear distinction between the public API and the connector implementation.

The same approach may be reused for other connectors too (such as AMQP, even if the API surface is much smaller).

Ladicek:
Ah I think I finally understand the purpose of `KafkaMetadataUtil`. On WildFly, people have access to the "official" MP Reactive Messaging API (which is a subset of the SmallRye fork of the API), and will get access to this new API module. It's a bit unfortunate, but it's probably the best we can do here.

cescoffier:
@Ladicek yes. Metadata is not part of the spec. I plan to propose it but didn't have the time yet. Note that metadata has been discussed for a very long time, something like a year. I built it in smallrye based on the last discussion.

Ladicek:
After using the API a little, I personally think it should be slightly different (`Metadata` would be called `MetadataMap`, and only classes implementing a marker interface `Metadata` would be allowed as keys), but other than that, I can't see anything that would block adding this to the spec. So +1 to that :-)

kabir:
> It looks like a great first step.
> We will probably move some of the API we are exposing in the connector into this module too, to create a clear distinction between the public API and the connector implementation.
> 
> The same approach may be reused for other connectors too (such as AMQP, even if the API surface is much smaller).

I have some thoughts about this, so feel free to discuss when the time comes. Or if I get around to adding the next set of classes I need first, my attempt will contain the idea

PR #1282, opened 2021-06-28
OutgoingKafkaMetadata type should be that of the key, not the payload

kabir:


cescoffier (approved):
Good catch - definitely a bug

PR #1289, opened 2021-06-30
Fix Reactive Streams TCK 

cescoffier:
Attempt to understand what's going on with the Reactive Streams TCK


cescoffier:
Seems to be because of: 
```
2021-06-30T07:11:03.6433598Z Caused by: java.lang.NullPointerException: Cannot invoke "io.opentelemetry.api.trace.Tracer.spanBuilder(String)" because "io.smallrye.reactive.messaging.kafka.KafkaConnector.TRACER" is null
```

PR #1294, opened 2021-07-06
Kafka Retry test and documentation

cescoffier:
Add a few test to verify the combination of Kafka with Fault Tolerance (@Retry).
Also extend the documentation to explain how to use it.


ozangunalp (dismissed):
For the example, it'd be interesting to show @Retry on a @Incoming/@Outgoing method.

cescoffier:
I've added tests and docs about methods using `@Incoming` and `@Outgoing`.

ozangunalp (approved):

PR #1306, opened 2021-07-09
Provide a minimal wrapper around KafkaAdminClient for health check

ozangunalp:
Remove usages of vertx-kafka-client, change dependency to test scope.
Use an internal ReactiveKafkaProducer for sending messages in dead letter queue.

Because the vertx-kafka-client is no longer brought as a dependency, JsonObject serde won't be available by default in applications. This is a possible breaking change.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaAdmin.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaAdmin.java
@@ -1,0 +1,20 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import java.util.Set;
+
+import org.apache.kafka.clients.admin.Admin;
+
+import io.smallrye.mutiny.Uni;
+
+/**
+ * Internal interface for Kafka admin client.
+ * To complete with remaining {@link org.apache.kafka.clients.admin.Admin} method wrappers, if decided to expose it externally.
+ */
+public interface KafkaAdmin {
+
+    Uni<Set<String>> listTopics();
+
+    Admin unwrap();
+
+    void closeAndAwait();
+}
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
@@ -2,14 +2,18 @@ package io.smallrye.reactive.messaging.kafka.fault;
 
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 import static org.apache.kafka.clients.CommonClientConfigs.CLIENT_ID_CONFIG;
-import static org.apache.kafka.clients.consumer.ConsumerConfig.*;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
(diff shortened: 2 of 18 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaAdmin.java
```diff
@@ -0,0 +1,20 @@
...
+ * Internal interface for Kafka admin client.
+ * To complete with remaining {@link org.apache.kafka.clients.admin.Admin} method wrappers, if decided to expose it externally.
+ */
+public interface KafkaAdmin {
+
+    Uni<Set<String>> listTopics();
+
+    Admin unwrap();
+
+    void closeAndAwait();
```
cescoffier, line 19: I would return the Uni<Void> instead, as the listTopics is returning a Uni.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
```diff
@@ -141,8 +138,12 @@ private String getThrowableMessage(Throwable throwable) {
...
     }
 
+    void addHeader(ProducerRecord<?, ?> record, String key, String value) {
+        record.headers().add(key, value.getBytes(StandardCharsets.UTF_8));
+    }
+
     @Override
     public void terminate() {
-        producer.closeAndAwait();
+        producer.close();
```
cescoffier, line 147: What's the return type of close?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/BaseHealth.java
```diff
@@ -21,7 +21,7 @@ public BaseHealth(String channel) {
     }
 
     public void close() {
-        KafkaAdminClient admin = getAdmin();
+        KafkaAdmin admin = getAdmin();
         if (admin != null) {
             try {
                 admin.closeAndAwait();
```
cescoffier, line 27: should be admin.close().await().atMost(...)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaAdmin.java
cescoffier: Just got the problem with the Uni approach. Let's keep closeAndAwait.

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaAdmin.java
ozangunalp: We can wrap the AdminClient#close call inside a Uni but if we choose not to have a worker thread for this client it'll still block the caller thread.

cescoffier: Yes, but unnecessary I believe.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
ozangunalp: void. It'll wait for the duration of `close-timeout` config of the outgoing channel.

cescoffier (approved):

Ladicek:
> Use an internal ReactiveKafkaProducer for sending messages in dead letter queue.

I totally missed this when introducing `ReactiveKafkaProducer`, thanks for fixing that!

PR #1314, opened 2021-07-13
First commit of rabbitmq connector

wabrit:
This commit encompasses a new smallrye reactive connector that supports access to RabbitMQ via the AMQP 0-9-1 protocol.

The connector includes support for:

- Sending and receiving messages from a RabbitMQ broker
- Optional declaration Exchanges, Queues, Bindings as well as DLQ/DLX via configuration
- Health reporting on both incoming and outgoing channels

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/pom.xml
+++ b/smallrye-reactive-messaging-rabbitmq/pom.xml
@@ -1,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.smallrye.reactive</groupId>
+    <artifactId>smallrye-reactive-messaging</artifactId>
+    <version>3.6.0</version>
+  </parent>
+
+  <artifactId>smallrye-reactive-messaging-rabbitmq</artifactId>
+
+  <name>SmallRye Reactive Messaging : Connector :: RabbitMQ</name>
+
+
+  <dependencies>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-provider</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-mutiny-vertx-rabbitmq-client</artifactId>
+      <version>${smallrye-vertx-mutiny-clients.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-rabbitmq-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.config</groupId>
+      <artifactId>smallrye-config</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>test-common</artifactId>
+      <version>${project.parent.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-connector-attribute-processor</artifactId>
+      <version>${project.parent.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-semconv</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <generatedSourcesDirectory>${project.build.directory}/generated-sources/</generatedSourcesDirectory>
+          <annotationProcessors>
+            <annotationProcessor>
+              io.smallrye.reactive.messaging.connector.ConnectorAttributeProcessor
+            </annotationProcessor>
+            <annotationProcessor>
+              org.jboss.logging.processor.apt.LoggingToolsProcessor
+            </annotationProcessor>
+          </annotationProcessors>
+        </configuration>
+      </plugin>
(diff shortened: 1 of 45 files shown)


smallrye-reactive-messaging-rabbitmq/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+  <artifactId>smallrye-reactive-messaging-rabbitmq</artifactId>
+
+  <name>SmallRye Reactive Messaging : Connector :: RabbitMQ</name>
+
+
+  <dependencies>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-provider</artifactId>
+      <version>${project.parent.version}</version>
```
cescoffier, line 21: project.version is enough

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+     * 
+     * @param header the name of the header
+     * @param type the required type
+     * @param <T> the type
+     * @return the cast header value, which may be null if the header is not present
+     * @throws IllegalArgumentException if the header value could not be cast as required
+     */
+    @SuppressWarnings("unused")
+    @Nullable
+    public <T> T getHeader(final String header, final Class<T> type) {
```
cescoffier, line 70: The other methods use optional. Should we use optional here too?

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/OutgoingRabbitMQMetadata.java
```diff
@@ -0,0 +1,228 @@
...
+         * @param value the header value
+         * @return this {@link Builder}
+         */
+        @Fluent
+        public Builder withHeader(final String header, final Object value) {
+            headers.put(header, value);
+            return this;
+        }
+
+        @Fluent
```
cescoffier, line 123: No need for this annotation. It's something used in the Vert.x code generation.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java
```diff
@@ -0,0 +1,272 @@
...
+     * Whatever has been requested can be sent by the {@link Publisher} so only signal demand for what can be safely handled.
+     * <p>
+     * A {@link Publisher} can send less than is requested if the stream ends but
+     * then must emit either {@link Subscriber#onError(Throwable)} or {@link Subscriber#onComplete()}.
+     *
+     * @param l the strictly positive number of elements to requests to the upstream {@link Publisher}
+     */
+    @Override
+    public void request(long l) {
+        if (l != Long.MAX_VALUE) {
```
cescoffier, line 216: This is not correct.
You need to keep track of the number of requests, add l and if the sum is greater that Long.MAX cap it to Long.MAX. 
Long.MAX means "unlimited" so, you won't need to decrease or keep track of the requests anymore.

request the new value if it was not Long.MAX before.

Note that we may need a way for `inflight` to indicate no-backpressure (even if not recommended)

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ack/RabbitMQAckHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import io.vertx.mutiny.core.Context;
+
+/**
+ * Implemented to provide message acknowledgement strategies.
+ */
+public interface RabbitMQAckHandler {
+
+    enum Strategy {
+        /**
+         * Do no explicitly ack as it will be handled by RabbitMQ itself.
```
cescoffier, line 15: For my own curiosity, when is it nacked? On reception? If so that may introduce message loss.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java
cescoffier: Ok, I got it now.
So this method is expected to be called only `once`.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java
```diff
@@ -0,0 +1,272 @@
...
+                    } catch (Exception e) {
+                        // Message can't be sent - nacking and skipping.
+                        message.nack(e);
+                        RabbitMQLogging.log.serializationFailure(configuration.getChannel(), e);
+                        return Uni.createFrom().nullItem();
+                    }
+                })
+                .subscribe().with(
+                        tuple -> {
+                            if (tuple != null) { // Serialization issue
```
cescoffier, line 155: Why this comment?

cescoffier (changes requested):
Awesome! 

Just a few cosmetic things.
The next step would be to write some documentation (in the docs module).

wabrit:
> Awesome!
> 
> Just a few cosmetic things.
> The next step would be to write some documentation (in the docs module).

Shall I include that as part of this PR?

cescoffier:
Would be great if you can.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ack/RabbitMQAckHandler.java
wabrit: Yes in auto-ack mode the RabbitMQ broker regards successful delivery as an implied ACK; it provides faster throughput of messages but only in situations where the consumer is prepared to accept that some messages may be lost (e.g. you might use it for blog comments but not for bank transfers).

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java
wabrit: It's "inherited" from the identical code in the `AmqpCreditBasedSender.java` file from the AMQP connector; I don't know what the issue is, but I thought it prudent to include it just in case.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ack/RabbitMQAckHandler.java
cescoffier: That's a great example! 

This needs to be documented. Kafka has the same "issue" if you do anything async.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java
cescoffier: I believe it's a left-over, you can remove it.

cescoffier (approved):

cescoffier:
And.... MERGED! Thank you so much for this work!

PR #1322, opened 2021-07-16
Added Startup health check

### LINKED ISSUE #1300: Allows Kafka incoming connector to report readiness once assigned
This would be a new health check strategy for readiness where we wait until we get assigned to at least a topic/partition before indicating the application to be ready.

ozangunalp:
The incoming channel is ready after partition assignments. Fixes #1300

Here is a recap of proposed health checks

### Incoming channel (Kafka Consumer)
**Startup** 
Broker is accessible
- with Admin client: topic list returns with or without results
- with Metric: connection-count > 0 (Is this still relevant with startup ?)

**Readiness**
Assigned topic partitions for the channel is not empty
- with Consumer client: assigned partitions is not empty 
 NOTE: If the `partitions` config passed to the application is greater than the number of topic partitions on the broker, a consumer will end up with no partition assignments, and this will fail the readiness check of the application. 
- with Metric: connection-count > 0

**Liveness**
Check: No failures reported for channel

### Outgoing channel (Kafka Producer)
**Startup**
Broker is accessible
- with Admin client: topic list returns with or without results
- with Metric: connection-count > 0

**Readiness**
- with Admin client: topic exists
- with Metric: connection-count > 0 (Is this still relevant with startup ?)

**Liveness**
Check: No failures reported for channel

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
@@ -57,8 +57,11 @@ import io.vertx.mutiny.core.Vertx;
 @ConnectorAttribute(name = "topic", type = "string", direction = Direction.INCOMING_AND_OUTGOING, description = "The consumed / populated Kafka topic. If neither this property nor the `topics` properties are set, the channel name is used")
 @ConnectorAttribute(name = "health-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether health reporting is enabled (default) or disabled", defaultValue = "true")
 @ConnectorAttribute(name = "health-readiness-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether readiness health reporting is enabled (default) or disabled", defaultValue = "true")
-@ConnectorAttribute(name = "health-readiness-topic-verification", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether the readiness check should verify that topics exist on the broker. Default to false. Enabling it requires an admin connection.", defaultValue = "false")
-@ConnectorAttribute(name = "health-readiness-timeout", type = "long", direction = Direction.INCOMING_AND_OUTGOING, description = "During the readiness health check, the connector connects to the broker and retrieves the list of topics. This attribute specifies the maximum duration (in ms) for the retrieval. If exceeded, the channel is considered not-ready.", defaultValue = "2000")
+@ConnectorAttribute(name = "health-readiness-topic-verification", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether the readiness check should verify that topics exist on the broker. Default to false. Enabling it requires an admin connection. Deprecated: Use 'health-topic-verification-enabled' instead.", deprecated = true)
+@ConnectorAttribute(name = "health-readiness-timeout", type = "long", direction = Direction.INCOMING_AND_OUTGOING, description = "During the readiness health check, the connector connects to the broker and retrieves the list of topics. This attribute specifies the maximum duration (in ms) for the retrieval. If exceeded, the channel is considered not-ready. Deprecated: Use 'health-topic-verification-timeout' instead.", deprecated = true)
+@ConnectorAttribute(name = "health-topic-verification-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether the startup and readiness check should verify that topics exist on the broker. Default to false. Enabling it requires an admin client connection.", defaultValue = "false")
+@ConnectorAttribute(name = "health-topic-verification-timeout", type = "long", direction = Direction.INCOMING_AND_OUTGOING, description = "During the startup and readiness health check, the connector connects to the broker and retrieves the list of topics. This attribute specifies the maximum duration (in ms) for the retrieval. If exceeded, the channel is considered not-ready.", defaultValue = "2000")
+
 @ConnectorAttribute(name = "tracing-enabled", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Whether tracing is enabled (default) or disabled", defaultValue = "true")
 @ConnectorAttribute(name = "cloud-events", type = "boolean", direction = Direction.INCOMING_AND_OUTGOING, description = "Enables (default) or disables the Cloud Event support. If enabled on an _incoming_ channel, the connector analyzes the incoming records and try to create Cloud Event metadata. If enabled on an _outgoing_, the connector sends the outgoing messages as Cloud Event if the message includes Cloud Event Metadata.", defaultValue = "true")
 
@@ -167,6 +170,13 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
             c = merge(config, matching.get());
         }
         KafkaConnectorIncomingConfiguration ic = new KafkaConnectorIncomingConfiguration(c);
+        // log deprecated config
+        if (ic.getHealthReadinessTopicVerification().isPresent()) {
+            log.deprecatedConfig("health-readiness-topic-verification", "health-topic-verification-enabled");
+        }
+        if (ic.getHealthReadinessTimeout().isPresent()) {
+            log.deprecatedConfig("health-readiness-timeout", "health-topic-verification-timeout");
+        }
         int partitions = ic.getPartitions();
         if (partitions <= 0) {
             throw new IllegalArgumentException("`partitions` must be greater than 0");
@@ -224,6 +234,13 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
             c = merge(config, matching.get());
         }
         KafkaConnectorOutgoingConfiguration oc = new KafkaConnectorOutgoingConfiguration(c);
+        // log deprecated config
+        if (oc.getHealthReadinessTopicVerification().isPresent()) {
+            log.deprecatedConfig("health-readiness-topic-verification", "health-topic-verification-enabled");
+        }
+        if (oc.getHealthReadinessTimeout().isPresent()) {
+            log.deprecatedConfig("health-readiness-timeout", "health-topic-verification-timeout");
+        }
         KafkaSink sink = new KafkaSink(oc, kafkaCDIEvents);
         sinks.add(sink);
         return sink.getSink();
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
@@ -286,6 +286,24 @@ public class KafkaConnector implements IncomingConnectorFactory, OutgoingConnect
     }
 
     @Override
+    public HealthReport getStartup() {
+        HealthReport.HealthReportBuilder builder = HealthReport.builder();
+        if (sources.isEmpty() && sinks.isEmpty()) {
+            return builder.add("kafka-connector", false).build();
+        }
+
+        for (KafkaSource<?, ?> source : sources) {
+            source.isStarted(builder);
+        }
+
+        for (KafkaSink sink : sinks) {
+            sink.isStarted(builder);
+        }
+
+        return builder.build();
+    }
+
+    @Override
     public HealthReport getReadiness() {
         HealthReport.HealthReportBuilder builder = HealthReport.builder();
         if (sources.isEmpty() && sinks.isEmpty()) {
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/KafkaSinkHealth.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/KafkaSinkHealth.java
@@ -27,7 +27,7 @@ public class KafkaSinkHealth extends BaseHealth {
         this.topic = config.getTopic().orElse(config.getChannel());
         this.config = config;
 
-        if (config.getHealthReadinessTopicVerification()) {
+        if (config.getHealthReadinessTopicVerification().orElse(config.getHealthTopicVerificationEnabled())) {
             // Do not create the client if the readiness health checks are disabled
             Map<String, Object> adminConfiguration = new HashMap<>(kafkaConfiguration);
             this.admin = KafkaAdminHelper.createAdminClient(adminConfiguration, config.getChannel(), true);
@@ -60,8 +60,9 @@ public class KafkaSinkHealth extends BaseHealth {
     @Override
     protected void clientBasedStartupCheck(HealthReport.HealthReportBuilder builder) {
         try {
+            long timeout = config.getHealthReadinessTimeout().orElse(config.getHealthTopicVerificationTimeout());
             admin.listTopics()
-                    .await().atMost(Duration.ofMillis(config.getHealthReadinessTimeout()));
+                    .await().atMost(Duration.ofMillis(timeout));
             builder.add(channel, true);
         } catch (Exception failed) {
             builder.add(channel, false, "Failed to get response from broker for channel "
@@ -72,8 +73,9 @@ public class KafkaSinkHealth extends BaseHealth {
     protected void clientBasedReadinessCheck(HealthReport.HealthReportBuilder builder) {
         Set<String> topics;
         try {
+            long timeout = config.getHealthReadinessTimeout().orElse(config.getHealthTopicVerificationTimeout());
             topics = admin.listTopics()
-                    .await().atMost(Duration.ofMillis(config.getHealthReadinessTimeout()));
+                    .await().atMost(Duration.ofMillis(timeout));
             if (topics.contains(topic)) {
                 builder.add(channel, true);
             } else {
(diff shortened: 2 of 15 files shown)


smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/health/HealthCheckTest.java
```diff
@@ -66,45 +105,82 @@ public void testHealthOfApplicationWithoutOutgoingTopicUsingAdminCheck() {
...
+
+        Multi<Integer> channel = bean.getChannel();
+        channel
+                .select().first(10)
+                .collect().asList()
+                .await().atMost(Duration.ofSeconds(10));
+
+        // after subscription to channel
+
+        HealthReport startup = getHealth().getLiveness();
```
Ladicek, line 162: :-)

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/health/HealthCheckTest.java
```diff
@@ -114,11 +190,14 @@ public void testHealthOfApplicationWithChannel() {
                 .collect().asList()
                 .await().atMost(Duration.ofSeconds(10));
 
+        HealthReport startup = getHealth().getLiveness();
```
Ladicek, line 193: :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -330,12 +330,20 @@ public void isAlive(HealthReport.HealthReportBuilder builder) {
...
-        if (health != null) {
+        if (health != null && this.configuration.getHealthReadinessEnabled()) {
             health.isReady(builder);
         }
         // If health is disable do not add anything to the builder.
     }
 
+    public void isStarted(HealthReport.HealthReportBuilder builder) {
+        // This method must not be called from the event loop.
+        if (health != null) {
```
Ladicek, line 341: Should we also have a config for enabling/disabling startup health checks?

Ladicek: Hmm, probably no. We don't have one for liveness either. And the startup checks are rather basic, so we should be fine.

Ladicek:
I think this looks fine, barring the 2 typos in the test.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
ozangunalp: In terms of config I wasn't sure, that's why this PR is a draft. I have several pending questions : 
- Do we have a config for enable/disable startup-topic-verification ? 
- A separate config for startup topic verification timeout ? 
- Or do we deprecate/change the existing `health-readiness-topic-verification` flag and `health-readiness-timeout` config options to `health-topic-verification` and `health-check-timeout` ?
I'd vote for the third one.
@Ladicek let me know what do you think

Ladicek: Alright, let me try to recap all the probes and what we do. Because I'm a little lost personally :-)

### Startup

If a pod has a startup probe defined, Kubernetes will start with it and will ignore liveness and readiness (at the beginning). Once the startup probe returns "up", Kubernetes will start probing for liveness and readiness (and startup won't be checked anymore). If a pod doesn't have a startup probe, Kubernetes will start directly probing for liveness and readiness.

This is mostly useful for applications that take a long while to start, and during that time, the liveness and readiness handling doesn't make sense.

In our case, I'm not sure startup probe has a lot of meaning -- at the very most, it should just check that we have a connection to the broker. That's exactly what you're doing in this PR, so it's fine.

There could be a config option to disable startup probe, but if it only returns "down" when we're sure we didn't start yet, I think we can do without the config option.

### Liveness

If a pod has a liveness probe, Kubernetes will check it and restart the pod if liveness reports "down".

This is useful when we can tell that the application can't make progress. Restarting is usually a decent way to get out of such situation.

In our case, we know we can't make progress if a channel has failed -- because that is irreparable. That's what our liveness probe checks, so again, that's fine.

I don't think there should be a config option for disabling this. Channel failing is fatal, messages can't be processed anymore.

### Readiness

If a pod has a readiness probe, Kubernetes will check it and remove the pod from load balancer rotation if readiness reports "down".

This is useful when we can tell that the application won't be able to respond to users. Like, if all endpoints of the application require sending a message through Reactive Messaging, and we know sending such message is not gonna work, then it makes sense to mark the pod as not ready.

In our case, frankly, I'm not sure what's the value of readiness.

For a producer, we check:
- metric based: we have a connection to the broker;
- admin client based: we have a connection to the broker and the topic (to which we produce) exists on the broker.

For a consumer, we check:
- metric based: we have a connection to the broker;
- admin client based: we have a connection to the broker and we have at least 1 partition (of the topic we consume) assigned to ourselves.

I may be able to understand why we report readiness for producers, but really not sure about consumers. Especially if the number of partitions may easily be less than the number of applications in the same consumer group.

If anything, I can totally see why there must be a config option to disable readiness.

Sorry, this probably isn't very helpful, but it's my current opinion :-)

Ladicek: That said, I can see that with this PR, the readiness config would affect the startup probe too, and that's not nice. So we should do something about that for sure.

ozangunalp: :) nevertheless it is a good recap. 
+1 for what you wrote on startup and liveness checks.
For readiness, I think there is a use case where one would like to report a pod as "not ready" if it doesn't have any partition assignments, so it won't be consuming any data and it may be that one does not want to accept traffic to that pod.
As you said it makes sense that we have an option to disable the readiness check altogether.

ozangunalp: > That said, I can see that with this PR, the readiness config would affect the startup probe too, and that's not nice. So we should do something about that for sure.

Yes my doubts were more for the startup probe config:
- With this PR, startup is enabled whether the readiness check is enabled or not.
- But the admin client based check is enabled by the same config `health-readiness-topic-verification` – for the draft :)
- Idem for the `health-readiness-timeout`.

I suggest that we rename those config to `health-check-topic-verification` and `health-check-timeout`. 
With the admin based check, we do essentially the same call to the broker for startup and readiness. I think it makes sense to have the same config enabling the topic verification and configure the timeout for the call.

The other option is to have separate options for startup as `health-startup-topic-verification` and `health-startup-timeout`.

Ladicek: Yea I think I agree. With metric based health, startup and readiness are identical. With admin client based health, startup is a subset of readiness, which makes sense I guess.  I don't think there's a good reason to have one of them metric based and the other admin client based.

So +1 to the rename you propose.

(Alternatively, we could rename them to `health-admin-client-enabled` and `health-admin-client-timeout`. The term "topic verification" is nicely descriptive, but doesn't embody the fact that admin client is required, which is IMHO important.)

Not sure how to handle backwards compatibility though :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
```diff
@@ -217,4 +217,8 @@
     @LogMessage(level = Logger.Level.ERROR)
     @Message(id = 18249, value = "Unable to recover from the deserialization failure (topic: %s), configure a DeserializationFailureHandler to recover from errors.")
     void unableToDeserializeMessage(String topic, @Cause Throwable t);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 18250, value = "The configuration property '%s' is deprecated and is replaced by '%s'.")
+    void deprecatedConfig(String deprecated, String replace);
```
Ladicek, line 223: I was thinking maybe we could add `@Once` here, because if you have multiple Kafka channels, you will get the warning multiple times. But with `@Once`, we would only get the first warning, and the warnings about other deprecated properties would be lost. So probably best to keep it as is.

Ladicek (dismissed):
I'm adding a few more comments, but nothing blocking IMHO.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
ozangunalp: Nice, I didn't know about `@Once`. It can come in handy in the future. Let's keep it as is.

cescoffier:
> Incoming channel (Kafka Consumer)
> Startup
> Broker is accessible

> with Admin client: topic list returns with or without results
> with Metric: connection-count > 0 (Is this still relevant with startup ?)

Not sure in this case. 
Note that there is the case of lazy subscriptions too.
However, we need a "compatibility path"

> Readiness
> Assigned topic partitions for the channel is not empty

> with Consumer client: assigned partitions is not empty
> NOTE: If the partitions config passed to the application is greater than the number of topic partitions on the broker, a > consumer will end up with no partition assignments, and this will fail the readiness check of the application.
> with Metric: connection-count > 0

It's fine, you should not send traffic to this pod. 

> Liveness
> Check: No failures reported for channel

Fatal errors only (I can't remember if we have the distinction yet)

> Outgoing channel (Kafka Producer)
> Startup
> Broker is accessible
> 
> with Admin client: topic list returns with or without results
> with Metric: connection-count > 0

> Readiness
> 
> with Admin client: topic exists
> with Metric: connection-count > 0 (Is this still relevant with startup ?)

Not sure in this case.
However, we need a "compatibility path"

> Liveness
> Check: No failures reported for channel

Fatal error only.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ConfigurationCleaner.java
```diff
@@ -56,6 +59,7 @@
             "consumer-rebalance-listener.name",
             "key-deserialization-failure-handler",
             "value-deserialization-failure-handler",
+            "graceful-shutdown",
```
cescoffier, line 62: Oh forgot to add that one!

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
cescoffier: Yes, I'm not sure "readiness" is very useful for the consumer side. 

For the producer yes. For the producer something I was thinking is the following: it should report _not ready_ if we start having issues/lag when writing to Kafka. There is a "request" protocol to determine the number of inflight records. If we reach the max, it would make sense to mark the pod as non-ready. Such behavior should be configurable. 

For consumer, it does not affect the traffic.

cescoffier: About `alias`, yes, it's not for renaming but for global "shorter" properties.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
```diff
@@ -37,6 +37,13 @@ public HealthReport getLiveness() {
         return builder.build();
     }
 
+    public HealthReport getStartup() {
+        HealthReport.HealthReportBuilder builder = HealthReport.builder();
+        reporters.forEach(r -> r.getStartup().getChannels().forEach(builder::add));
+        failures.forEach(rf -> builder.add(rf.source, false, rf.failure.getMessage()));
```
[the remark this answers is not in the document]
cescoffier: Yes

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -285,6 +302,24 @@ public ConfigValue getConfigValue(String propertyName) {
         };
     }
 
+    @Override
+    public HealthReport getStartup() {
+        HealthReport.HealthReportBuilder builder = HealthReport.builder();
+        if (sources.isEmpty() && sinks.isEmpty()) {
+            return builder.add("kafka-connector", false).build();
```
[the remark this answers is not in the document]
ozangunalp: I think it still makes sense to have that to false. `sinks` and `sources` are added after the connector is created.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/KafkaSinkHealth.java
```diff
@@ -14,20 +14,20 @@
 import io.smallrye.reactive.messaging.kafka.KafkaConnectorOutgoingConfiguration;
 import io.smallrye.reactive.messaging.kafka.impl.KafkaAdminHelper;
 
-public class KafkaSinkReadinessHealth extends BaseHealth {
+public class KafkaSinkHealth extends BaseHealth {
```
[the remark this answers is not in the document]
ozangunalp: 👍  in a future PR :)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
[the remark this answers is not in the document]
Ladicek: Ah in that case, `false` is probably the right choice.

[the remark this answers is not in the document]
Ladicek: I mean, I'm looking at it and you're right. Sources and sinks are created when calling `IncomingConnectorFactory.getPublisherBuilder` / `OutgoingConnectorFactory.getSubscriberBuilder`. These methods are called from `ConfiguredChannelFactory.register`, which is called from `ConfiguredChannelFactory.initialize`, which is called from `MediatorManager.start`. In the CDI Portable Extension, that is called in an `AfterDeploymentValidation` observer, while in Quarkus, it is called in an `StartupEvent` observer. That is, it's early in the application lifecycle, but health probes can already be called at that time.

The only problem I can see is if we somehow instantiated a connector that doesn't have any channel configured.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/HealthCenter.java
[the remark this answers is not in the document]
ozangunalp: Indeed, I wrongly assumed that failures in this level can occur on startup. They are all caught at message consumption.

ozangunalp:
> Note that there is the case of lazy subscriptions too.\

I handled this case. It now works as expected: If no subscription is made startup and readiness checks report true.

> with Metric: connection-count > 0 (Is this still relevant with startup ?)

I think it still is. As @Ladicek very aptly resumed, startup probe is optional. When an app disables the topic verification connection count is still a viable alternative for the readiness check.

> Fatal errors only

We have three places where failures are gathered for reporting health:
- On the mediator level when creating the subscriber for message consumption – contribute to liveness & readiness
- On the connector level for message ingestion, reported from deserialization, commit strategy, dead letter queue, consumer client etc. – contribute only to liveness
- On the connector level for message dispatch – contribute only to liveness

Only the one for message ingestion makes the distinction for fatal or not, and when fatal closed the client. So there is no turning back from that.
As far as I understand on fatal errors we are closing the client. So maybe a better way is to check for liveness and readiness that producer and consumer clients are not closed?

cescoffier (approved):

cescoffier:
@Ladicek should we merge?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/base/WeldTestBase.java
```diff
@@ -119,10 +119,15 @@ public HealthCenter getHealth() {
...
+        return getHealth().getStartup().isOk();
+    }
+
     public boolean isReady() {
         System.out
-                .println(getHealth().getReadiness().getChannels().stream().map(ci -> ci.getChannel() + " " + ci.isOk()).collect(
-                        Collectors.joining()));
+                .println(getHealth().getReadiness().getChannels().stream()
+                        .map(ci -> ci.getChannel() + " " + ci.isOk() + " " + ci.getMessage())
+                        .collect(Collectors.joining()));
```
Ladicek, line 130: At some point, we should probably remove this? :-)

Ladicek (approved):

Ladicek:
Yes, this is good to merge.

cescoffier:
We will need to do the "wiring" in the Quarkus extension.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/base/WeldTestBase.java
ozangunalp: It was very useful during tests :)

Ladicek: I'm sure it is, but I'm also pretty sure it is annoying otherwise :-) We could make it a trace log.

PR #1330, opened 2021-07-22
Avoid using the size() method on a blocking linked queue

cescoffier:
Instead, we track the size of the queue in the subscriber directly.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
@@ -4,11 +4,11 @@ import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
 import java.time.Duration;
 import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.reactivestreams.Subscription;
@@ -66,17 +66,8 @@ public class KafkaRecordStream<K, V> extends AbstractMulti<ConsumerRecord<K, V>>
          */
         private final Uni<ConsumerRecords<K, V>> pollUni;
 
-        private final Queue<ConsumerRecord<K, V>> queue;
+        private final RecordQueue<ConsumerRecord<K, V>> queue;
         private final long retries;
-
-        /**
-         * Store the size of the queue.
-         * Calling {@code .size()} on the queue traverses the queue, which is slow.
-         * So, we maintain the size:
-         * - add 1 for each poll records (from the kafka polling thread)
-         * - subtract the emitted records (from the Vert.x context)
-         */
-        private final AtomicLong size;
 
         /**
          * {@code true} if the subscription has been cancelled.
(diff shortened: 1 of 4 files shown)


loicmathieu:
If you can provide a SNAPSHOT version with this change we may be able to test it on the original application that discover this performance issue. 
Unfortunatly, we need to build the application inside the CI to be able to deploy it on an environment where we can load test it so building the branch seems complicated.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/perf/EndToEndPerfTest.java
```diff
@@ -0,0 +1,101 @@
...
+        consumer.subscribe(Collections.singletonList(output_topic));
+        boolean done = false;
+        long received = 0;
+        while (!done) {
+            ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(1000));
+            received = received + records.count();
+            if (received == COUNT) {
+                done = true;
+            }
+        }
```
ozangunalp, line 98: We could have replaced this consumer with `usage.consumeStrings`

ozangunalp (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
```diff
@@ -82,14 +90,18 @@ public KafkaRecordStreamSubscription(
...
             this.pollUni = client.poll()
                     .onItem().transform(cr -> {
                         if (cr.isEmpty()) {
                             return null;
                         }
-                        cr.forEach(queue::offer);
+                        for (ConsumerRecord<K, V> rec : cr) {
+                            queue.offer(rec);
+                        }
+                        size.addAndGet(cr.count());
```
Ladicek, line 104: I don't truly understand the code, so please treat everything I write here as nothing more than a comment... but this seems rather racy. You could for example easily end up with a negative `size`, which would be unexpected. But the code doesn't seem to care. And, I mean, it was racy before, so it's probably fine :-)

If `ConsumerRecords` was a `Collection` (it's `Iterable` :-/), we could `queue.addAll(cr)`, which has a single linearization point and returns shortly after it, and the race window would be much smaller.

(Interestingly, `ConcurrentLinkedDeque.addAll` could easily take an `Iterable`, because it doesn't use anything else from the ` Collection` API. But, of course, Java doesn't support contravariant parameters.)

Alternatively, we could increment after each `queue.offer`, which again makes the race window a lot smaller (at the cost of the `AtomicLong` overhead).

cescoffier: Yes, there is a race. Unfortunately, there is no good solution except re-implementing a Queue. I will try this.

cescoffier: Just pushed a different approach using a specialized queue.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/RecordQueue.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.util.ArrayDeque;
+import java.util.Collection;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+/**
+ * Stores the records coming from Kafka.
+ * Only two operations are allowed ({@link #addAll(ConsumerRecords)}, {@link #clear()} and {@link #poll()}.
```
Ladicek, line 11:
```suggestion
 * Only these operations are allowed: {@link #addAll(ConsumerRecords)}, {@link #clear()}, {@link #size()} and {@link #poll()}.
```

ozangunalp (approved):

PR #1332, opened 2021-07-27
Add `messaging.consumer_id` to span attribute for Kafka Consumer

kenfinnigan:
I wanted to do the same for AMQP, but I couldn't find where I can get the Client ID from.

If someone knows, I can add to this PR


Ladicek:
I assume this is related to https://github.com/open-telemetry/opentelemetry-specification/pull/1810 ? Would be good to have it in a comment.

As for AMQP, I don't know what AMQP concept would be a good fit. Perhaps the container ID, but it seems we set that to `null` if the user doesn't configure it, so it's probably almost always empty.

Ladicek:
One thing we could add relatively easily for AMQP is a correlation ID, though :-)

Ladicek (approved):

kenfinnigan:
Yes, that's the spec change to add `consumer_id`.

I thought there was an AMQP Client Id, but maybe I was wrong

cescoffier (approved):

PR #1333, opened 2021-07-28
Kafka connector option for consuming messages in batch

### LINKED ISSUE #1214: Allow receiving Kafka record by batch
At the moment, the records are received one by one, requiring a method processing the stream to recreate a _batch_ (by size). It would be convenient to allow the user to receive the batch directly:

```java
@Incoming("kafka")
void consumeBatch(List<Record> records) {
  // contains all the records from the batch
}
```

The underlying client should allow us to do this, and a converter should be able to call the method.

Acknowledgment would still be the same, except that when this method returns, we ack all the included messages.

ozangunalp:
Activated with `mp.messaging.incoming.$channel.batch=true`.

Allows incoming channels to consume messages in batch with following types: 
- `List<Payload>`
- `Message<List<Payload>>`
- `KafkaRecordBatch<Key, Payload>`
- `ConsumerRecords<Key, Payload>`

On `ack` applies the configured commit strategy for records with the last offset per partition.
On `nack` applies the configured failure strategy for all messages inside the record batch.

Adds `max-queue-size-factor` configuration for adjusting the max queue size for records waiting to be processed, using `max.poll.records` * `max-queue-size-factor`, for example by default a maximum of 500 * 2 = 1000 records will be queued.
In batch mode `max.poll.records` is considered `1` for calculating the max number of queued batch records (`ConsumerRecords`).

Closes #1214

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -149,11 +149,11 @@ Type: _int_ | false | `1000`
 
 Type: _boolean_ | false | `true`
 
-| *batch* | Whether the Kafka records are consumed in batch
+| [.no-hyphens]#*batch*# | Whether the Kafka records are consumed in batch. The channel injection point must consume a compatible type, such as `List<Payload>` or `KafkaRecordBatch<Payload>`.
 
 Type: _boolean_ | false | `false`
 
-| *max-queue-size-factor* | Multiplier factor to determine maximum number of records queued for processing, using `max.poll.records` * `max-queue-size-factor`. Defaults to 2. In `batch` mode `max.poll.records` is considered `1`.
+| [.no-hyphens]#*max-queue-size-factor*# | Multiplier factor to determine maximum number of records queued for processing, using `max.poll.records` * `max-queue-size-factor`. Defaults to 2. In `batch` mode `max.poll.records` is considered `1`.
 
 Type: _int_ | false | `2`
 
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -149,4 +149,12 @@ Type: _int_ | false | `1000`
 
 Type: _boolean_ | false | `true`
 
+| *batch* | Whether the Kafka records are consumed in batch
+
+Type: _boolean_ | false | `false`
+
+| *max-queue-size-factor* | Multiplier factor to determine maximum number of records queued for processing, using `max.poll.records` * `max-queue-size-factor`. Defaults to 2. In `batch` mode `max.poll.records` is considered `1`.
+
+Type: _int_ | false | `2`
+
 |===
--- a/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
+++ b/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
@@ -1,5 +1,6 @@
 package inbound;
 
+import java.time.Instant;
 import java.util.concurrent.CompletionStage;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -11,6 +12,7 @@ import org.eclipse.microprofile.reactive.messaging.Incoming;
 
 import io.smallrye.reactive.messaging.kafka.KafkaRecordBatch;
 import io.smallrye.reactive.messaging.kafka.KafkaRecord;
+import io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordBatchMetadata;
 import io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordMetadata;
 
 @ApplicationScoped
@@ -24,10 +26,11 @@ public class KafkaRecordBatchExample {
             record.getMetadata(IncomingKafkaRecordMetadata.class).ifPresent(metadata -> {
                 int partition = metadata.getPartition();
                 long offset = metadata.getOffset();
-                ConsumerRecord<String, Double> consumerRecord = metadata.getRecord();
+                Instant timestamp = metadata.getTimestamp();
                 //... process messages
             });
         }
+        // ack will commit the latest offsets (per partition) of the batch.
         return records.ack();
     }
 
--- a/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
+++ b/documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
@@ -1,0 +1,44 @@
+package inbound;
+
+import java.util.concurrent.CompletionStage;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
+import org.eclipse.microprofile.reactive.messaging.Incoming;
+
+import io.smallrye.reactive.messaging.kafka.KafkaRecordBatch;
+import io.smallrye.reactive.messaging.kafka.KafkaRecord;
+import io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordMetadata;
+
+@ApplicationScoped
+public class KafkaRecordBatchExample {
+
+    @SuppressWarnings("unchecked")
+    // tag::code[]
+    @Incoming("prices")
+    public CompletionStage<Void> consumeMessage(KafkaRecordBatch<String, Double> records) {
+        for (KafkaRecord<String, Double> record : records) {
+            record.getMetadata(IncomingKafkaRecordMetadata.class).ifPresent(metadata -> {
+                int partition = metadata.getPartition();
+                long offset = metadata.getOffset();
+                ConsumerRecord<String, Double> consumerRecord = metadata.getRecord();
+                //... process messages
+            });
+        }
+        return records.ack();
+    }
+
+    @Incoming("prices")
+    public void consumeRecords(ConsumerRecords<String, Double> records) {
+        for (TopicPartition partition : records.partitions()) {
+            for (ConsumerRecord<String, Double> record : records.records(partition)) {
+                //... process messages
+            }
+        }
+    }
+    // end::code[]
+
+}
--- a/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
@@ -280,7 +280,7 @@ include::consumer-rebalance-listener.adoc[]
 By default, incoming methods receive each Kafka record individually.
 Under the hood, Kafka consumer clients poll the broker constantly and receive records in batches, presented inside the `ConsumerRecords` container.
 
-In *batch* mode, your application can receive all the records returned by the consumer *poll*.
+In *batch* mode, your application can receive all the records returned by the consumer *poll* in one go.
 
 To achieve this you need to set `mp.messaging.incoming.$channel.batch=true` *and* specify a compatible container type to receive all the data:
 
@@ -289,16 +289,16 @@ To achieve this you need to set `mp.messaging.incoming.$channel.batch=true` *and
 include::example$inbound/KafkaRecordBatchPayloadExample.java[tags=code]
 ----
 
-The incoming method can also receive `Message<List<Payload>`, `KafkaBatchRecords<Payload>` `ConsumerRecords<Key, Payload>` types. These types give access :
+The incoming method can also receive `Message<List<Payload>`, `KafkaBatchRecords<Payload>` `ConsumerRecords<Key, Payload>` types, They give access to record details such as offset or timestamp :
 
 [source, java, indent=0]
 ----
 include::example$inbound/KafkaRecordBatchExample.java[tags=code]
 ----
 
-Note that the successful processing of the incoming batch record will commit the latest offsets for each partition received inside the batch. The configured commit strategy will be applied for these records only.
+Note that the successful processing of the incoming record batch will commit the latest offsets for each partition received inside the batch. The configured commit strategy will be applied for these records only.
 
-Conversely, if the processing error throws an exception, all messages are _nacked_, applying the failure strategy for all the records inside the batch.
+Conversely, if the processing throws an exception, all messages are _nacked_, applying the failure strategy for all the records inside the batch.
 
 === Configuration Reference
 
--- a/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
@@ -275,6 +275,31 @@ The `partitionkey` attribute is mapped to the record's key if any.
 
 include::consumer-rebalance-listener.adoc[]
 
+=== Receiving Kafka Records in Batches
+
+By default, incoming methods receive each Kafka record individually.
+Under the hood, Kafka consumer clients poll the broker constantly and receive records in batches, presented inside the `ConsumerRecords` container.
+
+In *batch* mode, your application can receive all the records returned by the consumer *poll*.
+
+To achieve this you need to set `mp.messaging.incoming.$channel.batch=true` *and* specify a compatible container type to receive all the data:
+
+[source, java, indent=0]
+----
+include::example$inbound/KafkaRecordBatchPayloadExample.java[tags=code]
+----
+
+The incoming method can also receive `Message<List<Payload>`, `KafkaBatchRecords<Payload>` `ConsumerRecords<Key, Payload>` types. These types give access :
+
+[source, java, indent=0]
+----
+include::example$inbound/KafkaRecordBatchExample.java[tags=code]
+----
+
+Note that the successful processing of the incoming batch record will commit the latest offsets for each partition received inside the batch. The configured commit strategy will be applied for these records only.
+
+Conversely, if the processing error throws an exception, all messages are _nacked_, applying the failure strategy for all the records inside the batch.
+
 === Configuration Reference
 
 include::connectors:partial$META-INF/connector/smallrye-kafka-incoming.adoc[]
--- a/smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
+++ b/smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
@@ -1,0 +1,46 @@
+package io.smallrye.reactive.messaging.kafka.api;
+
+import java.util.Set;
+
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
+
+/**
+ * Contains information about the batch of messages received from a channel backed by Kafka.
+ * Encapsulates underlying Kafka {@link ConsumerRecords} received from the consumer client.
+ *
+ * As this is an incoming message metadata it is created by the framework and injected into incoming batch messages.
+ *
+ * @param <K> The record key type
+ * @param <T> The record payload type
+ */
+public class IncomingKafkaRecordBatchMetadata<K, T> {
+
+    private final ConsumerRecords<K, T> records;
+
+    public IncomingKafkaRecordBatchMetadata(ConsumerRecords<K, T> records) {
+        this.records = records;
+    }
+
+    /**
+     * @return the underlying Kafka {@link ConsumerRecords}
+     */
+    public ConsumerRecords<K, T> getRecords() {
+        return records;
+    }
+
+    /**
+     * @return the number of records for all topics
+     */
+    public int count() {
+        return records.count();
+    }
+
+    /**
+     * @return the set of topic partitions with data in this batch record
+     */
+    public Set<TopicPartition> partitions() {
+        return records.partitions();
+    }
+
+}
--- a/smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
+++ b/smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
@@ -30,14 +30,14 @@ public class IncomingKafkaRecordBatchMetadata<K, T> {
     }
 
     /**
-     * @return the number of records for all topics
+     * @return the total number of records for all topic partitions
      */
     public int count() {
         return records.count();
     }
 
     /**
-     * @return the set of topic partitions with data in this batch record
+     * @return the set of topic partitions with data in this record batch
      */
     public Set<TopicPartition> partitions() {
         return records.partitions();
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
@@ -32,16 +32,11 @@ public class IncomingKafkaRecordBatch<K, T> implements KafkaRecordBatch<K, T> {
         List<IncomingKafkaRecord<K, T>> incomingRecords = new ArrayList<>();
         Map<TopicPartition, IncomingKafkaRecord<K, T>> latestOffsetRecords = new HashMap<>();
         for (TopicPartition partition : records.partitions()) {
-            IncomingKafkaRecord<K, T> latestOffsetRecord = null;
             for (ConsumerRecord<K, T> record : records.records(partition)) {
                 IncomingKafkaRecord<K, T> rec = new IncomingKafkaRecord<>(record, commitHandler, onNack,
                         cloudEventEnabled, tracingEnabled);
                 incomingRecords.add(rec);
                 latestOffsetRecords.put(partition, rec);
-                latestOffsetRecord = rec;
-            }
-            if (latestOffsetRecord != null) {
-                commitHandler.received(latestOffsetRecord);
             }
         }
         this.incomingRecords = Collections.unmodifiableList(incomingRecords);
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
@@ -1,0 +1,104 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
+import org.eclipse.microprofile.reactive.messaging.Metadata;
+
+import io.smallrye.mutiny.Multi;
+import io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordBatchMetadata;
+import io.smallrye.reactive.messaging.kafka.commit.KafkaCommitHandler;
+import io.smallrye.reactive.messaging.kafka.fault.KafkaFailureHandler;
+
+public class IncomingKafkaRecordBatch<K, T> implements KafkaRecordBatch<K, T> {
+
+    private final Metadata metadata;
+    private final List<KafkaRecord<K, T>> incomingRecords;
+    private final Map<TopicPartition, KafkaRecord<K, T>> latestOffsetRecords;
+
+    public IncomingKafkaRecordBatch(ConsumerRecords<K, T> records, KafkaCommitHandler commitHandler,
+            KafkaFailureHandler onNack, boolean cloudEventEnabled, boolean tracingEnabled) {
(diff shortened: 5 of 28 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaBatchRecords.java
```diff
@@ -0,0 +1,106 @@
...
+            KafkaFailureHandler onNack, boolean cloudEventEnabled, boolean tracingEnabled) {
+        List<IncomingKafkaRecord<K, T>> incomingRecords = new ArrayList<>();
+        Map<TopicPartition, IncomingKafkaRecord<K, T>> latestOffsetRecords = new HashMap<>();
+        for (TopicPartition partition : records.partitions()) {
+            IncomingKafkaRecord<K, T> latestOffsetRecord = null;
+            for (ConsumerRecord<K, T> record : records.records(partition)) {
+                IncomingKafkaRecord<K, T> rec = new IncomingKafkaRecord<>(record, commitHandler, onNack,
+                        cloudEventEnabled, tracingEnabled);
+                incomingRecords.add(rec);
+                latestOffsetRecord = latestOffsetRecords.compute(partition,
```
Ladicek, line 40: Just out of curiosity, wouldn't `latestOffsetRecords.put(partition, rec)` work as well? Records are consumed in sequence, so unconditionally putting each record into the map ends up with having the last record there, for each partition?

Ladicek: (And of course unconditionally setting `latestOffsetRecord = rec`.)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaBatchRecords.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public CompletionStage<Void> nack(Throwable reason, Metadata metadata) {
+        return Uni.combine().all().unis(this.incomingRecords.stream()
+                .map(record -> Uni.createFrom().completionStage(record.nack(reason, metadata)))
+                .collect(Collectors.toList()))
+                .collectFailures()
+                .discardItems()
+                .convert().toCompletionStage();
+    }
```
Ladicek, line 105: For `ack`, we only use the latest record for each partition, so that's fine, but here, we `nack` each record, so maybe order matters? (Not sure how `Uni.combine().all(...)` handles ordering, just thinking out loud.)

ozangunalp: I asked myself the same question. But I assume that the zip operation will execute sequentially on `Uni`.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaBatchRecords.java
ozangunalp: Yes, I may have over-thought this, or not enough :) 
Not having an ascending order inside a partition of `ConsumerRecords` would be very problematic.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaBatchRecords.java
ozangunalp: Took another look at this. `Uni.combine.all(...)` isn't really a zip implementation. What we need here is `Multi.createBy().concatenating().streams(...)`.

ozangunalp: I changed both `nack` and `ack` to Multi concatenate for the sake of simplicity.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
```diff
@@ -145,4 +145,8 @@ Type: _int_ | false | `1000`
 
 Type: _boolean_ | false | `true`
 
+| *batch* | Whether the Kafka records are consumed in batch
+
+Type: _boolean_ | false | `false`
```
Ladicek, line 150: Is that something we could derive from the `@Incoming` method parameter type (or `@Channel` injection point type)? Maybe only in Quarkus, like we autodetect the [de]serializer.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaBatchRecordExample.java
```diff
@@ -0,0 +1,46 @@
...
+    public CompletionStage<Void> consumeMessage(KafkaBatchRecords<String, Double> records) {
+        for (KafkaRecord<String, Double> record : records) {
+            IncomingKafkaRecordMetadata<String, Double> metadata = record.getMetadata(IncomingKafkaRecordMetadata.class).orElse(null);
+            if (metadata != null) {
+                int partition = metadata.getPartition();
+                long offset = metadata.getOffset();
+                ConsumerRecord<String, Double> consumerRecord = metadata.getRecord();
+            }
+        }
+        return CompletableFuture.completedFuture(null);
```
Ladicek, line 32: This is interesting. For a single message, we would `return msg.ack()`. How do we propagate ack here? Does `KafkaBatchRecords` have an `ack` method?

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaBatchRecordExample.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Incoming("prices")
+    public CompletionStage<Void> consumeRecords(ConsumerRecords<String, Double> records) {
+        for (TopicPartition partition : records.partitions()) {
+            for (ConsumerRecord<String, Double> record : records.records(partition)) {
+
+            }
+        }
+        return CompletableFuture.completedFuture(null);
```
Ladicek, line 42: Same here, how do we `ack`?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaBatchRecords.java
Ladicek: +1, I think this makes sense now.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaBatchRecords.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Represents a batch of Kafka records received by polling the {@link org.apache.kafka.clients.consumer.KafkaConsumer}
+ *
+ * This type extends the {@code Message<List<T>>} where {@code T} is the type of records' payloads.
+ * The complete list of Kafka record payloads are accessible via the {@link Message#getPayload()} method.
+ *
+ * @param <K> The record key type
+ * @param <T> The record payload type
+ */
+public interface KafkaBatchRecords<K, T> extends Message<List<T>>, Iterable<KafkaRecord<K, T>> {
```
Ladicek, line 18: For symmetry with Kafka's `ConsumerRecord`/`ConsumerRecords`, should we perhaps call this `KafkaRecords`?

Ladicek: Alternatively, `KafkaRecordBatch`.

`KafkaBatchRecords` seems disconnected from `KafkaRecord`.

Yes, I enjoy naming discussions :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
```diff
@@ -173,6 +175,20 @@ public void setRebalanceListener() {
...
+                .onItem().transformToMulti(v -> batchStream);
+    }
+
+    public Multi<ConsumerRecords<K, V>> subscribeBatch(Pattern topics) {
+        return runOnPollingThread(c -> {
+            c.subscribe(topics, rebalanceListener);
+        })
+                .onItem().transformToMulti(v -> batchStream);
+    }
+
```
Ladicek, line 191: This got me thinking: we don't do anything to prevent multiple subscriptions, do we? (Or multiple subscriptions/batchSubscriptions, that's the same thing.) If the user calls `subscribe` (or `subscribeBatch`) multiple times, they always get the same stream back. Which sounds weird.

Ladicek:
Sorry it took me so long to get back to this. I think it's OK, though I'm not sure how acking is supposed to work in the examples.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
ozangunalp: There will be another PR on Quarkus to adapt the serde autodetect. I was thinking more of a suggestion message rather than automatically setting a channel to batch.

A note here: an incoming channel can not have two consuming `@Incoming`/`@Channel` that are batch and not batch in the same time.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaBatchRecordExample.java
ozangunalp: yes in this example the `return records.ack()` would be correct.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaBatchRecordExample.java
ozangunalp: I need to look more into this case, as `ConsumerRecords` is injected as part of a converter. I assume cases with current converters ack the message with post processing by default.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaBatchRecords.java
ozangunalp: I was actually going to rename this to `KafkaRecordBatch` :) But I knew this discussion was coming. 
Kafka client internals (pre-deserialization) use `Record` and `RecordBatch`.
Lets go for `KafkaRecordBatch` ?

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
Ladicek: Right! Hence I thought that this can be unambiguously detected.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaBatchRecordExample.java
Ladicek: I'd guess that this method should just return `void` then?

Or it could accept `Message<ConsumerRecords<String, Double>>` and `return msg.ack()` maybe. I didn't really think it through, maybe there's a problem with that.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaBatchRecords.java
Ladicek: +1

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
ozangunalp: Now that I spent more time on `KafkaRecordStream`, the fact that it returns the same stream doesn't surprise me. 
I think we can turn `subscribe` methods into package private as they are intended to be called only by the owning `KafkaSource`.

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @SuppressWarnings("unchecked")
+    // tag::code[]
+    @Incoming("prices")
+    public CompletionStage<Void> consumeMessage(KafkaRecordBatch<String, Double> records) {
+        for (KafkaRecord<String, Double> record : records) {
+            record.getMetadata(IncomingKafkaRecordMetadata.class).ifPresent(metadata -> {
+                int partition = metadata.getPartition();
+                long offset = metadata.getOffset();
+                ConsumerRecord<String, Double> consumerRecord = metadata.getRecord();
```
Ladicek, line 27: This is equivalent to existing `record`, right?

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
```diff
@@ -274,6 +274,31 @@ The `partitionkey` attribute is mapped to the record's key if any.
 
 include::consumer-rebalance-listener.adoc[]
 
+=== Receiving Kafka Records in Batches
+
+By default, incoming methods receive each Kafka record individually.
+Under the hood, Kafka consumer clients poll the broker constantly and receive records in batches, presented inside the `ConsumerRecords` container.
+
+In *batch* mode, your application can receive all the records returned by the consumer *poll*.
```
Ladicek, line 282: "[receive all the records returned by the consumer *poll*] at the same time", "in one go" or something like that?

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
```diff
@@ -274,6 +274,31 @@ The `partitionkey` attribute is mapped to the record's key if any.
...
+
+To achieve this you need to set `mp.messaging.incoming.$channel.batch=true` *and* specify a compatible container type to receive all the data:
+
+[source, java, indent=0]
+----
+include::example$inbound/KafkaRecordBatchPayloadExample.java[tags=code]
+----
+
+The incoming method can also receive `Message<List<Payload>`, `KafkaBatchRecords<Payload>` `ConsumerRecords<Key, Payload>` types. These types give access :
```
Ladicek, line 291: Give access to what?

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
```diff
@@ -274,6 +274,31 @@ The `partitionkey` attribute is mapped to the record's key if any.
...
+
+[source, java, indent=0]
+----
+include::example$inbound/KafkaRecordBatchExample.java[tags=code]
+----
+
+Note that the successful processing of the incoming batch record will commit the latest offsets for each partition received inside the batch. The configured commit strategy will be applied for these records only.
```
Ladicek, line 298: "successful processing of the incoming batch record" -> "successful processing of the incoming record batch"?

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
```diff
@@ -274,6 +274,31 @@ The `partitionkey` attribute is mapped to the record's key if any.
...
+
+Conversely, if the processing error throws an exception, all messages are _nacked_, applying the failure strategy for all the records inside the batch.
```
Ladicek, line 300: "if the processing error throws an exception" -> "if the processing throws an exception"?

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
Ladicek: Also, what's "these records"? (I know it. The reader doesn't :-) )

smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @return the underlying Kafka {@link ConsumerRecords}
+     */
+    public ConsumerRecords<K, T> getRecords() {
+        return records;
+    }
+
+    /**
+     * @return the number of records for all topics
```
Ladicek, line 33: "for all topics" -> "for all partitions"? Or just "the total number of records in this record batch"?

smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public int count() {
+        return records.count();
+    }
+
+    /**
+     * @return the set of topic partitions with data in this batch record
```
Ladicek, line 40: "in this batch record" -> "in this record batch"?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -90,6 +89,8 @@
 @ConnectorAttribute(name = "graceful-shutdown", type = "boolean", direction = Direction.INCOMING, description = "Whether or not a graceful shutdown should be attempted when the application terminates.", defaultValue = "true")
 @ConnectorAttribute(name = "poll-timeout", type = "int", direction = Direction.INCOMING, description = "The polling timeout in milliseconds. When polling records, the poll will wait at most that duration before returning records. Default is 1000ms", defaultValue = "1000")
 @ConnectorAttribute(name = "pause-if-no-requests", type = "boolean", direction = Direction.INCOMING, description = "Whether the polling must be paused when the application does not request items and resume when it does. This allows implementing back-pressure based on the application capacity. Note that polling is not stopped, but will not retrieve any records when paused.", defaultValue = "true")
+@ConnectorAttribute(name = "batch", type = "boolean", direction = Direction.INCOMING, description = "Whether the Kafka records are consumed in batch", defaultValue = "false")
```
Ladicek, line 92: Maybe the description should also say what kind of types must be used in the `@Incoming` method declaration?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaRecordBatch.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * This type extends the {@code Message<List<T>>} where {@code T} is the type of records' payloads.
+ * The complete list of Kafka record payloads are accessible via the {@link Message#getPayload()} method.
+ *
+ * @param <K> The record key type
+ * @param <T> The record payload type
+ */
+public interface KafkaRecordBatch<K, T> extends Message<List<T>>, Iterable<KafkaRecord<K, T>> {
+    /**
+     * @return list of records contained in this batch message
```
Ladicek, line 20: "in this batch message" -> "in this message batch"?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ConfigurationCleaner.java
```diff
@@ -49,6 +49,7 @@
             "retry-attempts",
             "retry-max-wait",
             "broadcast",
+            "batch",
```
Ladicek, line 52: Should `"max-queue-size-factor"` be also added? (I don't really know, genuine question :-) )

Ladicek:
The code seems fine (though I don't really understand tracing, so can't comment on that :-) ). I added some comments about documentation, which I think could be improved a bit.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ConfigurationCleaner.java
ozangunalp: if not kafka client creation will log unrecognised config property

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaRecordBatch.java
ozangunalp: it is a "message batch" and a "batch message" :)

smallrye-reactive-messaging-kafka-api/src/main/java/io/smallrye/reactive/messaging/kafka/api/IncomingKafkaRecordBatchMetadata.java
ozangunalp: I think the term "topic partitions" is the most appropriate one. I've shamelessly copied the description for `ConsumerRecords#count`.

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
ozangunalp: "these records" refer to the previously mentioned "latest offsets for each partition received inside the batch"

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
ozangunalp: yes `KafkaRecord` is the `Message` wrapper, `ConsumerRecord` is the Kafka client type.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
cescoffier: > A note here: an incoming channel can not have two consuming @Incoming/@Channel that are batch and not batch in the same time.

I would consider that invalid. The workaround would be to consume it using batch twice and having an application method unwrapping the records. 

About the detection in Quarkus: +1

documentation/src/main/doc/modules/kafka/examples/inbound/KafkaRecordBatchExample.java
```diff
@@ -0,0 +1,46 @@
...
+    public CompletionStage<Void> consumeMessage(KafkaRecordBatch<String, Double> records) {
+        for (KafkaRecord<String, Double> record : records) {
+            record.getMetadata(IncomingKafkaRecordMetadata.class).ifPresent(metadata -> {
+                int partition = metadata.getPartition();
+                long offset = metadata.getOffset();
+                Instant timestamp = metadata.getTimestamp();
+                //... process messages
+            });
+        }
+        return records.ack();
```
cescoffier, line 33: Would be probably good to add a comment here such as:

// ack will commit the latest offsets (per partition) of the batch.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
```diff
@@ -0,0 +1,104 @@
...
+        return this::nack;
+    }
+
+    @Override
+    public CompletionStage<Void> ack() {
+        return Multi.createBy().concatenating().collectFailures()
+                .streams(this.latestOffsetRecords.values().stream()
+                        .map(record -> Multi.createFrom().completionStage(record::ack))
+                        .collect(Collectors.toList()))
+                .toUni().convert().toCompletionStage();
```
cescoffier, line 93: should be subscribeAsCompletionStage() no?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public CompletionStage<Void> nack(Throwable reason, Metadata metadata) {
+        return Multi.createBy().concatenating().collectFailures()
+                .streams(this.incomingRecords.stream()
+                        .map(record -> Multi.createFrom().completionStage(() -> record.nack(reason, metadata)))
+                        .collect(Collectors.toList()))
+                .toUni().convert().toCompletionStage();
```
cescoffier, line 102: should be subscribeAsCompletionStage() no?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
```diff
@@ -0,0 +1,104 @@
...
+        return Multi.createBy().concatenating().collectFailures()
+                .streams(this.latestOffsetRecords.values().stream()
+                        .map(record -> Multi.createFrom().completionStage(record::ack))
+                        .collect(Collectors.toList()))
+                .toUni().convert().toCompletionStage();
+    }
+
+    @Override
+    public CompletionStage<Void> nack(Throwable reason, Metadata metadata) {
+        return Multi.createBy().concatenating().collectFailures()
```
cescoffier, line 98: should we concatenate or merge?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public Function<Throwable, CompletionStage<Void>> getNack() {
+        return this::nack;
+    }
+
+    @Override
+    public CompletionStage<Void> ack() {
+        return Multi.createBy().concatenating().collectFailures()
```
cescoffier, line 89: should we concatenate or merge?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -195,23 +196,34 @@ void init() {
...
-            streams.add(source.getStream());
+            if (!ic.getBatch()) {
+                streams.add(source.getStream());
+            } else {
+                streams.add(source.getBatchStream());
+            }
         }
 
-        Multi<IncomingKafkaRecord<Object, Object>> multi = Multi.createBy().merging().streams(streams);
+        // TODO
```
cescoffier, line 225: ??

cescoffier (changes requested):
Looks great! 

What happen with tracing? You should discuss with @kenfinnigan about that topic.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
ozangunalp: it is the same thing under the hood

ozangunalp: no?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
ozangunalp: I went for concatenate to preserve the ordering.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/IncomingKafkaRecordBatch.java
cescoffier: Yes, but the actual implementation is in subscribeAsCompletionStage() to it avoids instantiating the `convert` group.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
ozangunalp: I forgot to complete this TODO. I would suggest a `MultiMerge#streams(List<Publisher>)` in Mutiny.

kenfinnigan:
@ozangunalp Looks ok, though I wonder whether there's a way to parallelize the batch span creation instead of a loop to reduce the performance impact?

Agree that `span.makeCurrent()` is odd in the batch context, as it will make whatever was the last created span as the parent for the processing method, though be fine for outgoing message spans.

OpenTelemetry is currently working on the semantic conventions for messaging and batch is certainly a scenario that will be covered so in a month or two there will be more direction as to how we should handle it

ozangunalp:
@kenfinnigan I went for the most naive implementation at first. But giving a second thought, I don't see the need to create a span for each record in a batch message. We can have a single span per "batch of records" message with some custom attributes. And when OpenTelemetry comes up with semantic conventions for batch messaging we'll adapt it accordingly. WDYT ?

kenfinnigan:
@cescoffier and I briefly discussed batch spans the other day.

There are a couple of problems with such a change. It's not possible for a span to have multiple parent spans, so the batch span would not be a child of every message producer span, resulting in an unusual trace view. Clement also mentioned the batching can be grouped differently between consumers.

At the moment, I'd still prefer having a span per message

ozangunalp:
Ok I pushed a small change to remove `span.makeCurrent()` on batch records. @cescoffier would you have any suggestions to parallelize span creation for batch records? Is it ok to use `Infrastructure.getDefaultExecutor()` ?

cescoffier:
About parallelization, what gains are we expecting? 
@kenfinnigan does the creation of a span needs I/O (in that case, we may need to revisit how we do it). If no I/O is involved, I don't see lots of benefits for now. The best I can think of would be to use the `commonPool`, but the thread switch and the potentially limited parallelism of the environment (containers) may lead to more problems. 

We can revisit that later if we see it's a contention point.

kenfinnigan:
I took another look and there is nothing I/O related in creating a span

ozangunalp:
Great thanks @kenfinnigan

PR #1334, opened 2021-07-28
Add privileged blocks to avoid having to add

kabir:
MBeanServerPermission("createMBeanServer")
	MBeanPermission("*", "registerMBean, unregisterMBean")

when running with a security manager. Before this fix, it throws AccessControlException for those permissions in JmxReporter.unregister()/reregister().

Additionally it removes the need for RuntimePermission for getClassLoader, modifyThread
and setContextClassLoader when closing down Vert.X.

Some more information in https://issues.redhat.com/browse/WFLY-15071

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
@@ -3,6 +3,8 @@ package io.smallrye.reactive.messaging.kafka.impl;
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaExceptions.ex;
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


Ladicek (approved):
Looks OK, though I don't actually understand anything `SecurityManager` :-)

kabir:
@Ladicek I added a link to a Jira issue explaining a bit more https://issues.redhat.com/browse/WFLY-15071

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
```diff
@@ -124,7 +127,16 @@ public void setRebalanceListener() {
     Uni<ConsumerRecords<K, V>> poll() {
         if (polling.compareAndSet(false, true)) {
             return runOnPollingThread(c -> {
-                return c.poll(pollTimeout);
+                if (System.getSecurityManager() == null) {
```
cescoffier, line 130: The security manager can't change after startup right? So maybe we can capture that in a constant?

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
ozangunalp: Apparently, it can change if the previous one allows it :) BTW I just saw that from Java 17 Security Manager is deprecated for removal: https://openjdk.java.net/jeps/411

kabir: @darranl Do you know anything about this removal of the SM?

darranl: The use of the security manager remains fully supported in the next LTS release of Java, at this stage it is deprecated for later removal but there are ongoing discussions regarding how it is being handled and API compatibility issues.

cescoffier (approved):

cescoffier:
Rebased and merged PR.

PR #1339, opened 2021-08-02
Refactor pause/resume logic to address performance regression

ozangunalp:
When the downstream processing can't keep up with incoming message ingestion, the Kafka consumer applied pause/resume logic as a backpressure strategy. This prevents the buffered messages waiting to be processed from overflowing and causing an out-of-memory error.

The previous pause/resume logic paused – and resumed – the consumer very frequently, with each flap waiting the amount of `poll-timeout` setting (1 second by default) for the `Consumer#poll` to return. This was the cause of the regression in consumption throughput.

The pause/resume logic is no longer based on the number of downstream requests. Instead, it is driven by the size of the message buffer. We pause the Kafka consumer if the buffer queue has more items than 2 * `max.poll.records`. We resume when the queue has fewer items than 1 * `max.poll.records`.

The pause/resume logic is moved to the poll loop. When paused, the polling thread keeps on calling `Consumer#poll` as this is required for the consumer's heartbeat, but won't wait for `poll-timeout` and will return immediately. This makes the pause/resume logic as reactive as possible.

In addition, the handling of rebalancing during paused state is improved. When the consumer is paused newly assigned partitions as a result of rebalancing are paused as well. Resume will act on currently assigned partitions instead of previously paused one's as this may have changed as a result of rebalancing.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
@@ -3,8 +3,6 @@ package io.smallrye.reactive.messaging.kafka.impl;
 import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
 import java.time.Duration;
-import java.util.Queue;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -21,6 +19,10 @@ import io.smallrye.reactive.messaging.kafka.KafkaConnectorIncomingConfiguration;
 import io.vertx.core.Context;
 
 public class KafkaRecordStream<K, V> extends AbstractMulti<ConsumerRecord<K, V>> {
+    private static final int STATE_NEW = 0; // no request yet -- we start polling on the first request
+    private static final int STATE_POLLING = 1;
+    private static final int STATE_PAUSED = 2;
+    private static final int STATE_CANCELLED = 3;
 
     private final ReactiveKafkaConsumer<K, V> client;
     private final KafkaConnectorIncomingConfiguration config;
@@ -45,6 +47,11 @@ public class KafkaRecordStream<K, V> extends AbstractMulti<ConsumerRecord<K, V>>
         private final MultiSubscriber<? super ConsumerRecord<K, V>> downstream;
         private final boolean pauseResumeEnabled;
 
+        /**
+         * Current state: new (no request yet), polling, paused, cancelled
+         */
+        private final AtomicInteger state = new AtomicInteger(STATE_NEW);
+
         private final AtomicInteger wip = new AtomicInteger();
         /**
          * Stores the current downstream demands.
@@ -52,27 +59,14 @@ public class KafkaRecordStream<K, V> extends AbstractMulti<ConsumerRecord<K, V>>
         private final AtomicLong requested = new AtomicLong();
 
         /**
-         * Started flag.
-         */
-        private final AtomicBoolean started = new AtomicBoolean();
-
-        /**
-         * Paused / Resumed flag.
-         */
-        private final AtomicBoolean paused = new AtomicBoolean();
-
-        /**
          * The polling uni to avoid re-assembling a Uni everytime.
          */
         private final Uni<ConsumerRecords<K, V>> pollUni;
 
+        private final int maxQueueSize;
+        private final int halfMaxQueueSize;
         private final RecordQueue<ConsumerRecord<K, V>> queue;
         private final long retries;
-
-        /**
-         * {@code true} if the subscription has been cancelled.
-         */
-        private volatile boolean cancelled;
 
         public KafkaRecordStreamSubscription(
                 ReactiveKafkaConsumer<K, V> client,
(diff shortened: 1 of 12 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -90,6 +90,7 @@
...
 @ConnectorAttribute(name = "poll-timeout", type = "int", direction = Direction.INCOMING, description = "The polling timeout in milliseconds. When polling records, the poll will wait at most that duration before returning records. Default is 1000ms", defaultValue = "1000")
 @ConnectorAttribute(name = "pause-if-no-requests", type = "boolean", direction = Direction.INCOMING, description = "Whether the polling must be paused when the application does not request items and resume when it does. This allows implementing back-pressure based on the application capacity. Note that polling is not stopped, but will not retrieve any records when paused.", defaultValue = "true")
+@ConnectorAttribute(name = "max.poll.buffer.size", type = "int", direction = Direction.INCOMING, description = "The maximum number of polled messages to buffer waiting to be consumed. This is used to limit the memory consumption while implementing back-pressure with `pause-if-no-requests`. Defaults to `max.poll.records` times 5.")
```
Ladicek, line 93: I think the naming convention is `foo.bar.baz` for Kafka properties, and `foo-bar-baz` for "our" properties. So this should IMHO be `max-poll-buffer-size`.

Ladicek:
The rest looks fine IMHO.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
ozangunalp: nice catch

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStream.java
```diff
@@ -45,50 +47,49 @@ public void subscribe(
...
             this.client = client;
             this.pauseResumeEnabled = config.getPauseIfNoRequests();
             this.downstream = subscriber;
+            // Kafka also defaults to 500, but doesn't have a constant for it
             int batchSize = config.config().getOptionalValue(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, Integer.class).orElse(500);
-            this.queue = new RecordQueue<>(2 * batchSize);
+            this.maxQueueSize = batchSize * 2;
+            this.halfMaxQueueSize = batchSize;
+            // we can exceed maxQueueSize by at most 1 batchSize
+            this.queue = new RecordQueue<>(3 * batchSize);
```
loicmathieu, line 83: @ozangunalp I wonder why you allocate 3*batchSize ? It's not max capacity so it could be a little too much, as it should normally not go beyond 2*batchSize, so we may loose some memory due to pre-allocated array.

As the queue is never recreated, if it goes beyond 2*batchSize it will be enlarged once to 50% more capacity (I check on the ArrayDeque code) so it will be the same, but lazily done if needed.

Ladicek: If the queue has size `2 * batchSize - 10`, we can add more items to it. Items come in batches, of size at most `batchSize`. If we obtained a batch of max size, the queue would suddenly have `3 * batchSize - 10` items -- and only at that point we would pause.

I thought the comment explains it succinctly but clearly, which obviously isn't the case :-)

Just noting that this sizing algorithm is simple, but not the only one possible. Feel free to suggest a better one.

loicmathieu: Oh, yes, I now remember that queue.size() is compared to the maxQueueSize to know if pause should be done or not, so we need some overflow capacity anyway as everything happens async, and in batch.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
```diff
@@ -123,7 +123,7 @@ public void setRebalanceListener() {
     @SuppressWarnings("unchecked")
     Uni<ConsumerRecords<K, V>> poll() {
         if (polling.compareAndSet(false, true)) {
-            return runOnPollingThread(c -> paused.get() ? c.poll(Duration.ZERO) : c.poll(pollTimeout))
+            return runOnPollingThread(c -> paused.get() ? c.poll(Duration.ofMinutes(10)) : c.poll(pollTimeout))
```
Ladicek, line 126: Whoops :rofl:

ozangunalp: 🥇

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/client/PauseResumeTest.java
```diff
@@ -97,17 +105,222 @@ void testPauseResumeWithRequests() {
...
+        });
+
+        // Received first
+        await().until(() -> subscriber.getItems().size() == 1);
+
+        // Await pause
+        await().until(() -> !consumer.paused().isEmpty());
+
+        // Still paused
+        await().until(() -> !consumer.paused().isEmpty());
```
Ladicek, line 284: This will immediately return?

Ladicek (approved):
Great work!

PR #1346, opened 2021-08-11
Fix health for connectors that have no channels

Ladicek:
In case a connector is present (on classpath), it is always
instantiated. That is because connectors are CDI beans and they
are all injected into `ConfiguredChannelFactory`.

(Actually `ConfiguredChannelFactory` injects  implementations of
`IncomingConnectorFactory` and `OutgoingConnectorFactory`, but
those are typically connectors.)

For that reason, a connector may easily exist even if there are
no channels configured for it. Alternatively, channels may be
configured for a connector, but may all be disabled at runtime
via configuration.

In such case, health must not report "down", because this is
a perfectly normal situation (as we've just established).

The Kafka implementation of health used to check whether
at least one channel is present, and if not, its health
would report "down". This is to make sure that we don't
accidentally report "up" before the channels are correctly
initialized.

In this commit, we handle in-progress initialization globally,
by always returning "down" until `MediatorManager.start()`
finishes successfully. The connectors should no longer care
about this case and should only report health of their
existing channels.

### DIFF
--- a/smallrye-reactive-messaging-health/src/main/java/io/smallrye/reactive/messaging/health/HealthChecks.java
+++ b/smallrye-reactive-messaging-health/src/main/java/io/smallrye/reactive/messaging/health/HealthChecks.java
@@ -4,6 +4,11 @@ import org.eclipse.microprofile.health.HealthCheckResponse;
 import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
 
 public class HealthChecks {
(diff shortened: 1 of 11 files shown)


ozangunalp (approved):
👍  seems like other health checks are not impacted by this

Ladicek:
Right, both AMQP and RabbitMQ are fine and don't attempt anything smart :-)

PR #1360, opened 2021-08-16
Fail by default on deserialization issues

### LINKED ISSUE #1350: Default deserialization failure handler is danger
https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/DeserializerWrapper.java#L139

```java
    private T tryToRecover(String topic, Headers headers, byte[] data, Exception exception) {
        if (deserializationFailureHandler != null) {
            try {
                return (T) deserializationFailureHandler.handleDeserializationFailure(topic, this.handleKeys,
                        delegate.getClass().getName(), data, exception, headers);
            } catch (Exception e) {
                KafkaLogging.log.deserializationFailureHandlerFailure(deserializationFailureHandler.toString(), e);
                source.reportFailure(e, true);
            }
        } else {
            KafkaLogging.log.unableToDeserializeMessage(topic, exception);   // <---- danger!!!
        }
        return null;
    }
```

I use Quarkus + smallrye-reactive-messaging to consume AVRO message produced by Debezium from MySQL binlog, I'm surprised the consumer offset is still advanced even when Quarkus fails to deserialize.

Isn't it safer to just fail Kafka consumer when there is no failure handler?

cescoffier:
Fix #1350 - When no deserialization failure handler is set and a deserialization issue happens, propagate the failure and mark the application unhealthy. 
Also add an attribute to ﻿produce `null`  instead of that behavior. Returning `null` is what is happening at the moment.

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -133,6 +133,10 @@ Type: _string_ | false |
 
 Type: _string_ | false | 
 
+| *fail-on-deserialization-failure* | When no deserialization failure handler is set and a deserialization failure happens, report the failure and mark the application as unhealthy. If set to `false` and a deserialization failure happens, a `null` value is forwarded.
+
+Type: _boolean_ | false | `true`
+
 | *graceful-shutdown* | Whether or not a graceful shutdown should be attempted when the application terminates.
 
 Type: _boolean_ | false | `true`
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/DeserializerWrapper.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/DeserializerWrapper.java
@@ -59,13 +59,15 @@ public class DeserializerWrapper<T> implements Deserializer<T> {
 
     private final DeserializationFailureHandler<?> deserializationFailureHandler;
     private final KafkaSource<?, ?> source;
+    private final boolean failOnDeserializationErrorWithoutHandler;
 
     public DeserializerWrapper(String className, boolean key, DeserializationFailureHandler<?> failureHandler,
-            KafkaSource<?, ?> source) {
+            KafkaSource<?, ?> source, boolean failByDefault) {
         this.delegate = createDelegateDeserializer(className);
         this.handleKeys = key;
         this.deserializationFailureHandler = failureHandler;
         this.source = source;
+        this.failOnDeserializationErrorWithoutHandler = failByDefault;
     }
 
     /**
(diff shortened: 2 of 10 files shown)


Ladicek:
Should we do this in 3.10, to avoid breaking changes in Quarkus 2.2?

cescoffier:
Probably a good idea, I've updated the milestone.

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
```diff
@@ -133,6 +133,10 @@ Type: _string_ | false |
 
 Type: _string_ | false | 
 
+| *fail-on-deserialization-failure* | When no deserialization failure handler is set and a deserialization failure happens, report the failure and mark the application as unhealthy. If set to `false`, the record is skipped.
+
```
Dieken, line 137: > the record is skipped.

I feel this is a little confused,  actually consumer still gets a null record. maybe rephrase like "a fallback value `null` is returned to Kafka consumer".

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/DeserializerWrapper.java
```diff
@@ -137,6 +139,13 @@ private T tryToRecover(String topic, Headers headers, byte[] data, Exception exc
             }
         } else {
             KafkaLogging.log.unableToDeserializeMessage(topic, exception);
+            if (failOnDeserializationErrorWithoutHandler) {
+                source.reportFailure(exception, true);
+                if (exception instanceof KafkaException) {
+                    throw (KafkaException) exception;
+                }
+                throw new KafkaException(exception);
+            }
```
Ladicek, line 148: Javadoc of this method needs an update :-)

documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
cescoffier: Fixed.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/DeserializerWrapper.java
cescoffier: Fixed.

Ladicek (approved):
LGTM. Ah maybe one more question -- should this new config property added to the "config cleaner" thingie (which I don't really understand, so this is an honest question :-) )?

cescoffier:
Oh, yes! We need to add it to the cleaner.

So, let me explain why do we need to "clean" the configuration (a rant about the Kafka configuration mechanism).
The Kafka consumer and producer can receive any properties as a Map<String, Object>. The consumer and producer configuration are known and documented. However, the used serde and other extensions receive the configuration. So, we don't know the full list of properties. The connector passes all the attributes as they may be used.

However, unknown properties (even the one used by the serde and co) are reported as a warning... To avoid having too many warnings printed on the console, we remove the properties which we know are not used by the client objects.

Ladicek:
Ah, I see, thank you!

cescoffier:
I've added the attribute to the cleaner list.

Ladicek (approved):

PR #1365, opened 2021-08-19
Assign a different context for each Kafka consumer

cescoffier:
When configuring the Kafka connector with `partitions` > 0, it should assign a different Vert.x context for each created consumer.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
@@ -31,6 +31,8 @@ import io.smallrye.reactive.messaging.kafka.fault.KafkaFailStop;
 import io.smallrye.reactive.messaging.kafka.fault.KafkaFailureHandler;
 import io.smallrye.reactive.messaging.kafka.fault.KafkaIgnoreFailure;
 import io.smallrye.reactive.messaging.kafka.health.KafkaSourceHealth;
+import io.vertx.core.impl.EventLoopContext;
+import io.vertx.core.impl.VertxInternal;
 import io.vertx.mutiny.core.Vertx;
 
 public class KafkaSource<K, V> {
@@ -52,7 +54,7 @@ public class KafkaSource<K, V> {
     private final Instance<DeserializationFailureHandler<?>> deserializationFailureHandlers;
     private final Instance<KafkaConsumerRebalanceListener> consumerRebalanceListeners;
     private final ReactiveKafkaConsumer<K, V> client;
-    private final io.vertx.mutiny.core.Context context;
+    private final EventLoopContext context;
 
     public KafkaSource(Vertx vertx,
             String consumerGroup,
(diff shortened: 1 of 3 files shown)


cescoffier:
@loicmathieu FYI

cescoffier:
This should probably be backported in a 3.9.1. WDYT @Ladicek ?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSource.java
```diff
@@ -81,7 +83,10 @@ public KafkaSource(Vertx vertx,
         }
 
         configuration = config;
-        context = vertx.getOrCreateContext();
+        // We cannot use vertx.getOrCreate context as it would retrieve the same one everytime.
+        // It associates the context with the caller thread which will always be the same.
+        // So, we force the creation of different event loop context.
```
Ladicek, line 88: Event loop contexts are load-balanced between event loop threads, IIUC?

Ladicek: Or actually -- each event loop context is assigned to an event loop thread (in a round robin way I guess?) and stays that way. I think?

cescoffier: Yes, exactly. 

An event loop context is associated with a specific event loop thread and will stay on that event loop thread. 
If you create more contexts than the number of event loops threads (generally 2 * the number of CPU Core), you will have multiple contexts associated with the same event loop thread.

Ladicek: Cool, didn't know that, thanks!

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/commit/PartitionTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+        await()
+                .atMost(30, TimeUnit.SECONDS)
+                .until(() -> application.count() == expected);
+        assertThat(application.getReceived().keySet()).hasSize(3);
+
+        Properties properties = new Properties();
+        properties.put("bootstrap.servers", getBootstrapServers());
+        Admin admin = Admin.create(properties);
+        ConsumerGroupListing group = admin.listConsumerGroups().all().get().stream().findFirst().orElse(null);
```
Ladicek, line 66: There's multiple tests in this class that share the same Kafka broker, and the incoming channel config doesn't define a group ID, so group IDs will be autogenerated. So there will be eventually 3 consumer groups, and you don't know which one you get. No?

cescoffier: OH you are absolutely right!

Ladicek: Defining a unique group ID in channel config and using `Admin.listConsumerGroupOffsets(String)` should work IMHO.

cescoffier: Fixed

Ladicek (approved):

PR #1379, opened 2021-08-25
Allow listing the channels inside the KafkaClientService

loicmathieu:


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaClientService.java
```diff
@@ -41,4 +43,18 @@
...
+     * Get the names of all the Kafka incoming channels managed by this connector.
+     *
+     * @return the names of the Kafka consumer incoming channels.
+     */
+    Set<String> getConsumerChannels();
+
+    /**
+     * Get the names of all the Kafka outgoing channels managed by this connector.
+     *
+     * @return the names of the Kafka producer outgoind channels.
```
Ladicek, line 56:
```suggestion
     * @return the names of the Kafka producer outgoing channels.
```

Ladicek (approved):
LGTM, baring the typo :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaClientService.java
loicmathieu: Arg! I should stop doing three things at the same time, I know I'm not capable of doing multiple jobs in parallel.

cescoffier (approved):
Thanks!

loicmathieu:
Oups tests are failing because I reversed sink and source :(
Fixed now, next run should be fine.

Ladicek:
Crap I totally missed that! Good that you added a test :-)

PR #1383, opened 2021-08-27
WIP: Fix rebalance issue on load with multiple consumers

cescoffier:
- Fix client id generation when using multiple consumer (using the partitions attribute)
- Extend description of the throttled.unprocessed-record-max-age.ms attribute to indicate how to disable the monitoring.
- Improve tracking and cleanup of the throttled commit strategy


ozangunalp:
We can close this one for #1396

cescoffier:
You will need to rebase :-D

PR #1388, opened 2021-08-30
Avoid attribute hyphenation in the connector configuration tables

cescoffier:
Avoid hyphenation of attributes in the connector configuration table.


Ladicek (approved):
I couldn't believe this would work, until I realized you changed the `ui-bundle.zip` :-)

Issue #1393, opened 2021-09-02
kafka protobuf support


msillence:
The documentation mentions support for avro: https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/3.9/kafka/kafka.html

specifically:

```
mp.messaging.incoming.[channel].value.deserializer=io.confluent.kafka.serializers.KafkaAvroDeserializer
mp.messaging.incoming.[channel].schema.registry.url=http://<your_host>:<your_port>/
mp.messaging.incoming.[channel].specific.avro.reader=true
```

the obvious change to support protobuf is 

`mp.messaging.incoming.[channel].value.deserializer=io.confluent.kafka.serializers.protobuf.KafkaProtobufDeserializer`

and I do see a com.google.protobuf.DynamicMessage deliverd to my consumer but I'm less clear on how to make it deserialise to the correct specific object and what the reader should be or even if that is supported
the obvious has no effect:

`mp.messaging.incoming.[channel].specific.protobuf.reader=true`

msillence:
Found the documentation here:
https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-protobuf.html

adding the following lines worked for me:

mp.messaging.incoming.figaro-currency.specific.protobuf.value.type=....Key$Key
mp.messaging.incoming.figaro-currency.specific.protobuf.key.type=....Value$Value

I've not manged to persuade the autodetection to work with .derive_type=true

Shall I create a PR for the docs?

cescoffier:
Auto-detection only works in Quarkus and has not been implemented for Protobug (@ozangunalp / @Ladicek). 
That being said, a PR updating the doc is more than welcome!

Ladicek:
Protobug is a lovely typo :rofl:

msillence:
I am using quarkus, any hints on configuring it would be welcome. Looking at the code it seems it's `derive.type` and not `derive_type`

`mp.messaging.incoming.[chanel].derive.type=true`

generates lots of errors saying I need to set either `java_outer_classname` or `java_multiple_files` and setting them doesn't make the error go away. I'm only consuming the messages here so may be there's some magic on the producer when this is set. When I look at the topic in kafka I can see the type so I'm not quite sure why it doesn't pick it up in the same way as avro but from the code I've explored that is going to be a confluent question.

I presumed that protobug was not a typo but bitter experience? Any warnings for someone new?

[...]

PR #1394, opened 2021-09-03
add protobuf schema config to documentation

msillence:
As per https://github.com/smallrye/smallrye-reactive-messaging/issues/1393


documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
```diff
@@ -0,0 +1,45 @@
+[#kafka-protobuf-configuration]
+== Using Google Protobuf serializer/deserializer
+
+If you are using https://developers.google.com/protocol-buffers/[Protocol Buffers] serializer/deserializer, please note the following configuration properties.
+
+=== For https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-protobuf.html#serdes-and-formatter-protobuf[Confluent] Schema Registry
```
cescoffier, line 6: I don't believe we support links as headers (as the header is a link itself).

cescoffier: Can you add a link in the introduction sentence?

documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+Confluent protobuf library is `io.confluent:kafka-protobuf-serializer`.
+Note that this library is not available in Maven Central, you need to use the https://docs.confluent.io/clients-kafka-java/current/overview.html[Confluent Maven repository].
+
+==== Consumer
+
+[options="header"]
+|===
+| Property | Recommended value
+| value.deserializer | io.confluent.kafka.serializers.protobuf.KafkaProtobufDeserializer
```
cescoffier, line 16:
```suggestion
| value.deserializer | `io.confluent.kafka.serializers.protobuf.KafkaProtobufDeserializer`
```

documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
```diff
@@ -0,0 +1,45 @@
...
+mp.messaging.incoming.[channel].specific.protobuf.value.type=your.package.DomainObjectKey$Key
+mp.messaging.incoming.[channel].specific.protobuf.key.type=your.package.DomainObjectValue$Value
+----
+
+==== Producer
+
+[options="header"]
+|===
+| Property | Recommended value
+| value.serializer | io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializer
```
cescoffier, line 36:
```suggestion
| value.serializer | `io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializer`
```

cescoffier (changes requested):

documentation/src/main/doc/modules/kafka/pages/kafka.adoc
```diff
@@ -22,6 +22,7 @@ include::inbound.adoc[]
 include::outbound.adoc[]
 include::default-configuration.adoc[]
 include::avro-configuration.adoc[]
+include::kafka-protobuf-configuration.adoc[]
```
Ladicek, line 25: I personally think the `kafka-` prefix is not necessary. It's in a `kafka` directory, after all. For Avro, we have `avro-configuration.adoc`, so for Protobuf, I'd say `protobuf-configuration.adoc` should be enough.

documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
Ladicek: We have the same in the Avro guide, so should be fine.

documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
Ladicek: The Avro guide doesn't have backticks here either. I considered adding them, but figured it's probably better to avoid them, to allow for word wrapping. But I don't really mind.

cescoffier: From another issue we had recently, it does not seem to avoid word-wrapping...

Ladicek: Ah interesting! In any case, we don't have it for Avro, so I'd be fine with leaving this as is :-)

documentation/src/main/doc/modules/kafka/pages/protobuf-configuration.adoc
```diff
@@ -0,0 +1,45 @@
+[#kafka-protobuf-configuration]
+== Using Google Protobuf serializer/deserializer
+
+If you are using https://developers.google.com/protocol-buffers/[Protocol Buffers] serializer/deserializer, please note the following configuration properties.
+
+=== For https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-protobuf.html#serdes-and-formatter-protobuf[Confluent] Schema Registry
+
+Confluent protobuf library is `io.confluent:kafka-protobuf-serializer`.
+Note that this library is not available in Maven Central, you need to use the https://docs.confluent.io/clients-kafka-java/current/overview.html[Confluent Maven repository].
+
```
ozangunalp, line 10: Here we may add copy-pasteable maven instructions for adding repository+dependency. I know instructions for Gradle will be missing but that's a separate global effort to improve documentation.

ozangunalp (changes requested):
Looks good!

cescoffier:
Thanks!

PR #1396, opened 2021-09-08
Fix Kafka throttled commit strategy rebalance issue on load with multiple partitions

ozangunalp:
Throttled commit strategy for Kafka connector didn't pick up from where it left off after a rebalance between multiple consumers in a group.

With this PR
- Fix client id generation with multiple partitions
- Extend description of the throttled.unprocessed-record-max-age.ms attribute to indicate how to disable the monitoring.
- Fix elapsed time logging unprocessed messages waiting for acknowledgment for too long
- Improved logging for throttled commit strategy
- Offset store for throttled commit strategy uses `KafkaConsumer#committed` to receive the last committed offset for a partition. After a rebalance a consumer can receive records that are already processed (and committed). These duplicates need to be cleared from the offset store until the last processed record offset (committed offset - 1).

### DIFF
--- a/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
+++ b/documentation/src/main/doc/modules/connectors/partials/META-INF/connector/smallrye-kafka-incoming.adoc
@@ -101,7 +101,7 @@ Type: _string_ | false | `fail`
 
 Type: _string_ | false | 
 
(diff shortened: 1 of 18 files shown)


cescoffier (changes requested):
Didn't I update the documentation in my PR about the consumer id computation? I can't remember. The rest looks good.

ozangunalp:
@cescoffier I'll add the doc for the client id change. But first I need to fix these flaky tests.

cescoffier (approved):

PR #1398, opened 2021-09-08
Improve Emitter error handling

cescoffier:
Also, when the emitter uses @OnOverflow(DROP), the message would be dropped anyway, so in this case, it nacks the message but does not throw an error.

This commit also documents the hasRequests method.

Related to https://github.com/quarkusio/quarkus/issues/19940

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
@@ -13,18 +13,23 @@ import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.subscription.BackPressureStrategy;
 import io.smallrye.mutiny.subscription.MultiEmitter;
 import io.smallrye.reactive.messaging.helpers.BroadcastHelper;
+import io.smallrye.reactive.messaging.helpers.NoStackTraceException;
 
 public abstract class AbstractEmitter<T> {
+    public static final NoStackTraceException NO_SUBSCRIBER_EXCEPTION = new NoStackTraceException(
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
```diff
@@ -120,23 +128,27 @@ public boolean hasRequests() {
...
             throw ex.illegalArgumentForNullValue();
         }
 
-        MultiEmitter<? super Message<? extends T>> emitter = verify(internal, name);
+        MultiEmitter<? super Message<? extends T>> emitter = verify();
+        if (emitter == null) {
+            if (overflow == OnOverflow.Strategy.DROP) {
+                // There are no subscribers, but because we use the DROP strategy, just ignore the event.
+                // However, nack the message, so the sender can be aware of the rejection.
+                message.nack(new IllegalStateException("Unable to process message - no subscriber"));
```
fbricon, line 148: isn't that costly? generating a new stacktrace on every message until there's a subscriber?

cescoffier:
I can use an exception without stack trace.

kenfinnigan (dismissed):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/extension/AbstractEmitter.java
cescoffier: Fixed. I used a constant exception that does not capture the stack trace.

ozangunalp (approved):
This looks good to me. At first, I thought this can be implemented like the `ThrowingEmitter` but this is as close to the emit as it can get.

kenfinnigan (approved):

Issue #1403, opened 2021-09-13
Kafka Throttled commit strategy sometimes fails to commit after a consumer rebalance


pcasaes:
Symptom: After a consumer rebalance there's a chance that the throttled commit strategy will _break_ and stop commiting for one or more assigned partitions. This will only occur if using a non sticky consumer assignor (the default) and the chances increase if the consumer is far behind, ie: during a replay. A secondary symptom I noticed was having the commited offset go back after a rebalance.

Cause: As of 3.9.1 the throttled commit strategy implementation isn't idempotent
Per topic partition it expects to always receive records newer then ones already received. The new reactive consumer keeps a queue of polled records that persists over a rebalance. After a rebalance the vanila kafka consumer will rewind to the latest commit offset (even for topics partitions that stayed with their consumer instances) and duplicate records wrt to the already filled queue. The old vertx kafka consumer didn't keep a buffered queue or implement flow control so it wasn't an issue (side note: big fan of this new reactive consumer with flow control).

Work Around: Configure the [CooperativeStickyAssignor](https://kafka.apache.org/28/javadoc/org/apache/kafka/clients/consumer/CooperativeStickyAssignor.html). The kafka docs suggest this assignor over the default "Users should prefer this assignor for newer clusters." With this assignor the consumer won't rewind for topic partitions that aren't moved to other instances. Note: Upgrading to the sticky assignor is not trivial, see kafka docs!

Current fix: Make the commit strategy idempotent
This is simple enough and https://github.com/smallrye/smallrye-reactive-messaging/pull/1396/ fixes the issue. Kudos @ozangunalp 

But breaking the throttled commit strategy isn't the only issue caused by keeping messages from revoked partitions in the queue. Keeping these messages in the queue will increase the chances of duplicates and worse increase the chances of concurrent processing of the same record. Concurrent record processing can have performance implications depending on the application, ie: database locks (the first sign I noticed of this issue was from a collegue who reported lots of database locks for 1-2 minutes after a rebalance). In the case of a replay or slow consumer we can end up processing 6 times the batch size which also has performance implications.

Proposal: On revoke clear the queue of records from revoked topic partitions. 
On the default Round Robin assignor this effectively means we'll flush the queue entirely since with that assignor all topic partitions are always revoked. With a sticky assignor only the revoked partitions will be cleared. 

FYI: @cescoffier

cescoffier:
[...]

PR #1404, opened 2021-09-13
Clear queue of kafka records from revoked partitions

pcasaes:
@cescoffier This work is based on the findings of this issue https://github.com/smallrye/smallrye-reactive-messaging/issues/1403

The jist of it is that it clears the stream's queue of records from revoked partitions. This greatly reduces dup and concurrent processing during a rebalance when replaying a topic.

### DIFF
--- a/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
@@ -99,7 +99,7 @@ However, it reduces the risk of duplicates.
 * `ignore` performs no commit.
 This strategy is the default strategy when the consumer is explicitly configured with `enable.auto.commit` to `true`.
 It delegates the offset commit to the Kafka client.
-When `enable.auto.commit` is `true` this strategy DOES NOT guarantee at-least-once delivery.
+When `enable.auto.commit` is `true` this strategy **DOES NOT** guarantee at-least-once delivery.
 However, if the processing failed between two commits, messages received after the commit and before the failure will be re-processed.
 
 IMPORTANT: The Kafka connector disables the Kafka _auto commit_ if not explicitly enabled.
--- a/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
+++ b/documentation/src/main/doc/modules/kafka/pages/inbound.adoc
@@ -99,7 +99,7 @@ However, it reduces the risk of duplicates.
 * `ignore` performs no commit.
 This strategy is the default strategy when the consumer is explicitly configured with `enable.auto.commit` to `true`.
 It delegates the offset commit to the Kafka client.
-This strategy provides _at-least-once delivery_ if the channel processes the message without performing any asynchronous operations and when `enable.auto.commit` is set to `true`.
+When `enable.auto.commit` is `true` this strategy DOES NOT guarantee at-least-once delivery.
 However, if the processing failed between two commits, messages received after the commit and before the failure will be re-processed.
 
 IMPORTANT: The Kafka connector disables the Kafka _auto commit_ if not explicitly enabled.
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
@@ -224,6 +224,8 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
      * @return the map of partition -> offset that can be committed.
      */
     private Map<TopicPartition, Long> clearLesserSequentiallyProcessedOffsetsAndReturnLargestOffsetMapping() {
+        cleanupPartitionOffsetStore();
+
         Map<TopicPartition, Long> offsetsMapping = new HashMap<>();
         for (Map.Entry<TopicPartition, OffsetStore> entry : new HashSet<>(offsetStores.entrySet())) {
             if (assignments.contains(entry.getKey())) {
@@ -362,7 +364,7 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
         }
 
         void received(long offset) {
-            if (offset >= lastProcessedOffset) {
+            if (offset > lastProcessedOffset) {
                 this.receivedOffsets.offer(OffsetReceivedAt.received(offset));
                 unProcessedTotal.incrementAndGet();
             } else {
@@ -399,7 +401,6 @@ public class KafkaThrottledLatestProcessedCommit extends ContextHolder implement
                 }
             }
 
-            cleanupPartitionOffsetStore();
             removeReceivedOffsetsFromLastProcessedOffset();
 
             return -1;
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
@@ -229,7 +229,7 @@ public interface KafkaLogging extends BasicLogger {
     @Message(id = 18250, value = "The configuration property '%s' is deprecated and is replaced by '%s'.")
     void deprecatedConfig(String deprecated, String replace);
 
-    @LogMessage(level = Logger.Level.INFO)
+    @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 18251, value = "Committed %s")
     void committed(Map<TopicPartition, OffsetAndMetadata> offsets);
 
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
@@ -1,6 +1,16 @@
 package io.smallrye.reactive.messaging.kafka.impl;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
 
 import io.smallrye.mutiny.operators.AbstractMulti;
 import io.smallrye.mutiny.subscription.MultiSubscriber;
@@ -12,6 +22,8 @@ public class KafkaRecordBatchStream<K, V> extends AbstractMulti<ConsumerRecords<
     private final ReactiveKafkaConsumer<K, V> client;
     private final KafkaConnectorIncomingConfiguration config;
     private final Context context;
+    private final Set<KafkaRecordStreamSubscription<K, V, ConsumerRecords<K, V>>> subscriptions = Collections
+            .newSetFromMap(new ConcurrentHashMap<>());
 
     public KafkaRecordBatchStream(ReactiveKafkaConsumer<K, V> client,
             KafkaConnectorIncomingConfiguration config, Context context) {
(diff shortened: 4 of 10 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -224,6 +224,8 @@ private void startFlushAndCheckHealthTimer() {
      * @return the map of partition -> offset that can be committed.
      */
     private Map<TopicPartition, Long> clearLesserSequentiallyProcessedOffsetsAndReturnLargestOffsetMapping() {
+        cleanupPartitionOffsetStore();
```
pcasaes, line 227: Move this here since it was being called multiple times unecessarily from within the for loop on like 230

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -362,7 +364,7 @@ long getLastProcessedOffset() {
         }
 
         void received(long offset) {
-            if (offset >= lastProcessedOffset) {
+            if (offset > lastProcessedOffset) {
```
pcasaes, line 367: equals appears to be incorrect here. If the received offsett was already processed (equal to lastProcessedOffset) then it's already been delivered.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/RebalanceListeners.java
```diff
@@ -33,8 +33,9 @@ static ConsumerRebalanceListener createRebalanceListener(
                 @Override
                 public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
                     log.executingConsumerRevokedRebalanceListener(consumerGroup);
-                    // TODO Why don't we call the commit handler?
                     try {
+                        reactiveKafkaConsumer.removeFromQueueRecordsFromTopicPartitions(partitions);
+                        commitHandler.partitionsRevoked(partitions);
```
pcasaes, line 38: // TODO Why don't we call the commit handler?

We definitely should. This looks like a mistake. It probably hasn't been an issue because if the user is supplying there own listener there's a high chance they are using the ignore strategy and dealing with offsets themselves, examples:

Storing offsets else where for exactly once delivery
Always replace from a given timestamp and never commits (kafka backed in memory caches)

ozangunalp: I totally agree with this comment. But should we remove records from the queue for revoked topic partitions for all commit strategies or only for throttled, which has a high risk of creating duplicates on rebalance? I am thinking that we risk dropping them if we remove records from the queue when the commit strategy is NONE and auto.commit is enabled.

pcasaes: Under none and auto commit it will suffer the same issue. On rebalance the vanilla Kafka consumer will rewind to committed offset for any assigned topic partition and for those that are revoked it shouldn't process be processing. 

Per the SRRM documentation if the strategy is nome and auto commit Is enabled we don't guarantee at least once delivery. Per the Kafka documentation it's not possible to do so with s reactive framework.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
```diff
@@ -229,7 +229,7 @@ void acknowledgementFromRevokedTopicPartition(long offset, TopicPartition topicP
     @Message(id = 18250, value = "The configuration property '%s' is deprecated and is replaced by '%s'.")
     void deprecatedConfig(String deprecated, String replace);
 
-    @LogMessage(level = Logger.Level.INFO)
+    @LogMessage(level = Logger.Level.DEBUG)
```
pcasaes, line 232: Do we really want to print this as INFO every 5 seconds?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/RebalanceListeners.java
pcasaes: I personally have a use case were I set the strategy to none and auto.commit to false and control the starting point from a supplied rebalance listened. I definitely don't want to process another instance's record after a rebalance. 

Ultimately that's the issue. Continuing to process records from revoked partitions is "wrong".  Under version 2.x of SRRM this wouldn't happen because of how the vertx client is designed irrespective of settings (strategy and auto commit).

pcasaes: Actually the documentation says that NONE with auto commit does guarantee at least once delivery under certain conditions

https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaIgnoreCommit.java#L13

But that's no longer true under 3.9.1 (even without this PR). WIth auto commit true the vanilla client will commit every 5 seconds on poll. Under 2.2 that was true because a poll would only happen on request. The documentation should be update to state that there is no guarantee under NONE.

ozangunalp: I am not that familiar with the 2.2 implementation but I think this guarantee wasn't there even between 2.2 and 3.9.1.

Either the consumer stops polling until processing all received records, and therefore risks not responding to the heartbeat (SMRM 2.2), or the consumer continues polling (paused or not) in a separate thread while processing records, but risks auto committing records that are not yet processed (SMRM 3.9.1 with ignore commit strategy and enable.auto.commit=true). 

This leads me to the conclusion that in order to guarantee at-least-once delivery with SMRM Kafka we need offset tracking like throttled commit strategy or some custom offset checkpointing. We need to fix the Javadoc you mentioned and other documentation citing this.

cescoffier: As Ozan said, the 2.2 implementations had a critical issue when not consuming (fast) enough and where you could be kicked out because you didn't call poll. We have to call poll frequently (even when all partitions are paused) because of the heartbeat.

pcasaes: Right. The 2.2 implementation would only call poll on downstream requests. So this was true:

> When `enable.auto.commit` is `true` this strategy provides at-least-once delivery **IF** the channel processes the message without performing any asynchronous processing.


Big IF. Frankly if the user really wants to guarantee at-least-once they should stick with the throttled implementation or roll there own via a rebalance listener. The kafka docs are pretty adamant that if you are processing requests reactively then you are on your own **wrt at-least-once** guarantees and should not set `enable.auto.commit` to true.

pcasaes: Went ahead and added a commit updating the docs

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
```diff
@@ -25,7 +40,53 @@ public void subscribe(MultiSubscriber<? super ConsumerRecords<K, V>> subscriber)
...
+                        .add(topicPartition.partition()));
+
+        Map<String, Map<Integer, TopicPartition>> topicPartitionsMap = new HashMap<>();
+
+        synchronized (queue) {
+            Iterator<ConsumerRecords<K, V>> iterator = queue.iterator();
+            ArrayDeque<ConsumerRecords<K, V>> replacementQueue = new ArrayDeque<>();
+            while (iterator.hasNext()) {
+                ConsumerRecords<K, V> cr = iterator.next();
+                Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new HashMap<>();
```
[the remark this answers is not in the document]
ozangunalp: Actually as for the `removeIf` in `KafkaRecordStream` I'd expose a synchronized `rewriteQueue(Function<T, T> map)`. I think it'll help to understand what's happening here.

ozangunalp (changes requested):
This is a great addition! And also thanks for rebasing :)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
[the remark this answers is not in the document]
pcasaes: You are right. Sonarlint is complaining about synchronizing on a method parameter.

pcasaes:
@ozangunalp Changes made

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
```diff
@@ -25,7 +40,60 @@ public void subscribe(MultiSubscriber<? super ConsumerRecords<K, V>> subscriber)
         // Enqueue ConsumerRecords by batches, max poll records is considered 1
         KafkaRecordStreamSubscription<K, V, ConsumerRecords<K, V>> subscription = new KafkaRecordStreamSubscription<>(
                 client, config, subscriber, context, 1, (cr, q) -> q.offer(cr));
+        subscriptions.add(subscription);
         subscriber.onSubscribe(subscription);
     }
 
+    void removeFromQueueRecordsFromTopicPartitions(Collection<TopicPartition> partitions) {
```
ozangunalp, line 47: I don't completely understand what is happening in this method. 
I think I have a cleaner alternative for this but then maybe I am missing something. 
```java
void removeFromQueueRecordsFromTopicPartitions(Collection<TopicPartition> revokedPartitions) {
    if (revokedPartitions.isEmpty()) {
        return;
    }
    for (KafkaRecordStreamSubscription<K, V, ConsumerRecords<K, V>> subscription : subscriptions) {
        subscription.rewriteQueue(cr -> {
            Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new HashMap<>();
            for (TopicPartition topicPartition : cr.partitions()) {
                if (!revokedPartitions.contains(topicPartition)) {
                    records.put(topicPartition, cr.records(topicPartition));
                }
            }
            if (!records.isEmpty()) {
                // replace ConsumerRecords with the new one
                return new ConsumerRecords<>(records);
            } else {
                // remove from the queue
                return null;
            }
        });
    }
}
```

And the `rewriteQueue` is simply the logic that you wrote but in a separate method: 
```java
void rewriteQueue(Function<T, T> mapFunction) {
    synchronized (this.queue) {
        ArrayDeque<T> replacementQueue = new ArrayDeque<>();
        Iterator<T> iterator = queue.iterator();
        while (iterator.hasNext()) {
            T newRecord = mapFunction.apply(iterator.next());
            if (newRecord != null) {
                replacementQueue.offer(newRecord);
            }
            iterator.remove();
        }
        queue.addAll((Iterable<T>) replacementQueue);
    }
}
```

Tell me what do you think

pcasaes: doing an empty check is nice!

The ** removeFromQueueRecordsFromTopicPartitions** for **KafkaRecordBatchStream** is much cleaner to read. I was worried it would reorder the records in the batch but the **KafkaConsumer** does in fact fill up the batch from one topic partition before moving to the next. Regardless, we _don't have to_ guarantee order between topic partitions.

The ** rewriteQueue** method you wrote for **KafkaRecordBatchStream** works fine because we _have to_ replace all items since **ConsumerRecords** is immutable and we are talking about at most 3 items (with defaults). But for **KafkaRecordStream** it's inneficient since we only have to remove revoked records. We don't have to recreate the queue and in this case we are talking about at most 3000 item (with defaults).

I'll make some changes.

pcasaes: @ozangunalp changes pushed

pcasaes: Actually wrong! Performing a clear and then re appending the elements is no slower but possibly faster than doing a bulk remove.

pcasaes: Done

ozangunalp (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordBatchStream.java
ozangunalp: For `KafkaRecordStream` (`RecordQueue` with ConsumerRecord) I actually thought about using `removeIf` method inside ArrayDeque, which has a bulkRemove implementation. But what you've done is fine by me.

ozangunalp:
@cescoffier take a second look if you will, this looks ok for me.

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
```diff
@@ -99,7 +99,7 @@ However, it reduces the risk of duplicates.
 * `ignore` performs no commit.
 This strategy is the default strategy when the consumer is explicitly configured with `enable.auto.commit` to `true`.
 It delegates the offset commit to the Kafka client.
-This strategy provides _at-least-once delivery_ if the channel processes the message without performing any asynchronous operations and when `enable.auto.commit` is set to `true`.
+When `enable.auto.commit` is `true` this strategy DOES NOT guarantee at-least-once delivery.
```
cescoffier, line 102:
```suggestion
When `enable.auto.commit` is `true` this strategy **DOES NOT** guarantee at-least-once delivery.
```

cescoffier (dismissed):
I just made a minor cosmetic suggestion.

documentation/src/main/doc/modules/kafka/pages/inbound.adoc
pcasaes: @cescoffier applied

ozangunalp (approved):

PR #1424, opened 2021-09-20
Reduce the log level of Kafka messages (18237 and 18248)

cescoffier:
Related to https://github.com/quarkusio/quarkus/issues/20170.

Also, remove unused methods.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
@@ -56,7 +56,7 @@ public interface KafkaLogging extends BasicLogger {
 
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 18209, value = "Sending message %s to Kafka topic '%s'")
-    void sendingMessageToTopic(org.eclipse.microprofile.reactive.messaging.Message message, String topic);
+    void sendingMessageToTopic(org.eclipse.microprofile.reactive.messaging.Message<?> message, String topic);
 
     @LogMessage(level = Logger.Level.ERROR)
     @Message(id = 18210, value = "Unable to send a record to Kafka ")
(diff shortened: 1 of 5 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
```diff
@@ -175,14 +167,10 @@ void waitingForAckForTooLong(long offset, TopicPartition topicPartition, long de
...
     void setThrottledCommitStrategyReceivedRecordMaxAge(String group, long unprocessedRecordMaxAge);
 
-    @LogMessage(level = Logger.Level.INFO)
+    @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 18237, value = "Setting client.id for Kafka producer to %s")
     void setKafkaProducerClientId(String name);
 
-    @LogMessage(level = Logger.Level.INFO)
-    @Message(id = 18238, value = "Setting client.id for Kafka consumer to %s")
-    void setKafkaConsumerClientId(String name);
```
ozangunalp, line 184: Actually it'd be nice log the consumer client id's somewhere here : https://github.com/smallrye/smallrye-reactive-messaging/blob/5296684541b3bcad5bdfc65c3db4666a0350c3e3/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java#L221

ozangunalp (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/i18n/KafkaLogging.java
cescoffier: I'm actually testing it with Quarkus and add a few more condensed log messages (like we lost the bootstrap server too).

ozangunalp (approved):

PR #1434, opened 2021-09-27
Handle Vert.x Context in MutinyEmitter.send

cescoffier:
When the `MutinyEmitter.send` method is called from a Vert.x context, the acknowledgement event should be emitted on the same context.

This PR captures the context in the `send` method, if there is one.
In this case, it emits the result on the same context.

### DIFF
--- a/smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterInjectionTest.java
+++ b/smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterInjectionTest.java
@@ -7,10 +7,7 @@ import static org.awaitility.Awaitility.await;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -37,8 +34,10 @@ import io.smallrye.reactive.messaging.WeldTestBaseWithoutTails;
 import io.smallrye.reactive.messaging.annotations.Merge;
 import io.smallrye.reactive.messaging.extension.EmitterConfiguration;
 import io.smallrye.reactive.messaging.extension.MutinyEmitterImpl;
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+import io.vertx.core.impl.VertxInternal;
 
-@SuppressWarnings("ConstantConditions")
 public class MutinyEmitterInjectionTest extends WeldTestBaseWithoutTails {
 
     @Test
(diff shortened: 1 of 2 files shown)


cescoffier:
CC @unfromage

Emily-Jiang (commented):
@cescoffier Does this mean the core SmallRye Reactive Messaging require Vert.x?

cescoffier:
@Emily-Jiang Not as mandatory, but some features use Vert.x. If you use the regular Emitter, Vert.x is not required. If you use the MutinyEmitter, then yes. Suppose you are interested in using the MutinyEmitter without Vert.x, you can propose a PR checking if Vert.x is available. 

BTW, last time I checked, Open Liberty was using an older version of SmallRye Reactive Messaging. This PR only targets the 3.x (main) branch. 2.x is not changed.

Emily-Jiang:
> @Emily-Jiang Not as mandatory, but some features use Vert.x. If you use the regular Emitter, Vert.x is not required. If you use the MutinyEmitter, then yes. Suppose you are interested in using the MutinyEmitter without Vert.x, you can propose a PR checking if Vert.x is available.
> 
Thank you @cescoffier ! If Vert.x is not avaiable, will MutinyEitter work at all? MutinyEmitter is in the core though. If I consume the core module, without Vert.x, the module will not function. Right?
> BTW, last time I checked, Open Liberty was using an older version of SmallRye Reactive Messaging. This PR only targets the 3.x (main) branch. 2.x is not changed.
Yes, you were right. However, Open Liberty plans to consume the latest SmallRye Reactive Messaging. Therefore, I am cautious on introducing mandatory dependencies on Vert.x in the core again.

cescoffier:
@Emily-Jiang You can still use the module without Vert.x. As I said some features won't work, but they are not part of the spec (so you can pass the TCK without Vert.x being present). This PR does not change this, there are other dependencies. If that's a concern for your integration, PRs are welcome.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterInjectionTest.java
```diff
@@ -108,6 +110,27 @@ public void testWithProcessor() {
...
+        AtomicReference<Context> reference = new AtomicReference<>();
+        await().until(() -> unis.size() == 1);
+        unis.get(0).invoke(x -> reference.set(Vertx.currentContext())).await().indefinitely();
+
+        assertThat(reference.get()).isEqualTo(context);
+
+        context = ((VertxInternal) vertx).createEventLoopContext();
+        context.runOnContext(x -> unis.add(bean.emitter().send("b")));
+        await().until(() -> unis.size() == 2);
+        unis.get(1).invoke(x -> reference.set(Vertx.currentContext())).await().indefinitely();
```
ozangunalp, line 131: Isn't there a second assert missing ?

ozangunalp (changes requested):
The change itself seems ok to me. Although `AbstractEmitter` is still a head-scratcher.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/inject/MutinyEmitterInjectionTest.java
cescoffier: Yes, fixed.

ozangunalp (approved):

PR #1441, opened 2021-10-05
Compute and embed the jandex index in the main artifacts.

cescoffier:


Ladicek:
Nothing wrong, just wanted to highlight one thing: Jandex Maven plugin 1.2.0 uses Jandex 2.4.0.Final, and so produces an index that is incompatible with Quarkus 2.2. If that is not a concern, then all fine, but if it may be a concern, then downgrading to Jandex Maven plugin 1.1.1 would be a better choice.

cescoffier:
Thanks @Ladicek, you may explain why I was unable to read my @ConnectorAttribute (CLASS retention) from Jandex in my extension....

Ladicek:
Jandex doesn't store runtime-invisible (= `CLASS` retension) annotations. At least not yet :-) They will be there starting with Jandex 3.0.

cescoffier:
Fortunately, I just realized that we still use our own `@ConnectorAttribute`...

PR #1448, opened 2021-10-08
Select metrics impl using producer

ebullient:
Not sure how to write a good test for the micrometer path .. the current test for metrics has to add all of the MP Metrics CDI extensions, too.

Is there config to enable/disable metrics? Should there be?


ebullient:
> Now _if_ we want to have only one active metric decorator, that's something else, we can create a `MetricPublisherDecorator` and decorate publishers separately from other `PublisherDecorator`s and choose the default one to inject.
> 
> @ebullient @cescoffier wdyt?

Metrics are not free. MP Metrics classes will be detected on the classpath in Quarkus, e.g., because the Micrometer extension provides an adaptive MP Metrics layer, and I'm guessing others may end up doing the same, so you'd end up with two, which would mean duplicated meters and double the storage. Not what you want at all. 
 
 I further think a config option (to flat out disable entirely) should be on the table. We can handle this in Quarkus (and then veto the beans).. but that seems wasteful.

cescoffier:
@ozangunalp should we merge?

ozangunalp (approved):

PR #1505, opened 2021-11-08
Fix split packages

cescoffier:
@ozangunalp here is a pretty basic approach. The question is how much breaking does that approach introduce.


cescoffier:
Ah thanks! I was going to rebase, and you already did it!

cescoffier:
Let's merge!

PR #1513, opened 2021-11-15
Better failure handlers for Kafka serde

ozangunalp:
Adds failure handling for Kafka serialization.
When provided both the deserialization and serialization failure handler is called with a Uni wrapping serde operation to allow chaining operations like retry and recover with a fallback.

One question is should an unrecovered serialization failure be reported to the Kafka connector?


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/SerializationFailureHandler.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param data the data to serialize
+     * @param headers the record headers. May be {@code null}
+     * @return the recovered serialization result byte array
+     */
+    default byte[] decorateSerialization(Uni<byte[]> serialization, String topic, boolean isKey, String serializer, T data,
+            Headers headers) {
+        return serialization.onFailure().recoverWithItem(
+                throwable -> handleSerializationFailure(topic, isKey, serializer, data,
+                        throwable instanceof Exception ? (Exception) throwable : new KafkaException(throwable), headers))
+                .await().indefinitely();
```
cescoffier, line 56: We need to be sure we are not on the event loop here. Thanks to the new producer it should be fine, but let's double-check.

cescoffier (approved):
LGTM.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/SerializationFailureHandler.java
ozangunalp: I double checked that producer thread is used to call serializers, as expected.

PR #1516, opened 2021-11-16
Fix the configuration of the Kafka Consumer streams

cescoffier:
- set the concurrency to the number of consumers
- allows configuring the requests


cescoffier:
@loicmathieu Could you try with this PR. Set `requests` to 1024 in your case.

loicmathieu:
I'll try it tomorrow and give you feedback.

ozangunalp (approved):

PR #1518, opened 2021-11-17
Kafka latest commit strategy commit async without waiting

### LINKED ISSUE #1507: [Kafka - Latest commit strategy] Last messages from Kafka process very slowly, as each one causes a poll() with 1-second timeout to fetch nothing new
When we are processing messages from kafka using a simple @Incoming annotation, we see that as soon as the kafka topic is exhausted, and all the messages are loaded into the in-memory "buffer" in KafkaRecordStreamSubscription.queue, each request for another message (aka each invocation of our annotated method) will FIRST poll() to Kafka, wait for the timeout (1 second), find no additional items, then return an item from the in-memory buffer to the consumer.  Given the default value of the max.poll.records being 500, this means that if you are unlucky enough to pull the last 500 messages from kafka in one poll(), each and every one of those 500 messages will process at a rate of 1-per-second, as they each will first invoke poll(), find nothing new, then return a single item from the buffer (which has lots).

We don't have a patch to submit, nor an easy test case to submit here, so all I have is this explanation for now.  I think "going back to the Kafka well for more" doesn't need to happen until the in-memory buffer of already-fetched-messages is either depleted, or "nearing depletion" (subjective, I know, sorry). :)  Or, that the "going back to the Kafka well for more" always happens on a parallel thread that does not stop the dispatch() call from continuing to drain the in-memory buffer of messages.

I think this is due to a direct, blocking, call to poll() in io.smallrye.reactive.messaging.kafka.impl.KafkaRecordStream.KafkaRecordStreamSubscription#request - instead of putting the work onto the worker thread like other places, it makes a direct call to poll() if the state is not currently polling, but won't call dispatch() ... poll() will only call dispatch() after the steps in pollUni run, which is where the call to Kafka occurs.

Hopefully this is understandable as to what we're hitting and why, and where we think the pinch is.

ozangunalp:
Fixes #1507

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
@@ -1,5 +1,7 @@
 package io.smallrye.reactive.messaging.kafka.commit;
 
+import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
+
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
@@ -45,7 +47,6 @@ public class KafkaLatestCommit extends ContextHolder implements KafkaCommitHandl
 
     @Override
     public <K, V> CompletionStage<Void> handle(IncomingKafkaRecord<K, V> record) {
-        CompletableFuture<Void> future = new CompletableFuture<>();
         runOnContext(() -> {
             Map<TopicPartition, OffsetAndMetadata> map = new HashMap<>();
             TopicPartition key = new TopicPartition(record.getTopic(), record.getPartition());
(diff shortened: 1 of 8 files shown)


cescoffier (dismissed):

cescoffier (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java
```diff
@@ -54,13 +55,11 @@ public KafkaLatestCommit(Vertx vertx, KafkaConnectorIncomingConfiguration config
             if (last == null || last < record.getOffset() + 1) {
                 offsets.put(key, record.getOffset() + 1);
                 map.put(key, new OffsetAndMetadata(record.getOffset() + 1, null));
-                consumer.commit(map)
-                        .subscribe().with(x -> future.complete(null), future::completeExceptionally);
-            } else {
-                future.complete(null);
+                consumer.commitAsync(map)
```
itzg, line 58: Perfect! I was previously curious why a reactive communication layer wasn't making use of an async API call.

PR #1527, opened 2021-11-23
Make camel test more robust

OpenGuidou:
I've aligned the camel timeout to the one defined for nats, and adding a retry mechanism on the test.

Could you run the CI a few times and check if it improves the situation ?

### DIFF
--- a/smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/outgoing/OutgoingCamelTest.java
+++ b/smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/outgoing/OutgoingCamelTest.java
@@ -12,6 +12,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.camel.Exchange;
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.RetryingTest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/outgoing/OutgoingCamelTest.java
```diff
@@ -132,9 +133,11 @@ void testWithBeanDeclaringASimpleProcessorThatForwardIncomingHeaders() throws Ex
 
             final CompletableFuture<Exchange> completableFuture = CompletableFuture
                     .supplyAsync(
```
cescoffier, line 135: I would not use supplyAsync with a specific executor. The behavior is not the one you think when the level of parallelism is 1.

smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/outgoing/OutgoingCamelTest.java
```diff
@@ -132,9 +133,11 @@ void testWithBeanDeclaringASimpleProcessorThatForwardIncomingHeaders() throws Ex
...
-                            () -> camelContext().createConsumerTemplate().receive("nats:out?servers=127.0.0.1:7656&connectionTimeout=60000", 5000));
+                            () -> camelContext().createConsumerTemplate()
+                                    .receive("nats:out?servers=127.0.0.1:7656&connectionTimeout=60000", 60000));
```
cescoffier, line 137: I don't believe it's a timeout issue, but let's see.

cescoffier (changes requested):

cescoffier:
I'm going to execute the CI build a few times to be sure it fixes the issue.

cescoffier (approved):

OpenGuidou:
Should I understand that it was better that before ?

cescoffier:
I ran the CI a few times and it seems to be fine. Let's see how it is going to behave in the next few days.

PR #1540, opened 2021-12-06
Amqp connector option for setting sender/receiver capabilities

ozangunalp:
Adds "capabilities" configuration for both incoming and outgoing channels.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -2,15 +2,20 @@ package io.smallrye.reactive.messaging.amqp;
 
 import static io.smallrye.reactive.messaging.amqp.i18n.AMQPExceptions.ex;
 import static io.smallrye.reactive.messaging.amqp.i18n.AMQPLogging.log;
-import static io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction.*;
+import static io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction.INCOMING;
+import static io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction.INCOMING_AND_OUTGOING;
+import static io.smallrye.reactive.messaging.annotations.ConnectorAttribute.Direction.OUTGOING;
 import static java.time.Duration.ofSeconds;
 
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -295,6 +310,14 @@ void init() {
                 .ignore();
     }
 
+    public List<String> getClientCapabilities(AmqpConnectorCommonConfiguration configuration) {
+        if (configuration.getCapabilities().isPresent()) {
+            String capabilities = configuration.getCapabilities().get();
+            return Arrays.stream(capabilities.split(",")).map(String::trim).collect(Collectors.toList());
```
cescoffier, line 316: Probably best to filter out blank strings.

cescoffier (changes requested):
Looks great, I just made a very small "defensive" comment.

cescoffier:
Do not forget to run `mvn clean install -DskipTests` from the project root to be sure to have the up do date connector configuration tables in the documentation (that's what the up_to_date_antora is reporting).

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -295,6 +310,16 @@ void init() {
                 .ignore();
     }
 
+    public List<String> getClientCapabilities(AmqpConnectorCommonConfiguration configuration) {
+        if (configuration.getCapabilities().isPresent()) {
+            String capabilities = configuration.getCapabilities().get();
+            return Arrays.stream(capabilities.split(","))
+                    .filter(s -> !s.isEmpty())
+                    .map(String::trim).collect(Collectors.toList());
```
cescoffier, line 318: you should invert the two statements as you would not filter out " "

cescoffier (changes requested):

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
ozangunalp: That's right!

cescoffier (approved):

Issue #1535, opened 2021-11-30
Forward headers in kafka processor


OpenGuidou:
Hello,

Following https://github.com/smallrye/smallrye-reactive-messaging/issues/1458, it seems that the kafka headers are not  forwarded either between incoming channel and outgoing channel when using a processor.

Can I open a PR to add this feature ?

ozangunalp:
One thing to be cautious about is that Serializers can add headers to the record.
That's why I think it can make sense to add this behind a flag, like the propagate record key feature (#1324).
Also, we should not override the existing headers in OutgoingKafkaRecordMetadata.

cescoffier:
This feature makes sense.
As @ozangunalp said, it should be a flag, because it could leak sensible or ephemeral data.

PR #1549, opened 2021-12-20
Add option to propagate headers in kafka connector

OpenGuidou:
To fix https://github.com/smallrye/smallrye-reactive-messaging/issues/1535

I've added an option to the outgoing connector, similar to what's defined in the microprofile specs for rest clients:
https://download.eclipse.org/microprofile/microprofile-rest-client-1.3/microprofile-rest-client-1.3.html#_specifying_additional_client_headers


OpenGuidou:
Hello,
Any review ?

cescoffier:
@OpenGuidou You will need to rebase. We wanted the new documentation system to be merged before adding any new feature.

OpenGuidou:
Rebase done

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ce/KafkaCloudEventHelper.java
```diff
@@ -20,6 +21,7 @@
 import org.apache.kafka.common.header.internals.RecordHeader;
 import org.eclipse.microprofile.reactive.messaging.Message;
 
+import io.opentelemetry.api.internal.StringUtils;
```
cescoffier, line 24: Most probably wrong.

cescoffier (changes requested):
Looks great! 

I believe you used the wrong StringUtils.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ce/KafkaCloudEventHelper.java
OpenGuidou: Good catch

smallrye-reactive-messaging-kafka/pom.xml
```diff
@@ -39,6 +39,11 @@
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-semconv</artifactId>
     </dependency>
+    
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
```
cescoffier, line 45: Instead of adding this dependency (which causes an issue in native), I would prefer to have a simple utility method. You only use the `isNotEmpty` which can easily be implemented.

Something like this:

```java
public boolean isNotBlank(String s) {
    return s != null && ! s.trim().isEmpty();
}
```

cescoffier (changes requested):

cescoffier (approved):
Thanks!

Sorry for the long delay, that's a great feature.

PR #1575, opened 2022-01-18
Cache cloud event configuration

### LINKED ISSUE #1574: Kafka connector: cloud event support perrormance issue
The Cloud Event support for the Kafka connector has some performance issue.

It uses the configuration on each message to get default configuration which lead to calls to `Config.getOptionalValue` .
Some profiles shows up to 10% samples on this method call.
Caching the config value (wich never changes) should improve performance.

CPU profile is attached. 
[profile-cpu.zip](https://github.com/smallrye/smallrye-reactive-messaging/files/7887795/profile-cpu.zip)

I'll have a look to propose a PR to workaround this.

loicmathieu:
Fixes  #1574

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
@@ -67,7 +67,6 @@ public class KafkaSink {
     private final long retries;
     private final int deliveryTimeoutMs;
 
-    private final KafkaConnectorOutgoingConfiguration configuration;
     private final List<Throwable> failures = new ArrayList<>();
     private final KafkaSenderProcessor processor;
     private final boolean writeAsBinaryCloudEvent;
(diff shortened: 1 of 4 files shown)


cescoffier (dismissed):

loicmathieu:
This is a prototype to avoid accessing the configuration for each message. I duplicate the used configuratio items inside a `CloudEventConfiguration` object.

I notice other configuration items are used on each message, maybe a more broad solution can be implemented like a `RuntimeConfiguration` object that will be used to cache all configuration items used for each message.

loicmathieu:
@cescoffier there are other configuration items access on each message like [here](https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java#L311) and [here ](https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java#L370) (less problematic as health check are run each 30s usually not for each messahe). 

Maybe I should create a `KafkaSinkRuntimeConfiguration` instead that will copy also those config items to also improve the performance there ?

cescoffier:
Yes, everything avoiding accessing the config too much would be great. 
We can introduce RuntimeConfiguration class which stores the value and provide access. 

I'm wondering about `Optional`, maybe it would be better to use `null` checks (while it would require a bit more changes).

loicmathieu:
@cescoffier I created a `RuntimeKafkaSinkConfiguration` initialized inside the `KafkaSink` connector and used in all places that access the configuration out of the connector (I remove the configuration instance field so the issue can normaly not occurs anymore).

A added a few new instance fields (channel and readiness) to mimic what's in the `KafkaSource`.

I also check quickly the `KafkaSource` and do some refactoring to avoid reading any configuration, there was less issue on it so I didn't had to create a runtime config object.

I fix an issue with header propagation where a `map` was used instead of a `forEach` as map is not a terminal operation I think the headers was not correctly propagated but I may be misleading.

I keep the original `Optional` to avoid updating too much code.

ozangunalp (approved):

cescoffier (approved):

PR #1601, opened 2022-02-06
Fix rabbitmq interal queue size

### LINKED ISSUE #1594: RabbitMQ not setting internal queue size cause OOM exception
If using the Rabbitmq connector with Quarkus have something like 
```
    @Incoming("incmoing")
    @Outgoing("outgiing")
    @Blocking(ordered = false)
    @Acknowledgment(Acknowledgment.Strategy.MANUAL)
    fun handleIncomingMessage(msg: Message<ByteArray>): Message<ByteArray>? {
    ##some slow processing 
    }
```
This causes an Out-of-memory exception over time since the internal `ArrayDeque` fill up. 
If the `max-incoming-internal-queue-size` is set this problem disappears. 
Currently in the [RabbitMQConnector.java](https://github.com/smallrye/smallrye-reactive-messaging/blob/3462e3ed308cfca18e4e0347efc1995ed7a2f202/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java#L245) the default is set to `Integer.MAX_VALUE` 
I think this is not desirable for most clients
my thought is to either have a default value on the property or change the `Integer.MAX_VALUE` to something more reasonable.

jogueber:
Fixes #1594


cescoffier (approved):
Yes, looks better than Integer.Max.

PR #1603, opened 2022-02-07
PubSubConnector correct message type

silviu-negoita:
This allows users of the `smallrye-gcp-pubsub` to send their own `com.google.pubsub.v1.PubsubMessage`

In theory, the initial code should handle this scenario, but the condition `message instanceof PubSubMessage` is never true, because form very initial phase in the emitter process, the message is transformed into `Message<T>`, regardless of the initial type:

`io.smallrye.reactive.messaging.providers.extension.EmitterImpl` -> `(Message.of(payload, Metadata.empty()`
``` 
public synchronized CompletionStage<Void> send(T payload) {
        if (payload == null) {
            throw ProviderExceptions.ex.illegalArgumentForNullValue();
        } else {
            CompletableFuture<Void> future = new CompletableFuture();
            this.emit(Message.of(payload, Metadata.empty(), () -> {
                future.complete((Object)null);
                return CompletableFuture.completedFuture((Object)null);
            }, (reason) -> {
                future.completeExceptionally(reason);
                return CompletableFuture.completedFuture((Object)null);
            }));
            return future;
        }
    }
```

So, my change will extract the `com.google.pubsub.v1.PubsubMessage` from the `Message<T>`, even if it is sent as `com.google.pubsub.v1.PubsubMessage` or as the connector's wrapper `io.smallrye.reactive.messaging.gcp.pubsub.PubSubMessage`.

### DIFF
--- a/smallrye-reactive-messaging-gcp-pubsub/src/main/java/io/smallrye/reactive/messaging/gcp/pubsub/PubSubConnector.java
+++ b/smallrye-reactive-messaging-gcp-pubsub/src/main/java/io/smallrye/reactive/messaging/gcp/pubsub/PubSubConnector.java
@@ -179,6 +179,8 @@ public class PubSubConnector implements IncomingConnectorFactory, OutgoingConnec
     private static PubsubMessage buildMessage(final Message<?> message) {
         if (message instanceof PubSubMessage) {
             return ((PubSubMessage) message).getMessage();
+        } else if (message.getPayload() instanceof PubSubMessage) {
+            return ((PubSubMessage) message.getPayload()).getMessage();
         } else {
             return PubsubMessage.newBuilder()
                     .setData(ByteString.copyFromUtf8(message.getPayload().toString()))
--- a/smallrye-reactive-messaging-gcp-pubsub/src/main/java/io/smallrye/reactive/messaging/gcp/pubsub/PubSubConnector.java
+++ b/smallrye-reactive-messaging-gcp-pubsub/src/main/java/io/smallrye/reactive/messaging/gcp/pubsub/PubSubConnector.java
@@ -181,6 +181,8 @@ public class PubSubConnector implements IncomingConnectorFactory, OutgoingConnec
             return ((PubSubMessage) message).getMessage();
         } else if (message.getPayload() instanceof PubSubMessage) {
             return ((PubSubMessage) message.getPayload()).getMessage();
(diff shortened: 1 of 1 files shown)


cescoffier (approved):

cescoffier:
LGTM, Thanks!

silviu-negoita:
@cescoffier do you guys know when will be the next release? I want schedule a version bump on or projects to benefit of this change. Thanks!

cescoffier:
@ozangunalp wdyt? End of this week?

smallrye-reactive-messaging-gcp-pubsub/src/main/java/io/smallrye/reactive/messaging/gcp/pubsub/PubSubConnector.java
```diff
@@ -179,6 +179,10 @@ private static Path getCredentialPath(final Config config) {
     private static PubsubMessage buildMessage(final Message<?> message) {
         if (message instanceof PubSubMessage) {
             return ((PubSubMessage) message).getMessage();
+        } else if (message.getPayload() instanceof PubSubMessage) {
+            return ((PubSubMessage) message.getPayload()).getMessage();
+        } else if (message.getPayload() instanceof PubsubMessage) {
```
hakan0xFF, line 184: I have a question just for my understanding. Here we now have two equal checks `else if (message.getPayload() instanceof PubSubMessage)`. So ` return ((PubsubMessage) message.getPayload());` would never be executed. Or am I wrong?

cescoffier: Ah yes!

ozangunalp: Wait am I the one missing something ? Second if checks payload against `io.smallrye.reactive.messaging.gcp.pubsub.PubSubMessage` and the third one against `com.google.pubsub.v1.PubsubMessage`.

hakan0xFF: Oh, there is a lowercase `s` in `PubsubMessage` in the third check. Sorry for the inconvenience :/

PR #1608, opened 2022-02-13
fix broken incoming header propagation

cescoffier:
- Fix header propagation when using Cloud Event (with Kafka)
- Avoid having incoming metadata in the AMQP outgoing message
- Fix assertions and small improvements.


ozangunalp (approved):

cescoffier:
@ozangunalp I've set the version to 3.14, but maybe it's more a 3.14.1 thing. WDYT?

ozangunalp:
Yes, I think we better release it with 3.14.1

PR #1633, opened 2022-02-23
Message dispatch on duplicated Vert.x context, when available

ozangunalp:
- `ContextAwareMessage`: Base Message to create new duplicated context and include context in message metadata using `LocalContextMetadata`.
- `LocalContextMetadata`: Holds message Context and provides util methods for invoking processing methods on message context.
- `ContextDecorator`: Ensures consuming channels execute the stream on a duplicated context for each message dispatch.
- `Emitter` #send(T payload) implementations create message with duplicated context 
- `@Blocking` methods execution emit result on Vert.x context
- AMQP, Kafka, MQTT and RabbitMQ connectors consume ContextAwareMessages. 
- For Kafka: commit strategies execute on message context.

### DIFF
--- a/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
+++ b/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
@@ -54,6 +54,7 @@ public class MqttSource {
                             }
                             return multi;
                         })
+                        .onOverflow().buffer()
                         .onCancellation().call(() -> {
                             ready.set(false);
                             return Uni
--- a/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
+++ b/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
@@ -54,7 +54,7 @@ public class MqttSource {
                             }
                             return multi;
                         })
-                        .onOverflow().buffer()
+                        .onOverflow().buffer(config.getBufferSize())
                         .onCancellation().call(() -> {
                             ready.set(false);
                             return Uni
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
@@ -4,6 +4,7 @@ import static io.smallrye.reactive.messaging.providers.i18n.ProviderExceptions.e
 import static io.smallrye.reactive.messaging.providers.i18n.ProviderLogging.log;
 import static io.smallrye.reactive.messaging.providers.i18n.ProviderMessages.msg;
 
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -24,6 +25,7 @@ import io.smallrye.reactive.messaging.providers.connectors.WorkerPoolRegistry;
 import io.smallrye.reactive.messaging.providers.extension.HealthCenter;
 import io.smallrye.reactive.messaging.providers.helpers.BroadcastHelper;
 import io.smallrye.reactive.messaging.providers.helpers.ConverterUtils;
+import io.smallrye.reactive.messaging.providers.locals.LocalContextMetadata;
 
 public abstract class AbstractMediator {
 
@@ -95,11 +97,15 @@ public abstract class AbstractMediator {
         }
     }
 
+    protected <T> Uni<T> invokeOnMessageContext(Message<?> message, Object... args) {
+        return LocalContextMetadata.invokeOnMessageContext(message, x -> invoke(args));
+    }
+
     @SuppressWarnings("unchecked")
-    protected <T> Uni<T> invokeBlocking(Object... args) {
+    protected <T> Uni<T> invokeBlocking(Message<?> message, Object... args) {
         try {
             return workerPoolRegistry.executeWork(
-                    Uni.createFrom().emitter(emitter -> {
+                    LocalContextMetadata.invokeOnMessageContext(message, (m, emitter) -> {
                         try {
                             Object result = this.invoker.invoke(args);
                             if (result instanceof CompletionStage) {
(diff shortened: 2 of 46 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -196,7 +196,7 @@ private void startFlushAndCheckHealthTimer() {
         Uni<OffsetStore> uni;
         if (offsetStore == null) {
             uni = consumer.committed(recordsTopicPartition)
-                    .emitOn(runnable -> context.runOnContext(x -> runnable.run())) // Switch back to event loop
+                    .emitOn(runnable -> runOnRecordContext(record, runnable)) // Switch back to event loop
```
cescoffier, line 199: Here we need to go back to the event loop, but you go back to the record context.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -254,7 +254,7 @@ private void startFlushAndCheckHealthTimer() {
         CompletableFuture<Void> future = new CompletableFuture<>();
         // Be sure to run on the right context. The context has been store during the message reception
         // or partition assignment.
-        runOnContext(() -> {
+        runOnRecordContext(record, () -> {
```
cescoffier, line 257: same here, going back to the event loop should be enough

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
```diff
@@ -54,6 +54,7 @@ public MqttSource(Vertx vertx, MqttConnectorIncomingConfiguration config) {
                             }
                             return multi;
                         })
+                        .onOverflow().buffer()
```
cescoffier, line 57: we should make that configurable.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
```diff
@@ -74,6 +76,8 @@ public void initialize(Object bean) {
                     try {
                         return this.configuration.getMethod().invoke(bean, args);
                     } catch (Exception e) {
+                        System.out.println(Arrays.toString(args) + " vs. "
```
cescoffier, line 79: to be removed.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -71,9 +73,26 @@ public void init() {
         }
         Objects.requireNonNull(uni, msg.actionNotProvided());
 
+        Context currentContext = Vertx.currentContext();
```
cescoffier, line 76: wondering if we should not retrieve the message context.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/locals/ContextAwareMessage.java
```diff
@@ -0,0 +1,69 @@
+package io.smallrye.reactive.messaging.providers.locals;
+
+import java.util.Optional;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.eclipse.microprofile.reactive.messaging.Metadata;
+
+import io.smallrye.common.annotation.CheckReturnValue;
+import io.smallrye.common.vertx.VertxContext;
+import io.vertx.codegen.annotations.Nullable;
```
cescoffier, line 10: wrong one.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/locals/ContextDecorator.java
```diff
@@ -0,0 +1,117 @@
...
+            @Override
+            public void onFailure(Throwable throwable) {
+                // Release context on terminal events
+                Context root = ROOT_CONTEXT_UPDATER.getAndSet(this, null);
+                // Do NOT check for the current context and trampoline, as we may short-cut the completion event
+                // before the processing of the items.
+                if (root == null) {
+                    super.onFailure(throwable);
+                } else {
+                    root.runOnContext(ignored -> super.onFailure(throwable));
```
cescoffier, line 59: it works because the duplicated contexts should be on the same event loop. However, when using multiple partitions, it will not work.

cescoffier (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
ozangunalp: 👍

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/locals/ContextDecorator.java
ozangunalp: We are running only terminal events (onComplete, onFailure) on the captured root event loop. Otherwise we need to apply the Context decorator on each partition stream independently.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
ozangunalp: Indeed, I inlined the executeBlocking on the worker pool.

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
ozangunalp: Added a buffer-size config on incoming connector

cescoffier (approved):

PR #1637, opened 2022-03-02
Test Strimzi Test Container 0.101.0-rc2

cescoffier:
https://github.com/strimzi/test-container/releases/tag/0.101.0-rc1


ozangunalp:
Will be retried after: https://github.com/strimzi/test-container/pull/33

PR #1653, opened 2022-03-08
Add script to move to Jakarta

radcortez:


.github/workflows/build-main-branches.yml
```diff
@@ -6,6 +6,7 @@ on:
         branches:
             - main
             - 2.x
+            - jakarta
```
cescoffier, line 9: Do we need a new branch?

to-jakarta.sh
```diff
@@ -0,0 +1,50 @@
...
+find . -type f -name 'pom.xml' -exec sed -i '' 's/smallrye-parent/smallrye-jakarta-parent/g' {} +
+
+# move versions that support jakarta
+mvn versions:set-property -Dproperty=weld.version -DnewVersion=4.0.1.SP1
+mvn versions:set-property -Dproperty=microprofile-reactive-messaging.version -DnewVersion=3.0-RC2
+mvn versions:set-property -Dproperty=microprofile-reactive-streams.version -DnewVersion=3.0-RC1
+mvn versions:set-property -Dproperty=microprofile-config.version -DnewVersion=3.0.1
+mvn versions:set-property -Dproperty=microprofile-metrics-api.version -DnewVersion=4.0.1
+mvn versions:set-property -Dproperty=microprofile-health-api.version -DnewVersion=4.0
+mvn versions:set-property -Dproperty=smallrye-config.version -DnewVersion=3.0.0-RC1
```
cescoffier, line 16: How is this updated? Does dependabot handle these?

to-jakarta.sh
```diff
@@ -0,0 +1,50 @@
...
+mvn versions:set-property -Dproperty=smallrye-fault-tolerance.version -DnewVersion=6.0.0-RC1
+mvn versions:set-property -Dproperty=yasson.version -DnewVersion=2.0.3
+mvn versions:set-property -Dproperty=artemis.version -DnewVersion=2.20.0
+
+# artemis jakarta
+find . -type f -name 'pom.xml' -exec sed -i '' 's/\<artifactId\>artemis-jms-client\<\/artifactId\>/\<artifactId\>artemis-jakarta-client\<\/artifactId\>/g' {} +
+find . -type f -name 'pom.xml' -exec sed -i '' 's/\<artifactId\>artemis-server\<\/artifactId\>/\<artifactId\>artemis-jakarta-server\<\/artifactId\>/g' {} +
+
+# camel not ready yet
+find . -type f -name 'pom.xml' -exec sed -i '' 's/\<module\>smallrye-reactive-messaging-camel\<\/module\>/\<\!--\<module\>smallrye-reactive-messaging-camel\<\/module\>-->/g' {} +
```
cescoffier, line 30: That's very unfortunate.... We need to start a conversation with them.

cescoffier (changes requested):
Asking for some precisions.

.github/workflows/build-main-branches.yml
radcortez: It was to have the result of the transformation: https://github.com/smallrye/smallrye-reactive-messaging/tree/jakarta. 

How do you want to handle it here? For other SR projects, I keep a `jakarta` branch, to do the RC releases. When I need to update from `main`, I reset the branch to main and apply the transformations again.

to-jakarta.sh
radcortez: This will update the pom directly and keep the exact same pom structure and order. Once updated, dependabot will start looking into these versions (if they are not ignored in its config file)

to-jakarta.sh
radcortez: Sure.

radcortez:
Sorry, I should have given some more context here.

The idea was to use OpenRewrite to try out the migration. Unfortunately, we are blocked due to https://github.com/openrewrite/rewrite/issues/1422, so I wrote this script instead. If we are able to unblock OpenRewrite, we can reuse the script to rewrite the rules.

cescoffier:
Thanks for the context! What a shame to not be able to use the tool we are going to recommend. 
The rest looks good. What's the next step? Just run that script periodically?

cescoffier (approved):

radcortez:
> Thanks for the context! What a shame to not be able to use the tool we are going to recommend.

It seems that the OpenRewrite team is paying attention to the issue and they pushed a fix. I'll give it a try later today.

> What's the next step? Just run that script periodically?

We can automate it. Right now, I'm just doing it manually when required.

ozangunalp:
@radcortez About the CI failure you are having on Java 17 https://github.com/smallrye/smallrye-reactive-messaging/runs/5471908722?check_suite_focus=true

The failing test class is `io.smallrye.reactive.messaging.ft.RetryTest` with the following failures on Weld startup :
``` 
java.lang.reflect.GenericSignatureFormatError:
Signature Parse error: Expected Field Type Signature
	Remaining input: java.time.temporal.ChronoUnit
	at io.smallrye.reactive.messaging.ft.RetryTest.testImpossibleRecovery(RetryTest.java:141)
```

This is the same issue as in https://github.com/smallrye/smallrye-fault-tolerance/issues/412, which was due to https://github.com/jbossas/jboss-classfilewriter/issues/24.

AFAIS we are pulling  weld-se-core:4.0.1.SP1 pulls jboss-classfilewriter:1.2.4.Final instead of 1.2.5.Final.
If we can bump weld-se-core to 4.0.2.Final I think it'll resolve this issue.

ozangunalp:
Oh while I was writing the previous comment @cescoffier already merged this :) 
I confirm that  weld-se-core:4.0.2.Final fixes the issue.

radcortez:
Ah great! Thanks! I'll update. I'll run it locally first to confirm.

PR #1658, opened 2022-03-10
Use empty beans.xml and fix examples

cescoffier:
- Use blank beans.xml file to avoid the differences between javax and jakarta (and also be ready for CDI-Lite)
- Fix examples


ozangunalp (approved):
LGTM!

Issue #1660, opened 2022-03-11
Flaky HealthWithSubscriberTest / WeldTestBaseWithoutTails


IsSkyfalls:
After running `ChannelNameConflictTest`, which installs an invalid config file, the config values are not correctly cleared. Resulting in the next test, if using `WeldTestBaseWithoutTails`, consistently fail in `ConfiguredChannelFactory.detectNameConflict`. For me this is `HealthWithSubscriberTest`.

It seems that I'm the only one who has this issue, pretty strange.

Test logs:
```text
[INFO] Running io.smallrye.reactive.messaging.providers.connectors.ChannelNameConflictTest
Installed configuration:
# You should not be able to use the same channel name in an outgoing and incoming configuration
mp.messaging.incoming.name.connector=dummy
mp.messaging.incoming.name.increment=3
mp.messaging.outgoing.name.connector=dummy
---------
sources: [name]
sinks: [name]
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.012 s - in io.smallrye.reactive.messaging.providers.connectors.ChannelNameConflictTest
[INFO] Running io.smallrye.reactive.messaging.providers.extension.HealthWithSubscriberTest
sources: [name]
sinks: [name]
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.015 s <<< FAILURE! - in io.smallrye.reactive.messaging.providers.extension.HealthWithSubscriberTest
[ERROR] testHealthCheckWithSubscriber  Time elapsed: 0.015 s  <<< ERROR!
org.jboss.weld.exceptions.DeploymentException: SRMSG00073: Invalid configuration, the following channel names cannot be used for both incoming and outgoing: [name]
	at io.smallrye.reactive.messaging.providers.extension.HealthWithSubscriberTest.testHealthCheckWithSubscriber(HealthWithSubscriberTest.java:28)
Caused by: javax.enterprise.inject.spi.DeploymentException: SRMSG00073: Invalid configuration, the following channel names cannot be used for both incoming and outgoing: [name]
	at io.smallrye.reactive.messaging.providers.extension.HealthWithSubscriberTest.testHealthCheckWithSubscriber(HealthWithSubscriberTest.java:28)
```

Debug:
```test
/**
 * By spec, you cannot use the same channel name in an `incoming` configuration and `outgoing` configuration.
[...]

PR #1661, opened 2022-03-11
Fix failing HealthWithSubscriberTest

IsSkyfalls:
After running ChannelNameConflictTest, which installs an invalid config file, the config values are not correctly cleared.
Resulting in the next test, if using WeldTestBaseWithoutTails, consistently fail.

It's strange but this one-line change does fix it.

Issue: #1660


cescoffier (approved):

cescoffier:
Thanks!

Issue #1642, opened 2022-03-04
smallrye-rabbitmq: Unclear error message on IncomingRabbitMQMessage.getPayload()


IsSkyfalls:
When consuming messages from a rabbitmq queue, I got this horrible stacktrace.
```text
2022-03-04 22:08:24,899 ERROR [io.sma.rea.mes.rabbitmq] (pool-1-thread-1) out: io.smallrye.mutiny.CompositeException: Multiple exceptions caught:
	[Exception 0] java.lang.ClassCastException: class [B cannot be cast to class io.vertx.core.json.JsonObject ([B is in module java.base of loader 'bootstrap'; io.vertx.core.json.JsonObject is in unnamed module of loader io.quarkus.bootstrap.classloading.QuarkusClassLoader @65e98b1c)
	[Exception 1] io.smallrye.reactive.messaging.providers.ProcessingException: SRMSG00103: Exception thrown when calling the method xyz.skyfalls.placeholder.CrawlRequestProcessor#process
	at io.smallrye.mutiny.operators.uni.UniOnItemOrFailureFlatMap$UniOnItemOrFailureFlatMapProcessor.performInnerSubscription(UniOnItemOrFailureFlatMap.java:89)
	at io.smallrye.mutiny.operators.uni.UniOnItemOrFailureFlatMap$UniOnItemOrFailureFlatMapProcessor.onFailure(UniOnItemOrFailureFlatMap.java:65)
	at io.smallrye.mutiny.operators.uni.builders.UniCreateFromDeferredSupplier.subscribe(UniCreateFromDeferredSupplier.java:28)
	at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
	at io.smallrye.mutiny.operators.uni.UniOnItemOrFailureFlatMap.subscribe(UniOnItemOrFailureFlatMap.java:27)
	at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
	at io.smallrye.mutiny.operators.uni.UniOnItemTransformToMulti.subscribe(UniOnItemTransformToMulti.java:39)
	at io.smallrye.mutiny.operators.AbstractMulti.subscribe(AbstractMulti.java:39)
	at io.smallrye.mutiny.operators.multi.MultiFlatMapOp$FlatMapMainSubscriber.onItem(MultiFlatMapOp.java:193)
[...]

PR #1667, opened 2022-03-12
Make incoming message content_type overridable

IsSkyfalls:
This is my attempt resolving issue #1642 

The original implementation just gives up if you have a message without the `content_type` attribute set to a supported MIME type(and not using byte[]).
I added a configuration entry `content-type-override` on the incoming channel, so you don't have to modify the message producer if that's in another codebase. Also, a warning will be thrown once if you reach the fallback code and doesn't explicitly declare that you want a byte[]. In the end this shouldn't break any existing code.

You would still get the confusing stack trace if content_type and your function signature mismatches, but that's a much rarer case.


cescoffier (approved):
LGTM!

cescoffier:
Can you squash your commits?

IsSkyfalls:
Done, but somehow ci didn't pass.

cescoffier (approved):

Issue #1671, opened 2022-03-16
Kafka error when consuming from multiple partitions with DLQ strategy


victorgawk:
Having a Kafka consumer with:

* `partitions >= 2`
* `failure-strategy = dead-letter-queue`

will raise a warning when Smallrye tries to create the DLQ producer for the second, third partition and so on. This is happening because Smallrye is trying to create one DLQ producer for every partition. An exception is thrown because all producers are being created with the same client ID.

```
2022-03-16 10:59:20,258 WARN  [org.apache.kafka.common.utils.AppInfoParser] Error registering AppInfo mbean
javax.management.InstanceAlreadyExistsException: kafka.producer:type=app-info,id=kafka-dead-letter-topic-producer-my-topic
	at java.management/com.sun.jmx.mbeanserver.Repository.addMBean(Repository.java:436)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerWithRepository(DefaultMBeanServerInterceptor.java:1855)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerDynamicMBean(DefaultMBeanServerInterceptor.java:955)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerObject(DefaultMBeanServerInterceptor.java:890)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerMBean(DefaultMBeanServerInterceptor.java:320)
	at java.management/com.sun.jmx.mbeanserver.JmxMBeanServer.registerMBean(JmxMBeanServer.java:522)
	at org.apache.kafka.common.utils.AppInfoParser.registerAppInfo(AppInfoParser.java:64)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:436)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:292)
	at io.smallrye.reactive.messaging.kafka.impl.ReactiveKafkaProducer.<init>(ReactiveKafkaProducer.java:89)
	at io.smallrye.reactive.messaging.kafka.fault.KafkaDeadLetterQueue.create(KafkaDeadLetterQueue.java:72)
	at io.smallrye.reactive.messaging.kafka.impl.KafkaSource.createFailureHandler(KafkaSource.java:328)
[...]

PR #1676, opened 2022-03-18
Fix DLQ producer client id

victorgawk:
Fix DLQ producer client id to use `kafka-dead-letter-topic-producer-${consumer-client-id}` instead of `kafka-dead-letter-topic-producer-${channel.name}` (#1671)

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
@@ -60,7 +60,7 @@ public class KafkaDeadLetterQueue implements KafkaFailureHandler {
                 conf.getDeadLetterQueueKeySerializer().orElse(getMirrorSerializer(keyDeserializer)));
         deadQueueProducerConfig.put(VALUE_SERIALIZER_CLASS_CONFIG,
                 conf.getDeadLetterQueueValueSerializer().orElse(getMirrorSerializer(valueDeserializer)));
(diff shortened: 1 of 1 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
```diff
@@ -60,7 +60,7 @@ public static KafkaFailureHandler create(Map<String, ?> kafkaConfiguration,
                 conf.getDeadLetterQueueKeySerializer().orElse(getMirrorSerializer(keyDeserializer)));
         deadQueueProducerConfig.put(VALUE_SERIALIZER_CLASS_CONFIG,
                 conf.getDeadLetterQueueValueSerializer().orElse(getMirrorSerializer(valueDeserializer)));
-        deadQueueProducerConfig.put(CLIENT_ID_CONFIG, "kafka-dead-letter-topic-producer-" + conf.getChannel());
+        deadQueueProducerConfig.put(CLIENT_ID_CONFIG, "kafka-dead-letter-topic-producer-" + kafkaConfiguration.get(CLIENT_ID_CONFIG));
```
cescoffier, line 63: Is it always set? I would fallback to the channel name just to be sure.

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
ozangunalp: IIRC `ReactiveKafkaConsumer` ensures that client_id is set on the consumer.

ozangunalp (approved):

cescoffier (approved):

PR #1678, opened 2022-03-21
Use OpenTelemetry Instrumenter

### LINKED ISSUE #1572: Refactor OTel integration to use the Instrumenter API
The OTel integration should move to use the Instrumenter API, which is the recommended way to integrate with OTel.

radcortez:
- Fixes #1572


radcortez:
This is not totally done yet, but I wanted to get some feedback before moving forward.

I've tried to only replace the current existing code with the OTel APIs. It seems to work for Kafka and AMQP, but the RabbitMQ one is lacking tests, so no way to verify them at the moment (I'll need to add them).

What I don't like is the Tracing code is all over the place. And each connector is handling it in different ways. I'm wondering if we should introduce a SPI (like Vert.x did for tracing stuff) and make this well defined (it can also be used for gathering Metrics). 

On a separate topic, I was wondering if we could inject the TracingMetadata automatically, so the user doesn't have to do that. We may require to adjust the `Message` to add the OTel Context as a Metadata automatically if exists. The tricky part may be to extract the Context back again and add it to the right place from a `Uni`.

@cescoffier @ozangunalp any thoughts?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaSink.java
```diff
@@ -140,6 +142,15 @@ public KafkaSink(KafkaConnectorOutgoingConfiguration config, KafkaCDIEvents kafk
 
         log.connectedToKafka(client.getClientId(), config.getBootstrapServers(), topic);
 
+        KafkaAttributesExtractor kafkaAttributesExtractor = new KafkaAttributesExtractor();
+        InstrumenterBuilder<KafkaTrace, Void> builder = Instrumenter.builder(
+                GlobalOpenTelemetry.get(),
+                "io.smallrye.reactive.messaging.opentelemetry",
```
knutwannheden, line 148: Should the instrumentation name here be `io.smallrye.reactive.messaging.kafka` as used by `KafkaConnector#TRACER`?

radcortez: The instrumentation name indicates the instrumentation library name. It depends if we want to be more fine-grained or not.

ozangunalp:
@radcortez I've not finished reviewing the code but `TracingAppToAmqpTest.testFromAppToAmqp` is failing

ozangunalp:
@radcortez I looked at the changes again. I noticed that we are doing the same sequence of operations on the instrumenter.

I separated that code into a common module which also imports the otel apis. Here is my test PR : https://github.com/ozangunalp/smallrye-reactive-messaging/pull/12

Along the way I think I fixed that flaky test. 

Tell me what do you think.

radcortez:
> @radcortez I looked at the changes again. I noticed that we are doing the same sequence of operations on the instrument.

Yes, my plan was to refactor this later. Thanks for doing it :)

> I separated that code into a common module which also imports the otel apis. Here is my test PR : [ozangunalp#12](https://github.com/ozangunalp/smallrye-reactive-messaging/pull/12)

+1

> Along the way I think I fixed that flaky test in AMQP connector. There is something wrong with the RabbitMQ connector tracing config: the `tracing.attribute-headers` is never taken into account. As we discussed I think the decorator is not needed as it is a connector-specific code and can be added directly to the incoming channel stream.

The `tracing.attribute-headers` were used in the old implementation, but I think they don't make sense. These allowed to record in the span a configurable set of message headers, but these could clash with the OTel-defined span names, so I think we should remove them.

> Tell me what do you think.

+1 Good to go!

ozangunalp:
@cescoffier I am willing to take this in for 3.23. Could you take a look ?

cescoffier (approved):
LGTM!

PR #1680, opened 2022-03-23
Kafka Transactional emitter and exactly-once procesing

ozangunalp:


api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
```diff
@@ -21,12 +21,16 @@
 
     void register(String name, MutinyEmitter<?> emitter);
 
+    void register(String name, Class<?> emitterType, Object emitter);
```
cescoffier, line 24: Would `<X> void register(String name, Class<X> emitterType, X emitter);` works?

api/src/main/java/io/smallrye/reactive/messaging/EmitterConfiguration.java
```diff
@@ -0,0 +1,20 @@
+package io.smallrye.reactive.messaging;
+
+import org.eclipse.microprofile.reactive.messaging.OnOverflow;
+
+import io.smallrye.reactive.messaging.annotations.EmitterFactoryFor;
+
+public interface EmitterConfiguration {
```
cescoffier, line 7: Didn't we have a class named EmitterConfiguration already?

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * <p>
+ * Custom implementations can provide a CDI {@code @Produces} method to make their custom Emitter interface injectable into
+ * managed beans.
+ *
+ * @param <T> emitter implementation type, extends {@link MessagePublisherProvider}
+ */
+public interface EmitterFactory<T extends MessagePublisherProvider<?>> {
+
+    /**
+     * Create emitter implementation instance
```
cescoffier, line 20:
```suggestion
     * Create an emitter implementation instance.
```

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * @param configuration emitter configuration
+     * @param defaultBufferSize default buffer size
+     * @return Emitter implementation
```
cescoffier, line 24:
```suggestion
     * @return the emitter instance
```

smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ProducerBuilder.java
```diff
@@ -85,6 +86,11 @@
      */
     private ExecutorService executorService;
 
+    /**
+     *
```
cescoffier, line 90: Didn't you want to write something?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
```diff
@@ -87,6 +92,10 @@ public ReactiveKafkaProducer(Map<String, Object> kafkaConfiguration, String chan
 
         kafkaWorker = Executors.newSingleThreadExecutor(KafkaSendingThread::new);
         producer = new KafkaProducer<>(kafkaConfiguration, keySerializer, valueSerializer);
+        if (kafkaConfiguration.containsKey(ProducerConfig.TRANSACTIONAL_ID_CONFIG)) {
+            // TODO is this the right way ?
+            initTransactions().subscribeAsCompletionStage();
```
cescoffier, line 97: No, it will trigger the operationimmediately, but won't wait for the completion. So, you may end up with something not totally initialized.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
```diff
@@ -136,6 +145,54 @@ public ReactiveKafkaProducer(Map<String, Object> kafkaConfiguration, String chan
...
+                    try {
+                        work.accept(multiEmitter);
+                    } catch (Throwable throwable) {
+                        multiEmitter.fail(throwable);
+                    } finally {
+                        multiEmitter.complete();
+                    }
+                })
+                        .onItem().transformToUniAndConcatenate(this::send)
+                        .collect().asList())
```
cescoffier, line 188: What's the reason to collect the iteam?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
```diff
@@ -136,6 +145,54 @@ public ReactiveKafkaProducer(Map<String, Object> kafkaConfiguration, String chan
...
+                .chain(this::commitTransaction)
+                .onFailure().recoverWithUni(throwable -> {
+                    log.warnf(throwable, "Aborting transaction for producer id &s in channel %s", this.clientId, this.channel);
```
cescoffier, line 191: &s -> %s

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Mark the current transaction for abort.
+     */
+    void markForAbort();
+
+    /**
+     * Is the current transaction marked for abort.
+     */
+    boolean isMarkedForAbort();
```
cescoffier, line 49: Shouldn't it be: isRollbackOnly?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param payload the <em>thing</em> to send, must not be {@code null}.
+     * @see io.smallrye.reactive.messaging.MutinyEmitter#send(Object)
+     */
+    void send(T payload);
+
+    /**
+     * Mark the current transaction for abort.
+     */
+    void markForAbort();
```
cescoffier, line 44: From where comes this name? Is it something coming from Kafka? JTA TXs have a different working.

cescoffier (changes requested):
Looks good! I've made a few comment but nothing major.

api/src/main/java/io/smallrye/reactive/messaging/EmitterConfiguration.java
ozangunalp: Yes, I extracted an interface and moved to the api module. It is passed to the `EmitterFactory` interface, which is also created in the api module. The provider contains the default impl `DefaultEmitterConfiguration`, similarly to `MediatorConfiguration`.

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
ozangunalp: 👍

ozangunalp: To my defense, emitter is created once at materialization :)

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
ozangunalp: 👍

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
ozangunalp: I hesitated to run it on the starter thread. But I think it is better to run it on the sender thread and in case of failure report it to the channel.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
ozangunalp: It was to ensure that all records from `MultiEmitter` were actually sent before committing the transaction. 
IIRC `toUni` did only took the first item.

BTW maybe should remove this method, though it is working, it was an early test I did and replaced by the `KafkaTransactionsImpl`.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
ozangunalp: Kafka transactions are aborted, not rolled back. I think the `abort` flag may a utility for Kafka transactions. 
If a transaction is marked for abort the records would be produced normally, only to be aborted (so not committed) at the end.

smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ConsumerBuilder.java
```diff
@@ -433,13 +460,29 @@ public long position(TopicPartition partition) {
...
+     * Reset to the last committed position
+     */
+    public void resetToLastCommittedPositions() {
+        Map<TopicPartition, OffsetAndMetadata> committed = committed();
+        for (TopicPartition tp : assignment()) {
+            OffsetAndMetadata offsetAndMetadata = committed.get(tp);
+            if (offsetAndMetadata != null) {
+                getOrCreateConsumer().seek(tp, offsetAndMetadata.offset());
+            } else {
+                getOrCreateConsumer().seekToBeginning(Collections.singleton(tp));
```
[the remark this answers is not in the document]
Ladicek: Ah, this is in the test companion. Agree that in tests, it's probably better to rewind to the beginning.

api/src/main/java/io/smallrye/reactive/messaging/ChannelRegistry.java
Ladicek: I guess I'd add the type parameter here just for symmetry with `getEmitter` :-)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/ReactiveMessagingExtension.java
```diff
@@ -53,26 +65,13 @@
...
-        Channel stream = ChannelProducer.getChannelQualifier(pip.getInjectionPoint());
-        if (stream != null) {
-            mutinyEmitterInjectionPoints.add(pip.getInjectionPoint());
-        }
-    }
-
-    @SuppressWarnings("deprecation")
-    <T extends io.smallrye.reactive.messaging.annotations.Emitter<?>> void processStreamLegacyEmitterInjectionPoint(
-            @Observes ProcessInjectionPoint<?, T> pip) {
+    <T> void processStreamEmitterInjectionPoint(@Observes ProcessInjectionPoint<?, T> pip) {
```
Ladicek, line 68: The type parameter here isn't used anywhere, this is equivalent to `@Observes ProcessInjectionPoint<?, ?> pip`. That's generally not recommended (it will likely be slow), but if you really want the set of emitter types to be open-ended, there's no better way I guess.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/providers/wiring/EmitterFactoryForLiteral.java
```diff
@@ -0,0 +1,30 @@
+package io.smallrye.reactive.messaging.providers.wiring;
+
+import javax.enterprise.util.AnnotationLiteral;
+
+import org.eclipse.microprofile.reactive.messaging.Emitter;
+
+import io.smallrye.reactive.messaging.MutinyEmitter;
+import io.smallrye.reactive.messaging.annotations.EmitterFactoryFor;
+
+public final class EmitterFactoryForLiteral extends AnnotationLiteral<EmitterFactoryFor> implements EmitterFactoryFor {
```
Ladicek, line 10: Nitpicking: a `Literal` class is typically added directly inside the qualifier annotation.

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,85 @@
...
+
+!!!warning "Experimental"
+    Kafka Transactions is an experimental feature.
+
+[Kafka transactions](https://docs.google.com/document/d/11Jqy_GjUGtdXJK94XGsEIK7CP1SnQGdp2eF0wSw9ra8/edit) enable atomic writes to multiple Kafka topics and partitions.
+Inside a transaction, a producer writes records to the Kafka topic partitions as it would normally do.
+If the transaction completes successfully, all the records previously written to the broker inside that transaction will be _committed_, and will be readable for consumers.
+If an error during the transaction causes it to be _aborted_, none will be readable for consumers.
+
+There are a couple of fundamental things to consider before using transactions are :
```
Ladicek, line 11:
```suggestion
There are a couple of fundamental things to consider before using transactions:
```

smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ConsumerBuilder.java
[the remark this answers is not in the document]
ozangunalp: That's a very good point. And it is the same in `ReactiveKafkaConsumer#resetToLastCommittedPositions`. 

I think this is an issue only if the `auto.offset.reset` of the consumer is `latest` and it had never committed any offsets. 
In that case, I don't know what the correct behaviour is. If we do `seekToEnd`, we would miss any records that were written in the meantime. In the context of a transaction, if no committed offset exists, we could capture the position before and return to that position. WDYT ?

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,85 @@
...
+
+* Each transactional producer is configured with a unique identifier called the `transactional.id`.
+This is used to identify the same producer instance across application restarts.
+By default, it is not configured, and transactions cannot be used.
+When it is configured, the producer is limited to idempotent delivery, therefore the `enable.idempotence=true` is implied.
```
Ladicek, line 16:
```suggestion
When it is configured, the producer is limited to idempotent delivery, therefore `enable.idempotence=true` is implied.
```

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,85 @@
...
+
+* For __only__ reading transactional messages, a consumer needs to explicitly configure it's `isolation.level` property to `read_committed`.
+This will make sure that the consumer will deliver only records committed inside a transaction, and filter out messages from aborted transactions.
+
+* It should also be noted that, this does not mean all records atomically written inside a transaction will be read atomically by the consumer.
```
Ladicek, line 21:
```suggestion
* It should also be noted that this does not mean all records atomically written inside a transaction will be read atomically by the consumer.
```

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/providers/wiring/EmitterFactoryForLiteral.java
ozangunalp: 👍  At first, I only needed it for tests. But I think it could be useful for Quarkus build time as well.

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,85 @@
...
+It can be used as a regular emitter `@Channel`:
+
+``` java
+{{ insert('kafka/outbound/KafkaTransactionalProducer.java') }}
+```
+
+!!!note
+    When `transactional.id` is provided `KafkaProducer#initTransactions` is called when the underlying Kafka producer is created.
+
+The function given to the `withTransactions` method receives a `TransactionalEmitter` for producing records, and returns a `Uni` that will provide the result for a successful transaction.
```
Ladicek, line 41:
```suggestion
The function given to the `withTransactions` method receives a `TransactionalEmitter` for producing records, and returns a `Uni` that will provide the result of a successful transaction.
```

Also perhaps this?

```suggestion
The function given to the `withTransactions` method receives a `TransactionalEmitter` for producing records, and is must return a `Uni` that provides the result of the transaction.
```

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConsumer.java
```diff
@@ -181,4 +182,16 @@
...
+     */
+    Uni<ConsumerGroupMetadata> consumerGroupMetadata();
+
+    /**
+     * For each assigned topic partition reset to last committed offset or the beginning
+     *
+     * @return the Uni emitting {@code null} when the reset has been executed.
+     */
+    Uni<Void> resetToLastCommittedPositions();
+
```
Ladicek, line 196: The existing methods are all annotated `@CheckReturnValue`, maybe these should too?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaProducer.java
```diff
@@ -80,6 +86,52 @@
...
+     * <p>
+     * Note that emitted {@link ProducerRecord}s are lazily sent, meaning that returned {@code Uni} will need to be subscribed
+     * for the
+     * whole transaction to execute.
+     *
+     * @param work {@code MultiEmitter} consumer for emitting {@code ProducerRecord}s
+     * @return {@code Uni} representing the transaction execution.
+     */
+    Uni<Void> withTransaction(Consumer<MultiEmitter<? super ProducerRecord<K, V>>> work);
+
```
Ladicek, line 134: Same here, do these need `@CheckReturnValue`?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
Ladicek: This should be fine, no? Since all sending actions are serialized on the sending thread, nothing can start actually sending before the init transactions process completes -- asuming that `Producer.initTransactions()` is synchronous, which I don't know :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
Ladicek: I thought the `ReactiveKafka{Consumer,Producer}` are supposed to expose the reactive analogues of the low-level Kafka API and anything higher-level (like this `withTransactions`) is supposed to live elsewhere? (In other words, I guess I agree with removing this, since the higher-level API people should actually use is `KafkaTransactions`.)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactions.java
```diff
@@ -0,0 +1,113 @@
...
+ * Uni&lt;Void&gt; process(KafkaRecordBatch&lt;String, String&gt; batch) {
+ * return txEmitter.withTransactions(batch, emitter -> {
+ * for (KafkaRecord&lt;String, String&gt; rec : batch) {
+ * txEmitter.send(KafkaRecord.of(record.getKey(), record.getPayload()));
+ * }
+ * return Uni.createFrom().voidItem();
+ * });
+ * }
+ *
+ *
```
Ladicek, line 50: This javadoc (and perhaps other places, if I recall correctly?) says `withTransactions`, but the actual method is called `withTransaction`.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
Ladicek: I would say we don't have to follow the JTA terminology. And probably _shouldn't_, because Kafka transactions are a wholly different beast.

Ladicek:
Added a ton of nitpicking comments and one that seems pretty important, the one about thread safety in `KafkaTransactionsImpl`.

Also, do we really want to expose the ability to define custom emitters to users? To be able to implement an emitter factory, they have to reach into internals (inject `ChannelRegistry` and call its methods) anyway, no?

smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ConsumerBuilder.java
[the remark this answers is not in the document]
Ladicek: Good point, I totally overlooked that `KafkaTransactionsImpl` calls `ReactiveKafkaConsumer#resetToLastCommittedPositions` and that rewinds to the beginning on transaction abort. That seems tricky :-)

I'd guess your suggestion (to remember the offset at the beginning of the transaction and rewind there, if we don't have a committed offset) makes sense. I'm obviously no expert, but I can't think of anything better. Perhaps there's a configuration property already, in Kafka, that affects this behavior?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/ReactiveMessagingExtension.java
Ladicek: Technically we could add a marker interface to our own types and treat the spec-defined `Emitter` as a special case.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactionsImpl.java
```diff
@@ -0,0 +1,175 @@
...
+import io.smallrye.reactive.messaging.providers.extension.MutinyEmitterImpl;
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+
+public class KafkaTransactionsImpl<T> extends MutinyEmitterImpl<T> implements KafkaTransactions<T> {
+
+    private final KafkaClientService clientService;
+    private final KafkaProducer<?, ?> producer;
+
+    private volatile Transaction<?> currentTransaction;
```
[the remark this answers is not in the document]
ozangunalp: A transaction is actually bound to a producer, with its `transactional.id` so there can only be one transaction from a producer. In terms of the usage, there will be only one `KafkaTransactionImpl` emitter instance created per channel (so producer instance), which will hold the current transaction.

ozangunalp:
@Ladicek Thanks for the review! 

I kept the EmitterFactoryForLiteral as is, I figured in Quarkus we'll need a little more.

Indeed I think custom emitters is more destined for connector development. 
`ChannelRegistry` is already part of the `api` (although not in spec) and its implementation is provided as `@ApplicationScoped` bean. So we are not exposing anything that isn't already there.

Ladicek:
Oh I didn't notice `ChannelRegistry` is a public API, I always thought of that as a private thing :-) I'm still not sure we want to expose the emitter factory stuff publicly, but I'll leave that choice to you.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactionsImpl.java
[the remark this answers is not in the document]
Ladicek: Okay, but that doesn't really answer my question, does it?

Imagine I have this simple JAX-RS resource, using a plain old `Emitter` like we document everywhere:

```java
@Path("/emit")
public class EmitToKafka{
    @Inject
    @Channel("my-channel")
    Emitter<SomeDto> emitter;

    @POST
    public void emit(SomeDto data) {
        emitter.send(data);
    }
}
```

Nothing wrong with that, right?

Now, if I understand correctly, `KafkaTransactions` is yet another kind of `Emitter`, so I can do this:

```java
@Path("/emit")
public class EmitToKafka{
    @Inject
    @Channel("my-channel")
    KafkaTransactions<SomeDto> emitter;

    @POST
    public void emit(SomeDto data) {
        emitter.withTransaction(em -> {
            em.send(data);
        });
    }
}
```

And as far as I'm aware, there's exactly _nothing_ preventing the `EmitToKafka.emit` method from being invoked concurrently on multiple threads.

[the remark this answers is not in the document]
ozangunalp: Oh look there are still some `@CheckReturnValue` missing...
You are right, calling `withTransaction` concurrently doesn't make sense. We need to protect it against this.
After going back and forth I pushed a change doing that. 
For the sake of simplicity I also removed the ability to do nested calls inside the same transaction.

Ladicek:
So, do I understand it correctly that at any given point in time, there may be at most 1 Kafka transaction? (Given that Kafka is an append-only log, that seems plausible?)

Ladicek:
Ah, yes, per producer, gotcha. Thanks!

api/src/main/java/io/smallrye/reactive/messaging/MessagePublisherProvider.java
```diff
@@ -0,0 +1,14 @@
+package io.smallrye.reactive.messaging;
+
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.reactivestreams.Publisher;
+
+/**
+ * Framework-facing interface for the Emitter implementations.
+ *
+ * @param <T> message payload type.
```
cescoffier, line 9: Minor: maybe mention that the emitter is emitting Message<T>. While the current wording is correct, it led to confusion.

documentation/src/main/docs/concepts/emitter.md
```diff
@@ -200,3 +200,45 @@ an event, you can `sendAndForget`:
...
+It'll use the emitter factory to create the emitter implementation and will register the implementation into the `ChannelRegistry`.
+
+Note that the `CustomEmitterFactory` also uses the `ChannelRegistry` and provides the custom emitter with `@Produces`.
+
+Finally, the application can inject and use the `CustomEmitter` as a normal emitter channel:
+
+``` java
+{{ insert('emitter/EmitterFactoryExamples.java', 'custom-emitter-usage') }}
+```
+
```
cescoffier, line 244: Maybe mention that the emitter can be called concurrently.

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,92 @@
+# Kafka Transactions and Exactly-Once Processing
+
+!!!warning "Experimental"
+    Kafka Transactions is an experimental feature.
+
+[Kafka transactions](https://docs.google.com/document/d/11Jqy_GjUGtdXJK94XGsEIK7CP1SnQGdp2eF0wSw9ra8/edit) enable atomic writes to multiple Kafka topics and partitions.
```
cescoffier, line 6: Are you sure about linking that document?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
```diff
@@ -229,6 +229,34 @@ public ConsumerRebalanceListener getRebalanceListener() {
         }
     }
 
+    @Override
+    @CheckReturnValue
```
cescoffier, line 233: Minor: Do we need the annotation here, as it's on the interface?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
```diff
@@ -229,6 +229,34 @@ public ConsumerRebalanceListener getRebalanceListener() {
...
+            c.pause(assignments);
+            Map<TopicPartition, OffsetAndMetadata> committed = c.committed(assignments);
+            for (TopicPartition tp : assignments) {
+                OffsetAndMetadata offsetAndMetadata = committed.get(tp);
+                if (offsetAndMetadata != null) {
+                    // Seek to next offset position
+                    c.seek(tp, offsetAndMetadata.offset());
+                } else {
+                    // Seek to beginning of the topic partition
+                    c.seekToBeginning(Collections.singleton(tp));
```
cescoffier, line 252: I am wondering if we should enforce the reset strategy in this case. I'm happy with the current choice, but maybe mention that this is done on purpose.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/MutinyEmitterImpl.java
```diff
@@ -9,6 +9,7 @@
 import io.smallrye.common.annotation.CheckReturnValue;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.subscription.Cancellable;
+import io.smallrye.reactive.messaging.EmitterConfiguration;
```
cescoffier, line 12: Minor: Any reaosn for this import?

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/providers/wiring/EmitterFactoryForLiteral.java
cescoffier: it can still be added to the annotation as a static class.

cescoffier (changes requested):
Looks almost good. I did a few minor comments.

documentation/src/main/docs/kafka/transactions.md
ozangunalp: Other than this document, I could not find any mention of the Kafka transactions on the official kafka.apache.org site.  We can include a link to one of the Confluent sites though.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/MutinyEmitterImpl.java
ozangunalp: It receives `EmitterConfiguration` as a constructor parameter.

documentation/src/main/docs/kafka/transactions.md
cescoffier: Let me ask around. Let's keep the google doc for now.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
ozangunalp: I agree, removing this.

smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/providers/wiring/EmitterFactoryForLiteral.java
ozangunalp: 👍

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaConsumer.java
ozangunalp: @cescoffier what do you mean by "enforce the reset strategy"? 
With @Ladicek we discussed this before [here](https://github.com/smallrye/smallrye-reactive-messaging/pull/1680#discussion_r839659815), but I am not convinced of the usefulness of the alternative I proposed. 
For now, I'll add more documentation for this.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/ReactiveMessagingExtension.java
ozangunalp: Indeed, I added `EmitterType` marker interface. This should speed things up a bit for bean discovery.

api/src/main/java/io/smallrye/reactive/messaging/annotations/EmitterFactoryFor.java
```diff
@@ -29,4 +31,27 @@
      * @return the emitter interface
      */
     Class<?> value();
+
+    final class EmitterFactoryForLiteral extends AnnotationLiteral<EmitterFactoryFor> implements EmitterFactoryFor {
```
Ladicek, line 35: Nitpicking: the literal class, when nested in the annotation type, is by convention called just `Literal`.

api/src/main/java/io/smallrye/reactive/messaging/annotations/EmitterFactoryFor.java
```diff
@@ -29,4 +31,27 @@
...
+        }
+
+        @Override
+        public Class<?> value() {
+            return value;
+        }
+
+        public static EmitterFactoryFor EMITTER = of(org.eclipse.microprofile.reactive.messaging.Emitter.class);
+        public static EmitterFactoryFor MUTINY_EMITTER = of(MutinyEmitter.class);
+        public static EmitterFactoryFor LEGACY_EMITTER = of(io.smallrye.reactive.messaging.annotations.Emitter.class);
```
Ladicek, line 54: Nothing wrong with these literals being here, but do we want to make them part of the public API?

ozangunalp: If the Literal becomes the public API, I don't see why these wouldn't.

api/src/main/java/io/smallrye/reactive/messaging/MessagePublisherProvider.java
ozangunalp: Done

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,93 @@
...
+The function given to the `withTransaction` method receives a `TransactionalEmitter` for producing records, and returns a `Uni` that provides the result of the transaction.
+If the processing completes successfully, the producer is flushed and the transaction is committed.
+If the processing throws an exception, returns a failing `Uni`, or marks the `TransactionalEmitter` for abort, the transaction is aborted.
+
+If this method is called on a Vert.x context, the processing function is also called on that context.
+Otherwise, it is called on the sending thread of the producer.
+
+!!!important
+    A transaction is considered _in progress_ from the call to the `withTransaction` until the returned `Uni` results in success or failure.
+    While a transaction is in progress subsequent calls to the `withTransaction`, including nested one's inside the given function, will throw `IllegalStateException`.
```
Ladicek, line 50:
```suggestion
    While a transaction is in progress, subsequent calls to the `withTransaction`, including nested ones inside the given function, will throw `IllegalStateException`.
```

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,93 @@
...
+    Note that, in Reactive Messaging the execution of processing methods is already serialized, unless `@Blocking(ordered = false)` is used.
```
Ladicek, line 51:
```suggestion
    Note that in Reactive Messaging, the execution of processing methods is already serialized, unless `@Blocking(ordered = false)` is used.
```

documentation/src/main/docs/kafka/transactions.md
```diff
@@ -0,0 +1,93 @@
...
+    If `withTransaction` can be called concurrently, for example from a REST endpoint, it is recommended to limit the concurrency of the execution.
+    This can be done using the `@Bulkhead` annotation from Microprofile Fault Tolerance.
```
Ladicek, line 53: Or by marking the method (which calls `withTransaction`) as `synchronized`?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaProducer.java
```diff
@@ -80,6 +85,40 @@
...
+     * @return the Uni emitting {@code null} when the abort transactions executes.
+     */
+    @CheckReturnValue
+    Uni<Void> abortTransaction();
+
+    /**
+     *
+     * @param offsets topic partition offsets to commit into transaction
+     * @param groupMetadata consumer group metadata of the exactly-once consumer
+     * @return the Uni emitting {@code null} when the init transactions executes.
```
Ladicek, line 116: I'm not sure when exactly the returned `Uni` completes, but it sure isn't when "init transactions executes" :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactions.java
```diff
@@ -0,0 +1,127 @@
...
+ * &#64;Incoming("in-channel")
+ * Uni&lt;Void&gt; process(KafkaRecordBatch&lt;String, String&gt; batch) {
+ * return txEmitter.withTransaction(batch, emitter -> {
+ * for (KafkaRecord&lt;String, String&gt; rec : batch) {
+ * txEmitter.send(KafkaRecord.of(record.getKey(), record.getPayload()));
+ * }
+ * return Uni.createFrom().voidItem();
+ * });
+ * }
```
Ladicek, line 49: Code snippet should have indentation IMHO.

documentation/src/main/docs/kafka/transactions.md
ozangunalp: Only making the caller synchronized would not help, as the transaction is ongoing until the `Uni` returns. That's why the `@Bulkhead(1)` works, as it'll schedule another request (if any) once the previous one returns.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaProducer.java
ozangunalp: :) true

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param msg the <em>Message</em> to send, must not be {@code null}
+     * @throws IllegalStateException if the channel has been cancelled or terminated or if an overflow strategy of
+     *         {@link OnOverflow.Strategy#THROW_EXCEPTION THROW_EXCEPTION} or {@link OnOverflow.Strategy#BUFFER BUFFER} is
+     *         configured and the emitter overflows.
+     * @see io.smallrye.reactive.messaging.MutinyEmitter#send(Message)
+     */
+    <M extends Message<? extends T>> void send(M msg);
+
+    /**
+     * /**
```
Ladicek, line 29: Superfluous line :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
```diff
@@ -0,0 +1,50 @@
...
+ * Send methods immediately return by calling the underlying Emitter.
+ *
+ * @param <T> payload type
+ */
+public interface TransactionalEmitter<T> {
+
+    /**
+     * Sends a message to the Kafka channel.
+     * <p>
+     * Immediately returns without waiting the result by dispatches the message to the underlying Emitter.
```
Ladicek, line 17:
```suggestion
     * Immediately returns without waiting for the result by dispatching the message to the underlying Emitter.
```

Is probably what you meant?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/TransactionalEmitter.java
```diff
@@ -0,0 +1,50 @@
...
+     *         configured and the emitter overflows.
+     * @see io.smallrye.reactive.messaging.MutinyEmitter#send(Message)
+     */
+    <M extends Message<? extends T>> void send(M msg);
+
+    /**
+     * /**
+     * Sends a payload to the channel.
+     * <p>
+     * Immediately returns without waiting the result by dispatches the message to the underlying Emitter.
```
Ladicek, line 32:
```suggestion
     * Immediately returns without waiting for the result by dispatching the message to the underlying Emitter.
```

Is probably what you meant?

ozangunalp: Exactly :)

documentation/src/main/docs/kafka/transactions.md
Ladicek: Ah ah, true! But bulkhead has a limited queue, so it will sooner or later start bailing out. It isn't meant for serializing requests, but for limiting concurrency, so this usage is a little creative :-)

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactionsImpl.java
```diff
@@ -0,0 +1,192 @@
...
+
+    private class Transaction<R> implements TransactionalEmitter<T> {
+
+        private final Uni<Void> beforeCommit;
+        private final Function<R, Uni<R>> afterCommit;
+
+        private final Uni<Void> beforeAbort;
+        private final Function<Throwable, Uni<R>> afterAbort;
+
+        boolean abort;
```
Ladicek, line 116: Should this be `volatile` in case `withTransaction` wasn't called on a Vert.x context?

(Genuine question. I'm not sure what executes on the sending thread and what on the original thread that called `withTransaction`.)

Ladicek:
OK, I spent like 30 or 40 minutes staring at `KafkaTransactionsImpl`, my eyes were bleeding and my head was threatening to explode into million pieces, each of them screaming "sequential composition!", but it all looks fine. Phew! :-)

documentation/src/main/docs/kafka/transactions.md
ozangunalp: I agree, but at least it is configurable. I honestly couldn't think of anything else that will "serialize" requests from imperative code, without being a reactive stream itself.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/transactions/KafkaTransactionsImpl.java
ozangunalp: I remember asking myself the same question for `abort`. But I don't know why I didn't make it volatile. 

So for the execution threads: 
If `withTransaction` is NOT called on Vert.x context, all pipeline, including the execution of the passed function will run on the producer sender thread. 
If it is called on Vert.x context, call the passed function on that context, but producer operations will run on the producer sender thread.

So yes if called on Vert.x context `abort` can be accessed by both the sender thread and the Vert.x thread.

PR #1704, opened 2022-04-11
Disable tracing for amqp broadcast test

ozangunalp:
Attempt to fix failing AmqpSourceTest#testBroadcast. 

Draft because I don't understand how it passed before.


cescoffier:
Oh, is this the reason of the many unstable build we had over the weekend?

ozangunalp:
Yes and I still can't understand how it passed before. Now it fails on every ref in local.
Anyways I am merging.

PR #1706, opened 2022-04-12
Make number of RabbitMQ connections configurable per queue

### LINKED ISSUE #1699: Allow multiple consumers per channel for RabbitMQ connector
Currently for each incoming Smallrye messaging channel there is one RabbitMQ consumer created.
It would be beneficial to allow the creation of multiple RabbitMQ consumers per channel.
- concurrency could be increased (there is [a similar issue for Kafka](https://github.com/smallrye/smallrye-reactive-messaging/issues/329))
- support for sharded queues

Our use case is the following:
We have a high load environment where one RabbitMQ node is not able to process the full load, so our queues are sharded and thus distributed over all RabbitMQ nodes. 
If there are fewer RabbitMQ consumers than RabbitMQ nodes, not all messages will be fetched.

If each of our clients could be configured to create multiple consumers (greater than the number of RabbitMQ nodes), they would be evenly distributed over all queues, thus all messages could be consumed. 

[Here is the corresponding code location.](https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java#L245)

pravussum:
Implements  #1699
This adds a new config parameter `connection-count` for incoming channels which will control how many new connections are created. 
This allows to consume from a [sharded queue](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbitmq_sharding) - which is distributed over several RabbitMQ cluster nodes - using a single instance of an application. Otherwise the application would only consume from a single RabbitMQ node and the messages delivered to the other shards would not be consumed at all. 

Due to a limitation of the underlying Vert.x RabbitMQClient implementation, [a Vert.x RabbitMQ client is bound to a single connection and channel](https://github.com/vert-x3/vertx-rabbitmq-client/blob/master/src/main/java/io/vertx/rabbitmq/impl/RabbitMQClientImpl.java#L62). This might become inefficient, especially when connections get created and destroyed a lot. On the other hand, when just several consumers per incoming channel are created (without using a new client + connection + channel), the consumers will reuse the same channel and connection and will end up consuming from the same RabbitMQ node, rather than being distributed over several nodes.

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
@@ -37,6 +37,7 @@ import com.rabbitmq.client.impl.CredentialsProvider;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.operators.multi.processors.BroadcastProcessor;
+import io.smallrye.mutiny.tuples.Tuple2;
 import io.smallrye.reactive.messaging.annotations.ConnectorAttribute;
 import io.smallrye.reactive.messaging.health.HealthReport;
 import io.smallrye.reactive.messaging.health.HealthReporter;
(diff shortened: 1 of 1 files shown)


cescoffier (approved):
LGTM

kdubb:
@pravussum Can you open a new PR with documentation for this feature?

Issue #1722, opened 2022-04-26
Smallrye-reactive-messaging-mqtt should support pem files for tls connection


cdmikechen:
Mqtt client should support ca / client cert / client key pem files. 
Suppose we are using quarkus native image to deploy on k8s by using smallrye-reactive-messaging. We can't easily mount keystore or truststore files inside the pod, but we can easily use configmap/secret with pem content.

Adding some codes in the project and converting cert files before connecting mqtt is not very difficult, but it is very important for users.

cdmikechen:
I've tested connecting mqtt by pem files, And I've found that some ssl config descriptions are not very accurate and may need to be proofread.

PR #1725, opened 2022-04-28
Add some SSL configuration descriptions

cdmikechen:
Linked https://github.com/smallrye/smallrye-reactive-messaging/issues/1722#issuecomment-1111699774
The description SSL pem configurations is not very clear, so I added some details.


cescoffier (approved):

cdmikechen:
@cescoffier 
Hi~ I've noticed that there is an error in checks, but I don't think this is caused by my changed.
Do I need to rebase my code? I'm new, so I don't know some operations related to do my pr, think may need your help.

cescoffier:
No worried @cdmikechen, we have a bit of instability.

PR #1747, opened 2022-05-19
Fix RabbitMQ connector issues

### LINKED ISSUE #1593: Smallrye-rabbitmq: when queue/exchange are lost they are not recreated until restarted
We had a problem with are rabbitmq cluster and it needed to be restarted. After the restart the queues we used were gone. The application automatically restored the connection. Only the rabbitmq tries to listen for messages on the queue but it is gone and gets a 404 from rabbitmq. It shutdown the channel en restarts again. The health endpoint says everything is fine. 

Steps to reproduce
- start rabbitmq
- start app (app has automatic-recovery-enabled but doesn't make al lot of difference)
- kill rabbitmq
- start rabbitmq
- App shows  continues errors (starting rabbitmq, closing rabbitmq, starting rabbitmq, closing rabbitmq, ....)

My expectation was that the app would recreate the queues/exchanges on a reconnect. Or that the health endpoint detects there is a problem (queues are gone) and gives a DOWN.

The vert.x documentation says the following about recovery:
https://vertx.io/docs/vertx-rabbitmq-client/java/#_recovery_and_reconnections
To use the method: "addConnectionEstablishedCallback" But it is not available in mutiny-rabbitmq.

### LINKED ISSUE #1742: Possibly useless code in RabbitMQConnector
@cescoffier Is this line not useless?

On completion is creates a cold `Multi` that is abandoned. It seems like this should be removed _or_ switched from `invoke` to a `transform`.

https://github.com/smallrye/smallrye-reactive-messaging/blob/d09f9f9706626586cc0521700c37c66f4444a5d3/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java#L483

### LINKED ISSUE #1743: RabbitMQ reconnection doesn't work
## Vert.x Reconnection Issues

Currently when Vert.x reconnect kicks in none of the exchanges or queues are rebuilt. The connector uses a mutiny stream to setup exchanges and queues which is never resubscribed because Vert.x reconnection is handled inside `RabbitMQClient`.

A solution to this seems to be to build the exchanges/queues in the `RabbitMQClient.addConnectionEstablishedCallback` as suggested by the [Vert.x RabbitMQClient docs](https://vertx.io/docs/vertx-rabbitmq-client/java/#_recovery_and_reconnections).

## RabbitMQConnector Retries Never Happen
`RabbitMQConnector` has its a retry strategy on the publisher stream. Unfortunately this never kicks in.

First, it uses the same `reconnect-attempts` configuration option as the Vert.x reconnection strategy uses. This means that if you try to disable the Vert.x reconnection strategy, it simultaneously disables the `RabbitMQConnector` retry strategy.

Second, the `RabbitMQConnector` retry strategy never retries anything. It is possible to provide a custom `RabbitMQOptions` bean that explicitly disables the Vert.x reconnection strategy while still providing a non-zero value for `reconnect-attempts`, and thus enables the `RabbitMQConnector` retry strategy. In this case the stream is never retried.

## Automatic Recovery
Enabling the RabbitMQ Java SDK's "Automatic Recovery" (while disabling the Vert.x reconnection strategy) seems to work but it does not connect with the health system to report the connection as down. Also, there is little to no logging to tell you what is happening in this scenario.

### LINKED ISSUE #1744: RabbitMQ Server Named Queues are not supported
RabbitMQ encourages the use of server named queues when clients are using an "exclusive" and "auto-deleting" queue.

To create a server named queue you must provide an empty string as the queue name. Currently (as of 3.16) providing a blank name for the queue name causes a configuration error.

### LINKED ISSUE #1749: RabbitMQ: Dead letter exchange is always configured with default
When configuring a queue, the dead letter exchange is configured, even though I did not configure it.
This is a problem for me, as I want to connect to an existing RabbitMQ (which I have no control over), where the queue might be already declared without the dead letter exchange.

Error message when trying to connect to the existing RabbitMQ:

```
Caused by: com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=406, reply-text=PRECONDITION_FAILED - inequivalent arg 'x-dead-letter-exchange' for queue 'incoming' in vhost '/': received the value 'DLX' of type 'longstr' but current is none, class-id=50, method-id=10)
```

I understand that `dead-letter-exchange` has a default value, but from my understanding it should only be applicable when `auto-bind-dlq` is set (which I did not set).

When declaring the queue via SmallRye on an empty RabbitMQ I can see in the GUI of RabbitMQ that the `x-dead-letter-exchange` is set to `DLX` (the default value). I think that this related to my problem, but I might be off track.

Using:
- smallrye-reactive-messaging-api 3.13.0
- Rabbit MQ 3.10 from docker image
- Java 11.0.9

This is my configuration
```
rabbitmq-host=localhost
 rabbitmq-port=5672 
rabbitmq-username=guest 
rabbitmq-password=guest  

mp.messaging.incoming.create-incident.connector=smallrye-rabbitmq
 mp.messaging.incoming.create-incident.exchange.name=incoming
mp.messaging.incoming.create-incident.exchange.durable=true 
mp.messaging.incoming.create-incident.exchange.type=topic 
mp.messaging.incoming.create-incident.queue.name=incoming
```

kdubb:
This address a number of issues in the RabbitMQ connector. Separate issues are addressed in separate commits.

Fixes #1743, Fixes #1593
* Uses `addConnectionEstablishedCallback` to build topology upon (re)connection.
* Removes "retry" logic added to publisher stream.
* * Already handled by either Vert.x reconnection or "Automatic Recovery"

Fixes #1744 
* Queues named `(server.auto)` are server generated name queues
* * `durable`, `auto-detect` and `exclusive` are ignored.

Fixes #1749 
* Dead letter exchange properties are only set if `auto-bind-dlq` is enabled

Fixes #1742 
* Code removed

Other Changes
* Sets the RabbitMQ connection name to channel name + incoming/outing for easier server debugging
* Reports connector as "connected" once all requested connections are connected and their consumers created
* * Debug logs each individual connection as it connects

### DIFF
--- a/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
+++ b/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
@@ -203,6 +203,15 @@ mp.messaging.incoming.people.queue.declare=true
     name of the exchange; if this is not the case you would need to name the
     exchange explicitly using the `exchange.name` property.
 
+!!!note
+    The connector supports RabbitMQ's "Server-named Queues" feature to create
+    an exclusive, auto-deleting, non-durable and randomly named queue. To
+    enable this feature you set the queue name to exactly `(server.auto)`.
+    Using this name not only enables the server named queue feature but also
+    automatically makes ths queue exclusive, auto-deleting, and non-durable;
+    therefore ignoring any values provided for the `exclusive`, `auto-delete`
+    and `durable` options.
+
 If you want RabbitMQ to create the `people` exchange, queue and binding,
 you need the following configuration:
 
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQClientHelper.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQClientHelper.java
@@ -65,11 +65,15 @@ public class RabbitMQClientHelper {
     static RabbitMQClient getClient(Vertx vertx, RabbitMQConnectorCommonConfiguration config,
             Instance<CredentialsProvider> credentialsProviders) {
         try {
+            String connectionName = String.format("%s (%s)",
+                    config.getChannel(),
+                    config instanceof RabbitMQConnectorIncomingConfiguration ? "Incoming" : "Outgoing");
             String host = config.getHost();
             int port = config.getPort();
             log.brokerConfigured(host, port, config.getChannel());
 
             RabbitMQOptions options = new RabbitMQOptions()
+                    .setConnectionName(connectionName)
                     .setHost(host)
                     .setPort(port)
                     .setSsl(config.getSsl())
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
@@ -217,7 +217,7 @@ public class RabbitMQConnector implements IncomingConnectorFactory, OutgoingConn
 
         final Integer connectionCount = ic.getConnectionCount();
         Multi<? extends Message<?>> multi = Multi.createFrom().range(0, connectionCount)
-                .flatMap(ignore -> {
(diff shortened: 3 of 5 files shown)


kdubb:
@cescoffier I also have a branch with the same changes against the `3.15.0` version. The reconnection is a big enough issue that I think a point release should be considered. As a Quarkus 2.8/2.9 user, we're pretty screwed by these issues.

cescoffier (approved):

PR #1748, opened 2022-05-19
Add QOS support to RabbitMQ incoming connector

kdubb:
The options is named `max-outstanding-messages` as it seems a lot more descriptive than RabbitMQ's "QOS" terminology.


smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
```diff
@@ -271,10 +272,20 @@ public PublisherBuilder<? extends Message<?>> getPublisherBuilder(final Config c
     }
 
     private Uni<RabbitMQConsumer> createConsumer(RabbitMQConnectorIncomingConfiguration ic, RabbitMQClient client) {
-        return client.basicConsumer(ic.getQueueName(), new QueueOptions()
-                .setAutoAck(ic.getAutoAcknowledgement())
-                .setMaxInternalQueueSize(ic.getMaxIncomingInternalQueueSize())
-                .setKeepMostRecent(ic.getKeepMostRecent()));
+        return Uni.createFrom().nullItem()
```
cescoffier, line 275: Any reason for that uni? Can't you start the chain with the next one?

cescoffier (changes requested):

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
kdubb: It looked better to me than starting with an if/else, which meant I had to store the stream in a variable and start with that

cescoffier (approved):

cescoffier:
There is a conflict (due to the other PR I guess)

Issue #1162, opened 2021-04-17
Support intercepting and converting outgoing messages


vladykin:
There a several cross-cutting concerns associated with emitting messages to outgoing channel:
* serializing domain objects to wire format
* logging the message that is about to be sent (and the result of sending)
* updating metrics

Having to write this stuff in all places of application where something is emitted or returned from an @outgoing-annotated method is inconvenient. It would be very handy to have an interceptor that would process all outgoing messages and be able to log/count/convert/etc them.

For incoming messages there is MessageConverter which can be used for this purpose. For outgoing messages, AFAIK, these is nothing like that.

P.S. I'm using smallrye-reactive-messaging-amqp 2.9 with Quarkus 1.13.1.

cescoffier:
That's an interesting idea. One thing is that AMQP is a bit behind in term of feature (in comparison to Kafka which have been a priority for the last few months). I have a few questions (mostly to collect ideas):

> serializing domain objects to wire format

That's generally protocol specific? Typically in Kafka you can configure your serializer.  The AMQP protocol is strict in terms of type (AMQP Type system), however, for object (where we map the object into JSON at the moment) we can imagine a hook that would be something like:

```
public <T> byte[] serialize(T instance, OutgoingAmqpMetadata metadata);
```

Implementations would be able to serialize the instance as they want (producing a byte[]) as well as updating the outgoing message metadata (typically the content-type header).

I guess the HTTP connector can also benefit from such an interceptor (@michalszynkiewicz)


> logging the message that is about to be sent (and the result of sending)

That's definitely something we should add. However, the API is going to be protocol-specific. In Kafka, we will call the listeners with a Record (going to be sent) and RecordMetadata (result). HTTP would use something like an HTTP Request and an HTTP Response. 

The question is do we need multiple hooks or a single one receiving what has been sent and the result. Typically, what about:

```java
public <I, O> void onSuccess(I sent, O response);
public <I, O> void onFailure(I sent, Exception e);
```

Note that the `onFailure` does not allow retry or anything. This is already handled by the connector.

> updating metrics

We already expose metrics (at least for Kafka). We should add the AMQP metrics soon.

Ladicek:
Is this perhaps related to #1141? (If so, I guess we should keep this one open, as it has more details/ideas. I'd close #1141 as duplicate.)

vladykin:
>> serializing domain objects to wire format

[...]

PR #1772, opened 2022-06-16
Deprecate and replace MessageConverter with reactive IncomingMessageConverter

### LINKED ISSUE #1769: `MessageConverter` failures stall processing indefinitely with manual acknowledgment
If a `MessageConverter.convert(...)` method throw an exception the incoming channel is stalled indefinitely; exceptions are logged but nothing else happens.

Since the failure is in message conversion there is no chance for user code to ack or nack the messages so it always hangs.

This is being experienced specifically with RabbitMQ but it appears this might affect any broker.

kdubb:
Fixes #1769

## Deprecate `MesasgeConverter`
`MessageConverter`'s interface is synchronous which does not allow it to handle failures in any meaningful way. First, if message conversion fails the interface requires it to return _something_ which cannot currently be null; meaning there is no current mechanism to "ignore" a failed message. Second, it is required that `MessageConverter`s handle calling `ack` or `nack` for a failure. Unfortunately those methods return `CompletionStage`s. Converters must ignore the results of `ack`/`nack` or perform some sort of blocking wait for them to complete.

After talking with @cescoffier we decided changing the interface to use a reactive method would be best.
 
## Introducing IncomingMessageConverter
`IncomingMessageConverter` is the same interface but returns a `Uni<Message<?>>`. Changing its name serves two purposes. 1) The new name allows/prepares for outgoing message converters (as discussed in issues like #1162). 2) It allows continued use of (now deprecated) `MessageConverter` until such time as it is removed.

### Adapter AnyMessageConverter
To facilitate using both `IncomingMessageConverter` and `MessageConverter` a new base interface `AnyMessageConverter` was added. It is marked deprecated along with `MessageConverter` and when it is decided to remove `MessageConverter`, `AnyMessageConverter` will be removed as well.

### Wrapping Legacy Converters
Calls to the deprecated `MessageConverter.convert` are now wrapped with failure handling that logs the failure, rejects (aka `nack`s) and ignores the message. This solves issues with stalling of message handling (e.g. #1769) since failures cannot be handled by these converters.

Wrapping is not done for `IncomingMessageConverter`, now that the interface provides the full capabilities to handle failures gracefully it is expected that they will. All failures from streams returned from `IncomingMessageConverter` are ignored without logging or ack/nack (this is explained in the interface's API docs).

### `canConvert`
Previously due to caching of converters, the `canCovert` method was only ever called one time, when the converter was cached. This has the following problems:
1) The interface's API docs say that it will be called every time before `convert` is called.
2) The call to `canConvert` is passed an instance of `Message<?>` to make its decision but each call to `convert` will be passed a unique instance of `Message<?>` which may not be anything like the message passed to `canCovert`. 

For these reasons `canConvert` is now always called (as the API docs already suggest) even after converter caching. If a converter cannot be used with the current message instance a new converter will be cached. For converters that see the same exact messages all the time there is little to no penalty. Additionally converters that may see a wide range of message types will now work correctly.

### No `Identity` converter
The `ConverterUtils.convert` method checks for the ability to assign the payload type in the message to the type target payload type required by the injection target. There is no need to lookup and use an identity converter since the message payload is already in a form the injection target can use.

### RabbitMQ Conversion
`IncomingRabbitMQMessage`'s ad-hoc conversion in `IncomingRabbitMQMessage.getPayload()` has been replaced with a standard set of converts that replicate the previous behavior but with better error handling and more deterministic logic.

Additionally this removes the `content-type-override` for RabbitMQ incoming connectors. It was used as a hack to make conversion happen on unsupported content types.  This should be done in an `IncomingMessageConverter`, with proper error handling, whenever this functionality is needed.


kdubb:
@cescoffier As discussed in the Zulip chat. There are a 5 or 6 failing Kafka tests that seem to be a result of the now required `onItem().transformtoUniAndConcatenate()` used during conversion.

ozangunalp:
Closing this as improvements were brought in with #1837 and #1856

urbim:
@ozangunalp The mentioned PRs still don't fix the original issue #1769. I tested with ActiveMQ and Quarkus version 3.12.3 and the consumer still stops receiving all messages on any exception.

matejbizjak:
@ozangunalp I can also confirm that this issue still persists. We stop receiving all messages after any exception. Tested with ActiveMQ and Quarkus 3.16.4.

PR #1774, opened 2022-06-17
Only allow a single ack or nack to be sent to RabbitMQ per message

### LINKED ISSUE #1770: Calling `Message.ack` or `Message.nack` more than once causes RabbitMQClient restart
For RabbitMQ you are only allowed to call one of `ack` or `nack` once. If this mistake happens it causes a complete restart of the RabbitMQ client without any notice as to why it is happening.

Detecting this and logging the problem would be very helpful for developers.

kdubb:
Fixes #1770

Currently if the calls ack or nack more than a combine total of 1 time the client will send all the acks or nacks to the server. Since RabbitMQ explicitly disallows this, it causes a connection error which then causes a complete connection restart.  The worst part is that without some real examination you cannot tell this is from calling ack and/or nack multiple times.

The best example of this is when a user `nacks` a message in a message handler that uses the `POST_PROCESSING` strategy. The framework will then `ack` or `nack` the message based on the outcome of the message handler. This will cause the connection restart as described above. No logging or hints exist to tell users that this is what is causing connection restarts.

This commit replaces the ack and nack handlers, after the first call of either, with versions that are no-ops; silently ignoring any attempt to ack or nack after the first one.

Originally I had detected the error and reported it, but after reading the Eclipse MicroProfile Reactive Messaging Specification it appears there are scenarios where both the user and the framework might be expected to handle ack/nack. Ignoring duplicates felt like an easier system to use.

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
@@ -7,6 +7,7 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


kdubb:
Alternatively, if there is good reason too (maybe the spec. allows this?) we could just translate these into no-ops.

I am thinking of the example I gave above. If the use doesn't want to worry about it in a `POST_PROCESSING` handler except in one specific case down a few stack levels, it might be good to allow them to force it in that specific case and have default handling in all others.

🤔

cescoffier (approved):

PR #1800, opened 2022-07-07
Switch the RS TCK JUnit5 shim to that of Mutiny Zero

jponge:
The module has been transfered from Mutiny to Mutiny Zero.


jponge:
Oh, it's compiled with Java 11 🤣

cescoffier (approved):

PR #1806, opened 2022-07-12
Convert IncomingRabbitMQMessage payload eagerly

ozangunalp:
Default payload Conversion supports as before `String`, Vert.x `JsonObject`, Vert.x `Buffer` and `byte[]`. If conversions fail it falls back to the `byte[]`. 

This prevents the pipeline from failing when MessageConverter is applied on a RabbitMQ message with a _faulty_ payload. Secondly it avoids calling the conversion on each call to `getPayload`

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
@@ -58,6 +58,7 @@ public class IncomingRabbitMQMessage<T> implements ContextAwareMessage<T> {
     private RabbitMQFailureHandler onNack;
     private RabbitMQAckHandler onAck;
     private final String contentTypeOverride;
+    private final T payload;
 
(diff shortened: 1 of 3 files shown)


smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/serde/KeyDeserializerConfigurationTest.java
```diff
@@ -12,16 +12,6 @@
 import javax.enterprise.inject.AmbiguousResolutionException;
 import javax.enterprise.inject.UnsatisfiedResolutionException;
 
-import io.smallrye.reactive.messaging.kafka.CountKafkaCdiEvents;
```
cescoffier, line 15: Ashy so many import reorder?

cescoffier (changes requested):
A bit concerned with the imports. The rest looks good.

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/serde/KeyDeserializerConfigurationTest.java
ozangunalp: I'll double-check but I think it comes from a previously merged PR.

cescoffier (approved):

PR #1826, opened 2022-07-26
Switch to java.util.concurrent.Flow APIs

jponge:

PR #1835, opened 2022-08-03
Use adapters between RS and Flow APIs

jponge:
This does not migrate to Flow but allows using Flow-based Multi.

This compiles but does not work, there are issues to fix with the core of RM.


jponge:
This is an alternative to #1826 where we use adaptation. This compiles but does not work due to class cast issues.

ozangunalp:
Closing this in favour of #1896

Issue #1782, opened 2022-06-28
Support Queue type and mode configuration in rabbitmq


ozangunalp:
From https://github.com/quarkusio/quarkus/issues/25754

> add an option to set missing properties for queue definition through the application configuration. Support:
> 
> x-queue-type: [quorum, classic, stream]
> x-queue-mode: [lazy, default]

kdubb:
I've tried to implement this but I'm stuck up against SmallRye config. As I said in Quarkus#25754 this should be implemented as a freeform set of options. There are lots of options for queues and lots of options for exchanges. Not to mention RabbitMQ plugins can add more options to each.

Unfortunately after I implemented this trying to use a `java.util.Map` as the `type` for `ConnectionAttribute` SmallRye config complains `SRCFG00013: No Converter registered for interface java.util.Map`.

Is there any way around this?

ozangunalp:
@kdubb I am sorry, your comment went totally under my radar. 
IIRC it is not currently possible to use map as a connector attribute.
It is possible to extract free-form config from the channel configuration, we are doing this in the Kafka connector. 
We can pass config prefixed with `queue` to queue declare and build on what @cdmikechen did on #1836. 
I can give it a shot tomorrow.

ozangunalp:
Closed by #1836

PR #1836, opened 2022-08-04
Support RabbitMQ x-queue-type and x-queue-mode

cdmikechen:
The main purpose of this PR is to support `x-queue-typ` and `x-queue-mode` in rabbitmq. 
I have found that relevant issues have been raised before: https://github.com/smallrye/smallrye-reactive-messaging/issues/1782 and https://github.com/quarkusio/quarkus/issues/25754

@cescoffier Hi~ Sorry to disturb you.
I have added some test cases and tested them locally. Do I need to explain this problem in the email list for discussion?

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
@@ -102,6 +102,8 @@ import io.vertx.rabbitmq.RabbitMQPublisherOptions;
 @ConnectorAttribute(name = "queue.declare", direction = INCOMING, description = "Whether to declare the queue and binding; set to false if these are expected to be set up independently", type = "boolean", defaultValue = "true")
 @ConnectorAttribute(name = "queue.ttl", direction = INCOMING, description = "If specified, the time (ms) for which a message can remain in the queue undelivered before it is dead", type = "long")
 @ConnectorAttribute(name = "queue.single-active-consumer", direction = INCOMING, description = "If set to true, only one consumer can actively consume messages", type = "boolean")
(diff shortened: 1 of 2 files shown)


cescoffier:
I'm not a RabbitMQ expert, @kdubb can you have a look?

kdubb:
As I've said in the referenced issues, I would like to see this implemented as freeform parameters; there are already too many to add manually and with RMQ plugins It becomes impossible.

I've got my own version of the PR but I ran into an issue with SR Config not supporting freeform maps or properties.

@ozangunalp Said there is some way to handle freeform properties and was going to tackle it but that's where it was left off.

cescoffier:
Reactive messaging does not support freeform attributes. So, except if you define your own format (like some json blurb) or use something like we did in kafka (take into account all the attributes with a prefix) I don't see how to implement it. 

Would the latter be an option?

kdubb:
I think the latter is what @ozangunalp was referring to. An explanation of this technique would help move forward on this.

cescoffier:
it's relatively simple. Instead of using the connector attributes, you retrieve the full MicroProfile Config, iterate over the key, and select the values that match your prefix (`mp.messaging.incoming|outgoing.$channel.queue-args.bla=foo`)

However:

- these attributes won't be documented automatically - so you need to document them
- you won't have the type support

cdmikechen:
@cescoffier @kdubb 
In my opinion, these properties are variables that rabbitmq has explicit configuration instructions and definitions for. Only a small number of these variables are defined in case of big version changes. 
If we can declare them exactly in the configs, it will be easier for users to use and locate them, instead of doing a one-time map in a more generic way (maybe hard to use).

ozangunalp:
Sorry for my late reply. First of all, I think we should take this in, we already do support `x-message-ttl` and `x-single-active-consumer`.

I proposed to have free form config for queue args because IIUC "x-arguments" depend on rabbitmq plugins. 
I can collect a list of remaining common args such as `x-max-length-bytes`, `x-max-length`, `x-overflow`, `x-max-priority`, `x-expires`. 

We may not need to add a near-exhaustive list of x-args, as I understand, broker policies are the recommended way of assigning arguments to queues and exchanges.

ozangunalp (approved):

PR #1837, opened 2022-08-04
Mutiny 1.7.0 with no prefetch concatmap

ozangunalp:
- Bump mutiny to 1.7.0
- The hack needed for avoiding buffering is no longer needed
- Messages nacked when MessageConverter throws exception

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
 
     <micrometer.version>1.9.2</micrometer.version>
 
-    <mutiny.version>1.6.0</mutiny.version>
+    <mutiny.version>1.7.0</mutiny.version>
     <artemis.version>2.19.0</artemis.version>
 
     <jboss-log-manager.version>2.1.18.Final</jboss-log-manager.version>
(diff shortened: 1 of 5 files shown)


ozangunalp:
@kdubb for `MessageConverter` failure handling

kdubb:
@ozangunalp This converter function still caches the last used converter. This is wrong, ead my original PR for details.

ozangunalp:
@kdubb I understand that the documentation doesn't say that the canConvert function is called once on lookup. But I don't think that the current implementation is wrong per se. When reactive messaging looks for a converter for a specific injection point, it picks the first converter that can convert. Note that converters are cabled separately on each injection point, depending on the target type of the injection point. It is very unlikely that the converter will change mid-consumption for a given channel-injection target pair.

Nevertheless I'll keep this PR separate from the MessageConverter's change.

kdubb:
@ozangunalp The issue is the incoming messages changing. You cal `canConvert` on one instance and assume every instance to follow is exactly the same format.

The performance penalty to call `canConvert` first seems __very__ negligible. Messages that are formatted identically will respond yes and no lookup happens.

ozangunalp:
I understand that the penalty is negligible. I think `MessageConverter` was designed to adapt the message to the injection point type, rather than to parse the payload. 

If you don't mind, I'd like to keep your contribution for RabbitMQ converters in #1772 but adapt it to roll back the `IncomingMessageConverter` change. The message will still be nack'ed if the conversion fails.

I've prepared another PR for improving message interceptors, including outgoing messages.

cescoffier (approved):

cescoffier:
This PR should not touch the converter (and converters are cached by design).

PR #1849, opened 2022-08-15
Ignore null-values in headers

### LINKED ISSUE #1848: Kafka - NPE with CloudEvents enabled and a null header value
When consuming a Kafka record with CloudEvents enabled the current implementation in the KafkaCloudEventHelper breaks when a header with no value is present.

The problematic code is:
```
        // Build a map containing all the headers
        // We remove the entry at each access to filter out extension attribute.
        Map<String, String> headers = new HashMap<>();
        record.headers().forEach(kh -> {
            String key = kh.key();
            String value = new String(kh.value(), StandardCharsets.UTF_8); // Rules 3.2.3 - Force UTF-8
            headers.put(key, value);
        });
```
`new String(null, UTF-8)` causes the problem.

lostiniceland:
fixes #1848

Null values in a header-value can break the UTF-8 String conversion, causing the Channel to be unhealthy.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ce/KafkaCloudEventHelper.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ce/KafkaCloudEventHelper.java
@@ -135,13 +135,16 @@ public class KafkaCloudEventHelper {
             ConsumerRecord<?, T> record) {
         DefaultCloudEventMetadataBuilder<T> builder = new DefaultCloudEventMetadataBuilder<>();
 
-        // Build a map containing all the headers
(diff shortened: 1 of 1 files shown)


cescoffier (approved):

cescoffier:
Pretty sure the AMQP connector has the same flaw (pretty nice catch BTW!)

lostiniceland:
> Pretty sure the AMQP connector has the same flaw

I did a quick check there before I started since I suspected as much, but there, the implementation just reads from the main map and removes the CloudEvent-specific ones at the end. No UTF-8 conversion takes place in the AmqpCloudEventHelper...maybe something to consider as well.

Regarding the failing tests, I have no idea why this happens. Locally on 17 `mvn clean install` on the Kafka-Connector runs fine, but on the reactor another tests fails. Have to dig deeper into the Maven setup

cescoffier:
Thanks!

Issue #1784, opened 2022-06-28
OpenTelemetry support enhancement


cescoffier:
In order to improve the OpenTelemetry tracing support in reactive messaging we need:

1. Introduce an SPI that allows _decorating_ messages emitted by emitters. This SPI would be called on the same thread as the one calling `emitter.send(...)`, and the SPI implementor should be able to extract the current OTel Context and create the `TracingMetadata`. This metadata will be added to the emitted message.
2. When a message is associated with a duplicated context (the message could be created from an emitter or from an inbound connector), an SPI should be created to extract the potential `TracingMetadata` and attach it to the duplicated context. This would allow OTel to _re-attach_. 

In terms of ITs, there are already existing tracing-related IT, which verify the propagation along a pipeline and within a single protocol (like Kafka).
In addition to this, we need to check that:

- when emitting a message using an emitter, existing tracing details are propagated
- when emitting a payload using an emitter, existing tracing details are propagated
- when receiving a message from a protocol supporting tracing, the tracing details are propagated and associated with the duplicated context
- OTel can _re-attach_ from the duplicated context

\CC @radcortez @brunobat @ozangunalp

cescoffier:
@ozangunalp @radcortez, where are we on this one?

radcortez:
I believe that https://github.com/smallrye/smallrye-reactive-messaging/pull/1678 takes care of this.

ozangunalp:
Closing this as #1678 takes care of it.

Issue #1141, opened 2021-03-31
message interceptors


Ladicek:
It would be useful to have a notion of "message interceptors" for cases when you want to apply some cross-cuting logic to all messages on a number of channels (one, more than one, all). I hear from @cescoffier that this was discussed in MP Reactive Messaging, but couldn't find an issue in https://github.com/eclipse/microprofile-reactive-messaging/issues We could perhaps discuss the idea here and build a prototype in SmallRye.

Based on this discussion: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Use.20RequestScoped-bean.20in.20Kafka-Interceptor

cescoffier:
That was related to https://github.com/eclipse/microprofile-reactive-messaging/issues/72.

wicksim:
Is there any progress on this one? We are still looking for a solution for our problem discussed in https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Use.20RequestScoped-bean.20in.20Kafka-Interceptor. A solution with a CDI-Decorator does not seem to work, maybe because of the needed `@Channel`-annotation. Decorator is just not used.

So I'm still searching for a solution to intercept sending/receiving of Kafka-Messages in a way that I can read some values out of a `@RequestScoped` CDI-bean.

ozangunalp:
Initial implementation by #1856, to improve if in need

PR #1856, opened 2022-08-26
Decorator API for Publisher & Subscribers, Outgoing Message Interceptors

ozangunalp:
There have been previous discussions on message interceptors #1141 

- Using CDI interceptors for Reactive Messaging is not possible as consuming methods callbacks or stream processors are not called through CDI.
- Using CDI decorators was discussed for intercepting calls to Emitters.
- Another use case is to intercept outgoing messages for adapting outgoing payload and intercepting transmission result #1162.
- Interceptors need a way to access some sort of call context: i.e incoming message trace or `@RequestScoped` beans. More recently there is a need to improve OpenTelemetry support by propagating traces through RM streams #1784.

An internal `PublisherDecorator` extension point already exists for incoming channels, which is used for metrics and Vert.x context association.

This PR
- Moves the existing `PublisherDecorator` to the API module
- Adds a `SubscriberDecorator` for decorating `Multi`s that are sources for subscribers in the channel graph.
- Both decorators are prioritized and the graph wiring will call them from highest to lowest priority.
- They receive the channel name and `isConnector` flag for distinguishing channels bound to a connector.
- Introduces a default implementation of `SubscriberDecorator` which discovers `OutgoingInterceptor` beans for attaching interceptors to channels. 
- `OutgoingInterceptor`s can alter outgoing messages with `onMessage` callback, and when supported by the connector (Kafka and MQTT) can access transmission results with`onMessageAck` callback.


Ladicek:
I didn't review properly, but I'd suggest to avoid extending `Prioritized`. Instead, users should annotate an interceptor with `@Priority`. That's consistent with how CDI interceptors are declared. Also,  invocation order of interceptors should be from lowest priority value to highest priority value, to be consistent with CDI interceptors (see https://jakarta.ee/specifications/interceptors/2.0/interceptors-spec-2.0.html#use_of_the_priority_annotation_in_ordering_interceptors).

ozangunalp:
During my tests I couldn't reproduce CDI ordering the instances by default from the `@Inject @Any Instance<A> instances` injection. 
I changed the doc to match the implementation, `PublisherDecorator`s, `SubscriberDecorator`s and `OutgoingInterceptor`s invoked ordered from highest priority to lowest (from the least priority value to the greatest). It is the case for `MessageConverter`s, I added a clarification to the documentation as well. 

I see that `WeldInstance` does come with an instance comparator taking into account both `Prioritized` and `@Priority`. But we can't use it. Therefore for the sake of simplicity I am keeping only the `Prioritized` to describe the bean priority.

Ladicek:
Ah, right, iterating through `Instance` doesn't consider priority. Weld has proprietary means to do so, and ArC does guarantee order, but that's also a proprietary extension. You can use CDI mechanisms directly only when you need a single instance out of multiple. Oh well.

cescoffier (approved):
LGTM!

Issue #1868, opened 2022-09-05
MQTT should support shared subscription feature


cdmikechen:
Based on the new shared subscription feature of MQTT, we should also need to support this attribute in the topic, as the current topic processing filters out topic data with a `$share/GROUP/TOPIC_NAME` prefix.
After my testing, vertx is now able to get messages prefixed with `$share/GROUP`.

Link:
https://www.emqx.com/en/blog/introduction-to-mqtt5-protocol-shared-subscription
https://www.hivemq.com/blog/mqtt5-essentials-part7-shared-subscriptions/

PR #1869, opened 2022-09-05
MQTT support shared subscription feature

cdmikechen:
Issue link https://github.com/smallrye/smallrye-reactive-messaging/issues/1868
I think this feature is very useful when deploying multiple pods with quarkus/smallrye to avoid duplicate message consumption problems.


cescoffier (approved):

cescoffier:
Thanks!

PR #1900, opened 2022-09-18
Update the microprofile reactive messaging  dependency to 3.0

cescoffier:


cescoffier:
@ozangunalp, for some weird reason, revapi is complaining... we would need to have a look before the release.

ozangunalp (approved):

ozangunalp:
@cescoffier I pushed a diff to ignore revapi changes.

cescoffier:
@ozangunalp Thanks!  Yes, I don't understand that "breakage". These methods have been added months ago. MAybe the revapi baseline must be updated.

ozangunalp:
Because we have a 4.0.0.RC1 release, next 4.x will be compared to that release. I think ignoring is safer, as with jakarta release we'll have bunch of changes.

PR #1909, opened 2022-09-21
Kafka client.id prefix attribute

ozangunalp:
- Adding an optional incoming/outgoing channel attribute `client-id-prefix`.
- If no `client.id` is configured, generate it using the channel name, using the default prefix "kafka-consumer-" (current behaviour)
- If a `client.id` is configured, add the prefix with default value "" (empty)
- For consumer's currently appended suffix per partition consumer continues to work.


cescoffier (approved):

mrickly:
@ozangunalp : Thank you for your implementation!

PR #1926, opened 2022-10-03
Support tracing metadata with cloud events

### LINKED ISSUE #1888: 0   In our Quarkus based microservices using Smallrye Kafka for event streaming, we have implemented tracing using OpenTracing. Now, we're trying to propagate the tracing context between these services using io.smallrye.reactive.messaging.ce.OutgoingCloudEventMetadataBuilder
In our Quarkus based microservices using Smallrye Kafka for event streaming, we have implemented tracing using OpenTracing. Now, we're trying to propagate the tracing context between these services using io.smallrye.reactive.messaging.ce.OutgoingCloudEventMetadataBuilder:

public class CloudEventBuilder<T> {

  private static final String DATA_CONTENT_TYPE = "application/json";
  private TracingMetadata tracingMetadata;
  private T payload;

  private final OutgoingCloudEventMetadataBuilder<T> metadataBuilder =
      OutgoingCloudEventMetadata.builder();

  public CloudEventBuilder() {
    metadataBuilder
        .withSource(URI.create("my-event-source"))
        .withType("MyEvent")
        .withDataContentType(DATA_CONTENT_TYPE);
  }

  public CloudEventBuilder<T> withId(String id) {
    metadataBuilder.withId(id);
    return this;
  }

  public CloudEventBuilder<T> withPayload(T payload) {
    this.payload  = payload;
    return this;
  }

  public CloudEventBuilder<T> withTracingMetadata(TracingMetadata tracingMetadata) {
    this.tracingMetadata = tracingMetadata;
    return this;
  }

  public Message<T> build() {
    return Message.of(payload, Metadata.of(metadataBuilder.build(), tracingMetadata));
  }

}

However, our finding is that the tracing context is not being propagated in this way. On the producing side we see TracingMetadata including trace id's and spans being added to the cloud event, but in the consumer, that TracingMetadata seems to be overwritten by new trace id's and spans. Traces from the producer are lost.

Surprisingly, we found that tracing headers were propagated succesfully when omitting the io.smallrye.reactive.messaging.ce.OutgoingCloudEventMetadataBuilder. Using Messages, traces are propagated as expected and TracingMetadata had the same trace id on both sides:

@ApplicationScoped
public class KafkaService {

  @Channel("myevents")
  Emitter<MyEvent> emitter;

  @Inject TracedService tracedService;

  @WithSpan("myspan")
  public void send() {

    var myEvent = new MyEvent(45);
    var message = Message.of(myEvent);
    var msgWithMetadata = message.addMetadata(TracingMetadata.withCurrent(Context.current()));

    emitter.send(msgWithMetadata);
  }

  @WithSpan("myspan2")
  public void send2() {

    var myEvent = new MyEvent(45);
    var message = Message.of(myEvent);
    var msgWithMetadata = message.addMetadata(TracingMetadata.withCurrent(Context.current()));

    Message<MyEvent> ceEventMessage =
        new CloudEventBuilder<MyEvent>()
            .withPayload(myEvent)
            .withId("12345")
            .withTracingMetadata(TracingMetadata.withCurrent(Context.current()))
            .build();

    emitter.send(ceEventMessage);
  }
}

When sending messages using send() the tracing context is propagated as expected, but when using send2(), tracing information from the producer is lost.

Since we need to implement propagation of traces in a generic way, we want to use io.smallrye.reactive.messaging.ce.OutgoingCloudEventMetadataBuilder. Could you help us getting the propagation of tracing context to work? Here's a reproducer for this problem: https://github.com/ewelling/reproduce-failing-traces-propagation. Please check README.md for steps to reproduce. Thanks in advance!

ozangunalp:
Fixes #1888

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
@@ -1,15 +1,28 @@
 package io.smallrye.reactive.messaging.kafka.impl;
 
+import static io.smallrye.reactive.messaging.kafka.KafkaConnector.TRACER;
+
 import java.util.Arrays;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeaders;
+import org.eclipse.microprofile.reactive.messaging.Message;
 
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import io.smallrye.reactive.messaging.TracingMetadata;
 import io.smallrye.reactive.messaging.kafka.api.IncomingKafkaRecordMetadata;
 import io.smallrye.reactive.messaging.kafka.api.OutgoingKafkaRecordMetadata;
+import io.smallrye.reactive.messaging.kafka.tracing.HeaderInjectAdapter;
 
 public class KafkaRecordHelper {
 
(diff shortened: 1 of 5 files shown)


cescoffier (approved):
It looks a bit dirty in the sense it mixes things that should have been separated.

I understand the reason, but, it makes cloud events one step closer to soap: everything we do would need some kind of integration.

ozangunalp:
In this particular case, there is no integration investment: we didn't even call tracing when handling cloud events.

cescoffier:
There is:
```
        if (configuration.getTracingEnabled()) {
            KafkaRecordHelper.createOutgoingTrace(message, topic, partition, headers);
        }
```

which basically means that every time we introduce something using headers or metadata, we need to update the CloudEvent helper.

PR #1928, opened 2022-10-03
Add cause exception type to DLQ headers

### LINKED ISSUE #1913: Feature request for DLQ strategy.
As per my conversation with @cescoffier here: https://quarkusio.zulipchat.com/#narrow/stream/294206-smallrye/topic/Suggestion.20for.20DLQ.20strategy.2E/near/300863081

Currently when a message is DLQ, the headers contain the following info as an example:
```
{
	"dead-letter-reason": "Connection refused: localhost/127.0.0.1:18080",
	"dead-letter-offset": "98",
	"dead-letter-partition": "124",
	"dead-letter-cause": "Connection refused",
	"dead-letter-topic": "webhook-push-dlr-dev"
}
```

When creating consumers to read the dead letter topic and you want to create some sort of processing logic, you need to compare string messages. This is isn't the most ideal situation because you need to compare string text and you also need to know the exact text coming from the the exception.

That said it would be nice to have an additional header, with the class name of the exception. This makes it easier/safer to compare and also deal with.

```
{
	"dead-letter-exception": "com.java.net.ConnectionException.class",
	"dead-letter-reason": "Connection refused: localhost/127.0.0.1:18080",
	"dead-letter-offset": "98",
	"dead-letter-partition": "124",
	"dead-letter-cause": "Connection refused",
	"dead-letter-topic": "webhook-push-dlr-dev"
}

```

Which means now we can do this:
```
if("com.java.net.ConnectionException".equals(Headers.get("dead-letter-exception"))
```
or
```
if("com.abc.MyCustomRetriableException".equals(Headers.get("dead-letter-exception"))
```
instead of
```
if("Connection refused".equalsIgnoreCase(Headers.get("dead-letter-cause"))
```

ozangunalp:
Closes #1913

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
@@ -35,6 +35,9 @@ import io.vertx.mutiny.core.Vertx;
 @SuppressWarnings({ "rawtypes", "unchecked" })
 public class KafkaDeadLetterQueue implements KafkaFailureHandler {
 
+    public static final String DEAD_LETTER_EXCEPTION_CLASS_NAME = "dead-letter-exception-class-name";
+    public static final String DEAD_LETTER_CAUSE_CLASS_NAME = "dead-letter-cause-class-name";
+
     public static final String DEAD_LETTER_REASON = "dead-letter-reason";
     public static final String DEAD_LETTER_CAUSE = "dead-letter-cause";
     public static final String DEAD_LETTER_TOPIC = "dead-letter-topic";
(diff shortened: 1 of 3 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
```diff
@@ -35,6 +35,7 @@
 @SuppressWarnings({ "rawtypes", "unchecked" })
 public class KafkaDeadLetterQueue implements KafkaFailureHandler {
 
+    public static final String DEAD_LETTER_EXCEPTION = "dead-letter-exception";
```
cescoffier, line 38: I would name it exception-class-name. That would avoid the confusion between this one, the "reason" and the "cause". Actually, adding "cause-class-name" might be useful too.

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/fault/KafkaDeadLetterQueue.java
ozangunalp: Done

PR #1937, opened 2022-10-06
Use package private field injection

geoand:
This is done in order to avoid warnings in Quarkus (which does support private field injection but needs to use reflection)

Relates to: https://github.com/quarkusio/quarkus/issues/28420


cescoffier (approved):

ozangunalp:
That's great thanks @geoand !

geoand:
YW!

Issue #1939, opened 2022-10-07
MQTT and CleanSession=false and Unsubscribe on shutdown 


dometec:
Using MQTT and "auto-clean-session=false" means that I want leave a session with my subscription live on the broker even when my application is down. 
On reconnection phase, if the session is present on the broker side, the application does not send subscription requests (given that these should already be present on the broker).

The problem is that when the application is shut down, this sends UNSUBSCRIBE request, so then at the next start (since the session is present though without subscriptions) the app didn't send subscriptions, so it does not receive messages.

The UNSUBSCRIBE must never be done, since for clean session = true, on disconnection the broker remove all subscription independently, while for clean session = false, I just want the SUBSCRIPTION always.

As discussed with @cescoffier, it is better to add a variable to allow restore this old behavior in case someone needs it:
_unsubscribe-on-disconnection_ (default false)

cescoffier:
Fixed by https://github.com/smallrye/smallrye-reactive-messaging/pull/1940.

PR #1940, opened 2022-10-07
Fix MQTT message lost when using Persistent Session - Issues  #1935 and #1939

### LINKED ISSUE #1935: MQTT, Reconnection and CleanSession=false
Using CleanSession=false, If there is a reconnection for some network reason, the client does not receive any more messages.

If the client reconnects to the server and its response contains the flag Session Present = 0, the client does not send the subscribe request (ok, the subscription must be present on the broker side), but the client code is not notified when the broker sends a publish. This is because the subscribeCompleted(MqttSubAckMessage ack) method is never called (since we have not sent the subscribe message) and the topic on subscriptionStates remains UNSUBSCRIBED.

dometec:
This commit fixes the two issues on the subject. 

I also added a note on documentation about persistent sessions.

I'm sorry, I tried to implement a test but I have not found a solution to disconnect the client without restarting it.


cescoffier (approved):
Thanks!

PR #1946, opened 2022-10-14
RabbitMQ - Send to default exchange

### LINKED ISSUE #1641: smallrye-rabbitmq: Can't publish directly to queue with amq.default (or empty) exchange
The rabbitmq allow directly publish using the default exchange (amq.default)

image

I configure my quarkus app with the following values:

```propertie
mp.messaging.outgoing.sample-queue.connector=smallrye-rabbitmq
mp.messaging.outgoing.sample-queue.default-routing-key=sample-queue
mp.messaging.outgoing.sample-queue.exchange.name=amq.default
mp.messaging.outgoing.sample-queue.exchange.declare=false
```

Code:
``` java
import org.eclipse.microprofile.reactive.messaging.Channel;
import org.eclipse.microprofile.reactive.messaging.Emitter;
...
    @Channel("sample-queue")
    Emitter<OutputMessage> sampleQueue;
...
    sampleQueue.send(message);
```

But the message **was not reached to destine** and the output log says:

Log DEBUG:
```log
DEBUG [io.sma.rea.mes.rabbitmq] (vert.x-eventloop-thread-0) SRMSG17022: Sending a message to exchange `amq.default` with routing key sample-queue
```

If I publish manually using the same arguments the message was reached:

image

Is it possible to send a direct message from this library? I don't have any problem using exchange, just using the direct publication with the amq.default exchange.

I'm using Quarkus version `2.7.3.Final`  with dependency:
```xml
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-smallrye-reactive-messaging-rabbitmq</artifactId>
    </dependency>
```

hjklapp:
This PR fixes #1641.

To address specific queues on RabbitMQ, e.g. when using the [RPC pattern](https://www.rabbitmq.com/tutorials/tutorial-six-java.html), the message has to be sent to the *default exchange*.

Currently, it is not possible to send message to the *default exchange*, as its implicit name is `""`, and Microprofile Config does not allow setting configuration parameters to `""`. Furthermore, the *default exchange* should not be declared again, so a separate treatment is required.

The proposed solution adds a new configuration parameter for outgoing RabbitMQ channels `exchange.is-default`. When the parameter is set to `true` the exchange is
1. not declared (so all other parameters concerning the exchange are ignored) and
2. `""` is used as exchange name when sending messages to RabbitMQ.

The change does not break existing configurations.

### DIFF
--- a/documentation/src/main/docs/rabbitmq/sending-messages-to-rabbitmq.md
+++ b/documentation/src/main/docs/rabbitmq/sending-messages-to-rabbitmq.md
@@ -138,3 +138,12 @@ mp.messaging.outgoing.people-out.exchange.declare=true
     exchange or have a different default routing key, then the
     `exchange.type` and `default-routing-key` properties need to be
     explicitly specified.
+
+## Sending to specific queues via the default exchange
+
+To send a message to a specific queue (usually a reply queue), you have to configure the default exchange as an outgoing channel and set the name of the queue as routing key in the message metadata.
+
+```properties
+mp.messaging.outgoing.channel-name-for-default-exchange.connector=smallrye-rabbitmq
+mp.messaging.outgoing.channel-name-for-default-exchange.exchange.is-default=true
+```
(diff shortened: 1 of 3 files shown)


ozangunalp:
@hjklapp Thanks for the PR and the explanation. 
I remember that the default exchange has an empty string name, but didn't test if it actually worked or not. 

I think it'd be more intuitive to handle the case of an empty string in `exchange.name` config. MP Config passes the value directly without escaping quotes, so we can simply escape `"` and `'` for exchange names and treat them as empty string.

I see that you handled the establishExchange case with the `is-default` flag, what I'd suggest is once we escape quotes from exchange names and if we have an empty string:
- skip establishExchange
- skip queueBind to the default exchange
- send messages to ''

ozangunalp:
@hjklapp What I had in mind is something like this: https://github.com/smallrye/smallrye-reactive-messaging/pull/1949

Could you take a look?

PR #1982, opened 2022-11-12
RabbitMQ - Attribute to connect on queues with priority level "x-max-priority"

torgge:
Attribute to set "Pri".
On rabbitmq x-max-priority param, to work with priorizable queues
To solve: "reply-text=PRECONDITION_FAILED - inequivalent arg 'x-max-priority' for queue " error.


smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
```diff
@@ -119,6 +119,9 @@
 @ConnectorAttribute(name = "dead-letter-queue-type", direction = INCOMING, description = "If automatically declare DLQ, we can choose different types of DLQ [quorum, classic, stream]", type = "string")
 @ConnectorAttribute(name = "dead-letter-queue-mode", direction = INCOMING, description = "If automatically declare DLQ, we can choose different modes of DLQ [lazy, default]", type = "string")
 
+//Priority
+@ConnectorAttribute(name = "max-priority", direction = INCOMING, description = "Define priority level queue consumer", type = "int")
```
ozangunalp, line 123: Looking at other queue x-attributes can we rename this one to `queue.x-max-priority`?

ozangunalp (changes requested):

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
torgge: Hello @ozangunalp yes we can.

ozangunalp (approved):

PR #1994, opened 2022-11-22
Remove ContextOperator completely from connectors

ozangunalp:


ozangunalp:
Always fails on the CI and never locally. I'll investigate today.

PR #2007, opened 2022-11-30
Restore context decorator

ozangunalp:


pom.xml
```diff
@@ -110,7 +110,7 @@
     <jandex-maven-plugin.version>1.2.3</jandex-maven-plugin.version>
 
     <nats-embedded.version>2.1.0</nats-embedded.version>
-    <slf4j-log4j12.version>2.0.5</slf4j-log4j12.version>
+    <slf4j-log4j12.version>1.7.36</slf4j-log4j12.version>
```
cescoffier, line 113: we should ban slf4j 2 for now.

cescoffier (dismissed):

pom.xml
ozangunalp: I've fixed the dependabot config (I think)

cescoffier (approved):

PR #2024, opened 2022-12-11
HealtCheck on MQTT Connector

dometec:
Please double-check the class MqttSink, since I've changed the MQTT client startup.
 
As per javadoc on OutboundConnection ("Note that the connection to the <em>transport</em> or <em>broker</em> is generally postponed until the subscription." ) I've implemented a Processor and, on subscribe, I start the MQTT client.

Before the connection was made on the first message out, but I need to get the connection's health also when no messages are sent,


dometec:
Hi @cescoffier, do you have feedback on this?

PR #2034, opened 2023-01-06
Set metrics name for Vertx based clients on mqtt rabbitmq and amqp

ozangunalp:
So that micrometer metrics will collect metrics for the client.


cescoffier (approved):

ozangunalp:
@cescoffier I've been struggling to test Vert.x net client metrics using micrometer. Am I not supposed to find connection count etc. in the global registry?

cescoffier:
Not sure about connection count (it's a metric per time window).

Do you have any metrics with the expected prefix?

cescoffier:
You cannot do that - you need to be in Quarkus.

The vert.x micrometer metrics are not reporting the same metrics as (new) quarkus-micrometer. I aligned the reported metrics with the other quarkus-micrometer metrics.

The metrics name using `|` are read and interpreted by quarkus-micrometer only.

ozangunalp:
Yes, I later looked the micrometer extension. LGTM with a ad-hoc test.

PR #2052, opened 2023-01-19
HealtCheck on MQTT Connector

dometec:
Please double-check the class MqttSink, since I've changed the MQTT client startup.

As per javadoc on OutboundConnection ("Note that the connection to the transport or broker is generally postponed until the subscription." ) I've implemented a Processor and, on subscribe, I start the MQTT client.

Before this change, the connection was made on the first message out, but I need to get the connection's health also when no messages are sent.

(Is the same PR as 2024, but by mistake I had to create a new one)

### DIFF
--- a/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
+++ b/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
@@ -3,8 +3,6 @@ package io.smallrye.reactive.messaging.mqtt;
 import static io.smallrye.reactive.messaging.mqtt.i18n.MqttLogging.log;
 
 import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -13,18 +11,17 @@ import javax.enterprise.inject.Instance;
 import org.eclipse.microprofile.reactive.messaging.Message;
 import org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams;
 import org.eclipse.microprofile.reactive.streams.operators.SubscriberBuilder;
-import org.reactivestreams.Processor;
 import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
 
 import io.netty.handler.codec.mqtt.MqttQoS;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.vertx.AsyncResultUni;
 import io.smallrye.reactive.messaging.OutgoingMessageMetadata;
 import io.smallrye.reactive.messaging.health.HealthReport.HealthReportBuilder;
-import io.smallrye.reactive.messaging.mqtt.Clients.ClientHolder;
 import io.smallrye.reactive.messaging.mqtt.internal.MqttHelpers;
+import io.smallrye.reactive.messaging.mqtt.session.MqttClientSession;
 import io.smallrye.reactive.messaging.mqtt.session.MqttClientSessionOptions;
+import io.smallrye.reactive.messaging.providers.helpers.MultiUtils;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
@@ -41,7 +38,6 @@ public class MqttSink {
     private final SubscriberBuilder<? extends Message<?>, Void> sink;
 
     private final AtomicBoolean started = new AtomicBoolean();
-    private final AtomicBoolean ready = new AtomicBoolean();
     private final AtomicReference<Clients.ClientHolder> reference = new AtomicReference<>();
 
     public MqttSink(Vertx vertx, MqttConnectorOutgoingConfiguration config,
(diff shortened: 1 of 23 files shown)


ozangunalp (changes requested):

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
```diff
@@ -29,46 +33,118 @@
...
+
         MqttClientSessionOptions options = MqttHelpers.createClientOptions(config, instances);
-        topic = config.getTopic().orElseGet(config::getChannel);
+
+        channel = config.getChannel();
+        topic = config.getTopic().orElse(channel);
         qos = config.getQos();
+        healthEnabled = config.getHealthEnabled();
 
-        AtomicReference<Clients.ClientHolder> reference = new AtomicReference<>();
```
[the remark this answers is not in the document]
dometec: Hi @ozangunalp , is your comment referred to only this line or to all changes in this class?

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
```diff
@@ -29,46 +33,118 @@
...
                 .ignore();
+
     }
 
-    private CompletionStage<?> send(AtomicReference<Clients.ClientHolder> reference, Message<?> msg) {
-        MqttClientSession client = reference.get().getClient();
+    /*
+     * This processor let che client mqtt to connect on su
+     */
+    private class ConnectOnSubscribeProcessor implements Processor<Message<?>, Message<?>> {
```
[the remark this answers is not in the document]
dometec: I use the lock because could be a race condition between the two sides of processor. Could be possible to have a requestedItem > 0 (line 95) and, on the other side, skip the subscription.request() (line 118), and stay idle forever. If there is only one thread this could not happen, but since I do not sure of that, I've used a lock object.

The difference between ready and started is that started can pass only from false to true when the first connection is done (and stay true), ready can be false if the mqtt connection is down.

I'll remove the sysout.

dometec:
Hi @ozangunalp , do you have any feedback?

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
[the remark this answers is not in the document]
ozangunalp: It is tricky to write a processor from scratch. Instead of the lock, the usual way is to check on both `subscribe` and `onSubscribe` and call the `downstream.onSubscribe`  only when you have both downstream and upstream. But what happens if the `client.start` fails ? The implementation also needs to handle cancellation and downstream error propagation. 

I think it is correct to have the startup flag on the subscription time instead of the first outgoing message dispatch.
In the Mutiny Flow API branch, there is an easier way to write that operator using Mutiny `onSubscription`. I'd like to bring that to this branch for easier implementation. Plus it'll eliminate my concerns about conflict. I've pushed the change here if you'd like to integrate to this PR: https://github.com/smallrye/smallrye-reactive-messaging/commit/7636517ecd649c2604ade9998f3bebe556e06200

Otherwise, I still think having both started and ready flags is unnecessary. I would only keep the started flag for the startup probe and use the `client.isConnected()` for both readiness and liveness probes.

Btw that's only for the outgoing channels, for incoming ones what you did should work very well.

ozangunalp:
@dometec sorry I need some time to check stuff. I've responded to the comment directly.

dometec:
Hi @ozangunalp, I integrated your change to the PR and I did some tests, all ok! Thanks. Yes, writing a processor from scratch wasn't easy :) 

One last thing, reading here https://developers.redhat.com/blog/2020/11/10/you-probably-need-liveness-and-readiness-probes I changed readiness e liveness probe (see last commit). In detail, I understood that fail liveness leads to a restart of the service, so I set it to false when the reactive stream end. Moreover, in MqttSouce I removed the `started.set(false)` since the started goes only from false to true.

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
```diff
@@ -38,6 +38,7 @@ public class MqttSink {
     private final SubscriberBuilder<? extends Message<?>, Void> sink;
 
     private final AtomicBoolean started = new AtomicBoolean();
+    private final AtomicBoolean alive = new AtomicBoolean();
```
ozangunalp, line 41: Am I missing something or alive flag is not used later?

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
```diff
@@ -151,7 +162,7 @@ public void isReady(HealthReportBuilder builder) {
 
     public void isAlive(HealthReportBuilder builder) {
         if (healthEnabled)
-            builder.add(channel, isConnected());
+            builder.add(channel, !isDisconnected());
```
ozangunalp, line 165: Isn't having `isConnected` and `isDisconnected` methods confusing? 
With this line if `reference` is empty the liveness would be OK, is this intended?

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSource.java
```diff
@@ -108,12 +111,12 @@ public void isStarted(HealthReportBuilder builder) {
 
     public void isReady(HealthReportBuilder builder) {
         if (healthEnabled)
-            builder.add(channel, ready.get());
+            builder.add(channel, alive.get());
```
ozangunalp, line 114: Can't we keep this flag `ready` as it is used for readiness probe?

smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/MqttSink.java
dometec: You right, let me fix it... it was very late yesterday :)

ozangunalp (approved):
LGTM, let's go with that. 
As this is the first health check support for mqtt, future adjustments can be done later.

PR #2059, opened 2023-01-24
Kafka ProducerInterceptor Bean

ozangunalp:
Kafka connector outgoing channel attribute `interceptor-bean` for configuring a managed interceptor bean.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
@@ -24,10 +24,12 @@ import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerInterceptor;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.serialization.Serializer;
 
 import io.smallrye.common.annotation.CheckReturnValue;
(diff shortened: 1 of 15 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
```diff
@@ -147,18 +159,32 @@ public Uni<Void> runOnSendingThread(java.util.function.Consumer<Producer<K, V>>
...
+                        intercepted = interceptor.onSend(record);
                     } else {
-                        em.complete(metadata);
+                        intercepted = record;
                     }
-                })));
+                    return Uni.createFrom().emitter(em -> c.send(intercepted, (metadata, exception) -> {
+                        if (exception != null) {
+                            if (interceptor != null) {
+                                interceptor.onAcknowledgement(getRecordMetadataForFailure(intercepted), exception);
```
cescoffier, line 172: Wondering if we need a onNegativeAcknowledgement method instead (to be clear)

cescoffier, line 172: What if the method throws an exception? Probably better to handle that case.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
```diff
@@ -147,18 +159,32 @@ public Uni<Void> runOnSendingThread(java.util.function.Consumer<Producer<K, V>>
...
+                            }
+                            if (record.topic() != null) {
+                                log.unableToWrite(this.channel, record.topic(), exception);
+                            } else {
+                                log.unableToWrite(this.channel, exception);
+                            }
+                            em.fail(exception);
+                        } else {
+                            if (interceptor != null) {
+                                interceptor.onAcknowledgement(metadata, null);
```
cescoffier, line 182: What if the interceptor throws an exception?

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/ReactiveKafkaProducer.java
ozangunalp: The idea is to be able to use the standard ProducerInterceptor interface from the Kafka package.

ozangunalp: The interceptor is called inside the uni emitter. So a failure will be emitted, it'll be considered like a send error retried etc. finally nacking the message

ozangunalp:
@cescoffier I decided to log and ignore the exceptions as documented in `ProducerInterceptor` `onSend` and `onAcknowledge` methods.

ozangunalp:
Since the update revapi filters behave oddly.

cescoffier (approved):

PR #2060, opened 2023-01-24
Configure missing tracing attributes for kafka, amqp rabbitmq.

ozangunalp:
Restore system attributes from connector name to messaging system name

@radcortez I noticed that messaging.system attribute where changed, please review these fixes.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/tracing/AmqpAttributesExtractor.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/tracing/AmqpAttributesExtractor.java
@@ -4,7 +4,6 @@ import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
 import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesGetter;
-import io.smallrye.reactive.messaging.amqp.AmqpConnector;
 import io.smallrye.reactive.messaging.amqp.AmqpMessage;
 
 public class AmqpAttributesExtractor implements AttributesExtractor<AmqpMessage<?>, Void> {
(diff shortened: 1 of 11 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/tracing/AmqpAttributesExtractor.java
```diff
@@ -39,7 +38,7 @@ private static class AmqpMessagingAttributesGetter implements MessagingAttribute
         // Required
         @Override
         public String system(final AmqpMessage<?> amqpMessage) {
-            return AmqpConnector.CONNECTOR_NAME;
+            return "AMQP 1.0";
```
radcortez, line 41: I am wondering if this should be `amqp` only?

radcortez (approved):

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/tracing/AmqpAttributesExtractor.java
ozangunalp: I hesitated on that as well, then kept the one we had before.

radcortez: Ok, just leave that as is :)

cescoffier: AMQP is misleading. For most people it's the AMQP from RabbitMQ (8. something), and not AMQP 1.0. So, better be clear about that.

radcortez: We can set the system name and the protocol. For RabbitMQ we had `amqp` set as the protocol, but we cleared that out.

For AMQP, I guess that both the system and protocol are `amqp`.

PR #2075, opened 2023-02-06
Split package resolution for io.smallrye.reactive.messaging.providers.connectors

ozangunalp:
in-memory package moved to io.smallrye.reactive.messaging.providers.connectors.in.memory


documentation/src/main/java/testing/MyTest.java
```diff
@@ -8,9 +8,9 @@
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-import io.smallrye.reactive.messaging.providers.connectors.InMemoryConnector;
-import io.smallrye.reactive.messaging.providers.connectors.InMemorySink;
-import io.smallrye.reactive.messaging.providers.connectors.InMemorySource;
+import io.smallrye.reactive.messaging.in.memory.InMemoryConnector;
```
cescoffier, line 11: I'm not sure I like the `in.memory`, maybe just `memory`?

cescoffier (approved):

documentation/src/main/java/testing/MyTest.java
ozangunalp: 👍

PR #2079, opened 2023-02-08
Remove Message payload null check

ozangunalp:


api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
```diff
@@ -56,15 +56,12 @@ static <T> Message<T> of(T payload) {
      * Create a message with the given payload and metadata.
      * The acknowledgement and negative-acknowledgement are immediate.
      *
-     * @param payload The payload, must not be {@code null}.
+     * @param payload The payload.
      * @param metadata The metadata, if {@code null} an empty set of metadata is used.
```
cescoffier, line 60: This code comes from the spec - we cannot change it.

cescoffier (changes requested):

api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java
ozangunalp: I know but I don't see any non-null requirement in the API code https://github.com/eclipse/microprofile-reactive-messaging/blob/master/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Message.java

ozangunalp: Some would call this syndrome split-brain :D

documentation/src/main/docs/concepts/model.md
```diff
@@ -72,6 +72,12 @@ Creating `Messages` is done using the {{ javadoc('io.eclipse.microprofile.reacti
 {{ insert('messages/MessageExamples.java', 'creation') }}
 ```
 
+Messages accept `null` as payload.
```
cescoffier, line 75: I would warn the user that reactive streams do not support `null`, so you cannot consume a `Multi<P>` where the message (Message<P>) has a `null` payload.

cescoffier (commented):
Don't we have a section in the kafka doc about tombstone?

PR #2080, opened 2023-02-10
Re-enable camel connector with jakarta namespace

ozangunalp:
Camel module only added on java 17 build.


pom.xml
```diff
@@ -678,5 +677,14 @@
         <groups>slow,flaky</groups>
       </properties>
     </profile>
+    <profile>
+      <id>camel</id>
+      <activation>
+        <jdk>17</jdk>
```
cescoffier, line 683: I'm wondering if we need to configure the java target / source. 

If it stays like that, we need to see how we can release stuff without making it too complicated. Maybe all but camel and docs should use 11 as source and target, and 17 for the rest. We keep the CI build as it is currently (with your profile, but we may need to do the same for docs).

Then we cut releases using java 17 (I think we do already because of revapi).

smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/incoming/IncomingCamelTest.java
```diff
@@ -46,8 +46,12 @@ public void testWithABeanDeclaringACamelSubscriber() {
         initialize();
 
         await().until(() -> {
-            List<String> list = Files.readAllLines(path);
-            return list.size() == 1 && list.get(0).equalsIgnoreCase("abcd");
+            try {
```
cescoffier, line 49: I'm not sure it's required. That was more a 'what the heck???' thing.

cescoffier (changes requested):

pom.xml
ozangunalp: That makes sense, let's start releasing on java 17.

smallrye-reactive-messaging-camel/src/test/java/io/smallrye/reactive/messaging/camel/incoming/IncomingCamelTest.java
ozangunalp: I've removed it.

ozangunalp:
I've also deleted the CI branch running Kafka tests on broker 2.8.0.

cescoffier (approved):

cescoffier:
The documentation is missing. We should restore it, too (I let you decide if you want it in the same PR or another PR).

ozangunalp:
Totally forgot that we removed the camel doc because it wasn't compiling. I'll restore it in this PR

ozangunalp:
Added a warning to the Camel doc for Camel 4 and java 11 support drop.

documentation/src/main/macros/docissimo.py
```diff
@@ -12,13 +12,14 @@
 ```
 attributes:
   versions:
+    camel: 3.13.0
```
cescoffier, line 15: Shouldn't it be 4?

documentation/src/main/macros/docissimo.py
```diff
@@ -12,13 +12,14 @@
...
     spec: 2.0.1
     mutiny: 1.1.2
   project-version: '3.14.0-SNAPSHOT'
   smallrye-config-version: '2.7.0'
 ```
 
-You can access values using `{{ attributes.versions.spec}}` or `{{ attributes['project-version'] }}`.
+You can access values using `{{ attributes.versions.camel}}` or `{{ attributes['project-version'] }}`.
```
cescoffier, line 22: Why changing this one?

cescoffier (changes requested):

documentation/src/main/macros/docissimo.py
ozangunalp: It is part of the usage docs of this file. I reverse applied the patch we prepared to remove the camel connector docs.

cescoffier: Ah funny, didn't remember that!

documentation/src/main/macros/docissimo.py
ozangunalp: Same for the previous comment

cescoffier (approved):

Issue #2084, opened 2023-02-15
RabbitMQ - Prefetch configuration not respected after connection restart


maragud:
The configuration for `max-outstanding-messages` seems to be applied only during the initial connection of the application. After the connection is being restarted the RabbitMQ consumer's prefetch size is always 0.

The issue is caused because the `client.basicQos(ic.getMaxOutstandingMessages().get(), false);` is only invoked when the consumer is first created. I think that moving it to the connection established callback would resolve the issue, since I checked that the `establishQueue` method is invoked after restarting the connection.

To replicate the issue:
* Set the property `max-outstanding-messages` to a value larger than 0.
* Start the application and check the prefetch count through the RabbitMQ UI
* Close this connection through the RabbitMQ UI and check the prefetch count

PR #2086, opened 2023-02-16
RabbitMQ - Prefetch configuration not respected after connection restart #2084

maragud:
Changes to send a `basicQos` request after reconnecting to the RabbitMQ broker, to change the prefetch count
https://github.com/smallrye/smallrye-reactive-messaging/issues/2084


ozangunalp (approved):
lgtm thank you @maragud

Issue #2085, opened 2023-02-15
RabbitMQ - Failure handler discoverable by CDI


scrocquesel:
We would like to be able to handle failure differently depending on the `Throwable` and I would also want to add a failure handler that would send the message to a different queue to be reconsume latter.

Would it be possible to open the closed enum to a CDI based set of factory like the kafka connector ?

PR #2087, opened 2023-02-16
Refactor RabbitMQ failure handler

scrocquesel:
#2085

Factory interface is designed just to make the actual handler works without any change.

### DIFF
--- a/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
+++ b/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
@@ -173,6 +173,14 @@ controlled by the `failure-strategy` channel setting:
 -   `reject` - this strategy marks the RabbitMQ message as rejected
     (default). The processing continues with the next message.
 
+!!!warning "Experimental"
+    `RabbitMQFailureHandler` is experimental and APIs are subject to change in the future
+
+In addition, you can also provide your own failure strategy. To provide a failure strategy implement a bean exposing the interface
+{{ javadoc('io.smallrye.reactive.messaging.rabbitmq.fault.RabbitMQFailureHandler') }}, qualified with a `@Identifier`. Set the name of the bean
+as the `failure-strategy` channel setting.
+
+
 ## Configuration Reference
 
 {{ insert('../../../target/connectors/smallrye-rabbitmq-incoming.md') }}
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
@@ -46,14 +46,12 @@ import io.smallrye.reactive.messaging.connector.OutboundConnector;
 import io.smallrye.reactive.messaging.health.HealthReport;
 import io.smallrye.reactive.messaging.health.HealthReporter;
 import io.smallrye.reactive.messaging.providers.connectors.ExecutionHolder;
+import io.smallrye.reactive.messaging.providers.helpers.CDIUtils;
 import io.smallrye.reactive.messaging.providers.helpers.MultiUtils;
 import io.smallrye.reactive.messaging.rabbitmq.ack.RabbitMQAck;
 import io.smallrye.reactive.messaging.rabbitmq.ack.RabbitMQAckHandler;
 import io.smallrye.reactive.messaging.rabbitmq.ack.RabbitMQAutoAck;
-import io.smallrye.reactive.messaging.rabbitmq.fault.RabbitMQAccept;
-import io.smallrye.reactive.messaging.rabbitmq.fault.RabbitMQFailStop;
 import io.smallrye.reactive.messaging.rabbitmq.fault.RabbitMQFailureHandler;
-import io.smallrye.reactive.messaging.rabbitmq.fault.RabbitMQReject;
 import io.smallrye.reactive.messaging.rabbitmq.tracing.RabbitMQOpenTelemetryInstrumenter;
 import io.smallrye.reactive.messaging.rabbitmq.tracing.RabbitMQTrace;
 import io.vertx.core.json.JsonObject;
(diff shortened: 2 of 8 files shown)


ozangunalp (changes requested):
LGTM! Could you add a short section in http://smallrye.io/smallrye-reactive-messaging/4.3.0/rabbitmq/receiving-messages-from-rabbitmq/#failure-management ?

scrocquesel:
I'm trying to build a reconsume later strategy to test the apis exposed by the connector. I'm really new to reactive stream and I struggle to emit delayed message to a subscriber returned by the connector getSubscriber method.

In the following code, I don't know why the multi subscribe method expect a `super Message` and the connector provide a `extends Message`. Should the connector expose another method to return a more appropriate subcriber ? Am I misusing the Multi api ?

````java
public class RabbitMQReconsumeLater implements RabbitMQFailureHandler {
    private MultiEmitter<? super Message<?>> emitter;

    public RabbitMQReconsumeLater(RabbitMQConnector connector, RabbitMQConnectorIncomingConfiguration config) {

        // create outgoing config for a delayed channel from the incoming channel configuration
        final Config config = ...;

        // create a subscriber to push message to
        final Subscriber<? extends Message<?>> processor = connector.getSubscriber(null);

        Multi<Message<?>> multi = Multi.createFrom().<Message<?>> emitter(e -> emitter = e);      
        multi.subscribe(processor); // The method subscribe(Flow.Subscriber<? super Message<?>>) in the type Flow.Publisher<Message<?>> is not applicable for the arguments (Flow.Subscriber<capture#4-of ? extends Message<?>>)
    }

    @Override
    public <V> CompletionStage<Void> handle(IncomingRabbitMQMessage<V> msg, Context context, Throwable reason) {
        // new outgoing message with ttl from msg
        Message<V> message = ...;

        emitter.emit(message);

    }

scrocquesel:
@ozangunalp 

> I wouldn't have thought of using the connector to create another outgoing channel stream. The channel will be present on the health check, other than that I don't think there'll be a negative impact. It can be challenging to construct a `Config` object to pass to the `getSubscriber` method, you'd need to override some properties of the incoming channel config like `channel-name`.

That's right. Basicly, I would create an outgoing channel based on the incoming configuration. That's what the Kafka dead letter handler do but there is less coupling between the connector/connector config and the reactive client wrapper. For healthcheck, I guess the client should keep the channel opened and thus it could be monitored in some way. Maybe aggregating the status of both rabbitmq channels to the same connector channel status.

> For the Emitter, you can create a `MutinyEmitterImpl` and call `getPublisher` and subscribe to it using the subscriber you created earlier. To create a `MutinyEmitterImpl` you'd need an `EmitterConfiguration` implementation mostly for specifying the overflow strategy.
> 

I have seen that pattern in some test but and try it. As far as I remember, I fall on the same mismatch type (super/extends). I end up splitting getSubscriber to get a `RabbitMQMessageSender` directly. Still have to address reporting of the channel status.

> That being said I think in RabbitMQ one can configure the `dead-letter-exchange` and `dead-letter-routing-key` to implement redelivery.

Yes and you can configure a queue TTL on the DLQ to have the message back to original queue, but then:
- you can't have both a DLQ and a reconsume later queue.
- you can't have a per message TTL

The idea is to duplicate the message and send it with new metadata and ack the old one when the new one is received.

ozangunalp:
@scrocquesel If you'd like to work on it and contribute to the project I've continued on your idea to make it work: https://github.com/ozangunalp/smallrye-reactive-messaging/commit/ee54305d5a6b9a68ab22a86b735530c3e42560b3

Please feel free to continue on it to implement your use case.

scrocquesel:
> @scrocquesel If you'd like to work on it and contribute to the project I've continued on your idea to make it work: https://github.com/ozangunalp/smallrye-reactive-messaging/commit/ee54305d5a6b9a68ab22a86b735530c3e42560b3
> 
> Please feel free to continue on it to implement your use case.

Thanks, I'll definitely have a look.

scrocquesel:
@ozangunalp I have added a working reconsume later strategy. I tried not to expose publicly much of RabbitMQConnector method but I was in need of a way to create the reconsume queue.

Also, I'm not sure of what the overflow buffer size should be. In my understanding, the upstream is waiting for the original message to be ack/nack. So the reconsumer emitter should not have to deal with more than the upstream buffer size which is `max-incoming-internal-queue-size`.

scrocquesel:
> @ozangunalp I have added a working reconsume later strategy. I tried not to expose publicly much of RabbitMQConnector method but I was in need of a way to create the reconsume queue.
> 
> Also, I'm not sure of what the overflow buffer size should be. In my understanding, the upstream is waiting for the original message to be ack/nack. So the reconsumer emitter should not have to deal with more than the upstream buffer size which is `max-incoming-internal-queue-size`.

@ozangunalp Any progress on this one ? I rebased to latest main. It miss the doc for reconsume strategy but want to be sure you it could be part of the lib.

ozangunalp:
@scrocquesel Thank you very much for this, I am sorry I couldn't respond earlier.
I still try to understand how this is different than only setting up a [dead-letter-exchange](https://www.rabbitmq.com/dlx.html). I see that you've added also this. But I fail to grasp how both work together. Would you mind adding a couple of dedicated test cases for this code? 

Maybe the reconsume-later feature can be implemented only by setting up a `dead-letter-exchange` and a check in the failure strategy for the retry count? I'll try to set up some test cases on my side to see if we can do only that.

And if setting up a retry ttl per message is not that crucial, it wouldn't justify the complexity we are adding with this PR.

ozangunalp:
@scrocquesel Thank you very much for this, I am sorry I couldn't respond earlier.
I still try to understand how this is different than only setting up a [dead-letter-exchange](https://www.rabbitmq.com/dlx.html). I see that you've added also this. But I fail to grasp how both work together. Would you mind adding a couple of dedicated test cases for this code? 

Maybe the reconsume-later feature can be implemented only by setting up a `dead-letter-exchange` and a check in the failure strategy for the retry count? I'll try to set up some test cases on my side to see if we can do only that.

And if setting up a retry ttl per message is not that crucial, it wouldn't justify the complexity we are adding with this PR.

scrocquesel:
> @scrocquesel Thank you very much for this, I am sorry I couldn't respond earlier. I still try to understand how this is different than only setting up a [dead-letter-exchange](https://www.rabbitmq.com/dlx.html). I see that you've added also this. But I fail to grasp how both work together. Would you mind adding a couple of dedicated test cases for this code?
> 
> Maybe the reconsume-later feature can be implemented only by setting up a `dead-letter-exchange` and a check in the failure strategy for the retry count? I'll try to set up some test cases on my side to see if we can do only that.
> 
> And if setting up a retry ttl per message is not that crucial, it wouldn't justify the complexity we are adding with this PR.

A common pattern involves configuring a waiting queue with a time-to-live (TTL) and a dead-letter queue (DLQ) that targets the original queue. If a message is published to a queue with a TTL, it will expire after that duration. When a message expires in a queue, RabbitMQ will route it to its corresponding DLQ, if one exists (thereby returning it to the original queue), or delete it.

The failure strategy involves publishing a copy of the message to the waiting queue and acknowledging the original message to discard it. Later, the copy message will expire and will be reprocessed from the main queue.

Only one DLQ can be assigned to a queue. Thus, if a message is "nacked" to the DLQ as a retry mechanism, a real DLQ cannot exist. This lead to two cons:
- when retry count is exhausted, we can only ack the message and discard it.
- it should now be possible for users to implement their own failure strategy based on the exception and choose to either reconsume the message later or directly nack it to the DLQ/discard it using the existing set of failure strategies.

I agree for the complexity. While the strategy itself is simple, the code required to create a queue is not very reusable. Refactoring may be necessary to better share this code without resorting to the reconsume strategy trick with the config map.

scrocquesel:
@ozangunalp I reverted the branch to only refactor failure handlers with CDI (and add a missing settings for delivery limit). All in all with #2182 , it'll allow me to achieve what I want:  Retry latter on specific exception + keep poison message in a queue without consumer.

ozangunalp:
@scrocquesel In my earlier comment https://github.com/smallrye/smallrye-reactive-messaging/pull/2087#issuecomment-1507429515 I mentioned this possibility and tested it out, and hopefully before I had time to create a PR @andr3adc actually did it :) 

Thank you very much for your contribution!

ozangunalp (approved):

PR #2106, opened 2023-03-06
Decouple OpenTelemetry instrumenter

kabir:
If tracing is not enabled, it should be possible to use the Kafka and AMQP connectors without needing to include the OpenTelemetry instrumentor classes on the classpath.

### DIFF
--- a/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
+++ b/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
@@ -20,7 +20,6 @@ import java.util.stream.Collectors;
 
 import javax.net.ssl.SSLContext;
 
-import jakarta.annotation.PostConstruct;
 import jakarta.annotation.Priority;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.BeforeDestroyed;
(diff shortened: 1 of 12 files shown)


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -247,6 +220,10 @@ public Flow.Publisher<? extends Message<?>> getPublisher(Config config) {
 
         AmqpFailureHandler onNack = createFailureHandler(ic);
 
+        if (tracing && amqpInstrumenter == null) {
+            amqpInstrumenter = AmqpOpenTelemetryInstrumenter.createForConnector();
```
kabir, line 224: Moving this here is my biggest doubt. How would I go about getting the value of `tracing` from the @PostConstruct method?

kabir:
It turns out the main OpenTelemetry integration in WildFly needs these jars, so this is no longer necessary for the reactive messaging integration

kabir:
Reopened since the OTel instrument jars in WildFly are not compatible with the ones in 4.4.0.

kabir:
@ozangunalp I can confirm with this fix that I am able to update WildFly to a SNAPSHOT built from this branch. Since the instrumenter stuff doesn't get loaded, I get a clean run again.

As mentioned in chat, my pull request to upgrade to 4.4.0 had to downgrade to 4.3.0 due to API changes in the OTel classes. Essentially WildFly needs to stay on 1.20 of OTel for now, while SR RM is on 1.23/

ozangunalp (approved):
LGTM!

PR #2112, opened 2023-03-10
Fix AMQP tracing

ozangunalp:


smallrye-reactive-messaging-amqp/pom.xml
```diff
@@ -56,6 +56,12 @@
       <version>${artemis.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.14.2</version>
```
cescoffier, line 62: Should we add this to the dependency management? We have other modules using jackson no?

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpBroker.java
```diff
@@ -33,6 +33,7 @@ public Map<String, String> start(int port) {
                 config.setLargeMessagesDirectory("target/data/large-messages");
                 config.addAcceptorConfiguration("in-vm", "vm://0");
                 config.addAcceptorConfiguration("tcp", "tcp://127.0.0.1:" + port);
+                config.setMaxDiskUsage(99);
```
cescoffier, line 36: That's a sneaky way to postpone the cleanup of your file system :-D

cescoffier (approved):

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/AmqpBroker.java
ozangunalp: :) I blame the openshift local

Issue #2115, opened 2023-03-12
Usage without Quarkus?


craig-day:
I'm hoping to use this implementation of MP Reactive Messaging outside of Quarkus. I'm primarily trying to spin up some Kafka consumers in a Vert.x app. Initially it seemed straightforward by just adding the `ReactiveMessagingExtension` to my Weld container, however that opened a rabbit hole of other managed beans to add. I've been working through all the errors and adding the beans as Weld complains about being unable to find them. I ended up with something like this: 

```java
    weld.addExtension(new ReactiveMessagingExtension());

    // TODO: why do we need to set all this up for SmallRye?
    weld
        .addBeanClass(MediatorManager.class)
        .addBeanClass(HealthCenter.class)
        .addBeanClass(InternalChannelRegistry.class)
        .addBeanClass(WorkerPoolRegistry.class)
        .addBeanClass(MediatorFactory.class)
        .addBeanClass(Wiring.class)
        .addBeanClass(ExecutionHolder.class);

    weld.addPackage(true, KafkaConnector.class);
```

I've gotten through all the CDI errors, and now I'm getting stuck with the mediator complaining about channels without upstreams configured. 

Wondering if I'm missing something critical or if there is an implicit dependency on Quarkus.

cescoffier:
You can use smallrye reactive messaging without Quarkus. Most of our tests are just using Weld (SE). For the Kafka connector, you would need something like this:

https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/base/WeldTestBase.java#L67


However, if your application is using Vert.x and CDI, I recommend you to use Quarkus, as it will package both nicely and optimize a lot of things for you.

craig-day:
Thanks for this! Would you be open to adding a CDI class that does this directly in here? This got me a lot further, and I also found that I need to produced a mutiny vertx instance delegating to my regular vertx instance. 

Yes, I know Quarkus does this for me and we're working on getting our apps there, but there are lots of other things that aren't quarkus compatible for me right now so it's not a simple switch. Having a CDI support class that I could add that wires all the rest up would be super helpful.

cescoffier:
What do you mean with a CDI support class? (Sorry, not a cdi expert).

If we can do anything to make it easier, it's more than welcome!

\CC @mkouba the living CDI reference.

craig-day:
[...]

PR #2125, opened 2023-03-29
introduce two CDI helper extension to add beans explicitly

craig-day:
As discussed in #2115, this introduced some CDI extensions to act as a bridge for environments where auto bean discovery is not utilized. A consumer can chose to explicitly add these extensions to their CDI container to get the same support that auto discovery would have provided.

Opening as a draft to start to see how we feel about the approach before propagating it to all other connector implementations and updating tests.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIExtension.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaCDIExtension.java
@@ -1,0 +1,41 @@
+package io.smallrye.reactive.messaging.kafka;
+
+import io.smallrye.reactive.messaging.kafka.commit.FileCheckpointStateStore;
+import io.smallrye.reactive.messaging.kafka.commit.KafkaCheckpointCommit;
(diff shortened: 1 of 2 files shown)


manovotn:
Few comments coming to my mind.

1. From CDI perspective, this form of bean registration seems fine :+1: 
However, I'd add a comment into each extension denoting that these CDI extensions are *disabled* by default as they intentionally have no META-INF entry in this repository and that they are meant to be used explicitly by integrators.

2. As for the set of classes added, I have no idea which classes SR reactive messaging need so I'll trust you know that.
How likely is this set going to change or expand? Since this way it would be hard to remember to add these beans into extensions as well.

3. I think the tests using Weld SE should leverage these extensions - i.e. the base test class should register these extensions (whichever are needed) instead of manually registering the beans. That will also help cover the case of newly added CDI beans.

4. Last but not least, we should double check that all modules that should have `beans.xml` (and are subject to having the extension added) do have one because adding an extension into them might otherwise trigger "no archive behavior" as per [this spec part](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#bean_archive_full):

> An archive which:
> * contains a beans.xml file with the bean-discovery-mode of none, or,
> * contains a portable extension or a build compatible extension and no beans.xml file
>
> is not a bean archive.

PR #2144, opened 2023-04-29
RabbitMQ - Inject tracing in metadata

mprunier:
Hi,

The trace id is not properly shared, it doesn't reach the consumer.

I've added this small fix that works for me.

Before:
image

After:
image


ozangunalp:
Thank you @mprunier for noticing this, I think a more correct way to do this is to make `IncomingRabbitMQMessage` implement `MetadataInjectableMessage`. TracingUtils takes care of the rest. I've pushed a change to your PR branch

PR #2147, opened 2023-05-10
Implement incoming metadata injection

cescoffier:
This PR allows the injection of incoming metadata as method parameters.


cescoffier:
@ozangunalp well... I rewrote it. The integration into Quarkus proved to be slightly more complicated than expected. I had to collect more at build time and found a way to pass that data. I have a branch for Quarkus.

ozangunalp (approved):
Pushed a small fix to avoid looping over metadata objects every time we are getting a metadata by type

PR #2209, opened 2023-07-02
Fix the synchronization protocol in InternalChannelRegistry

cescoffier:
Use concurrentHashMap to avoid synchronized methods.

Fix https://github.com/smallrye/smallrye-reactive-messaging/issues/2207

PR #2211, opened 2023-07-04
Revert "Fix the synchronization protocol in InternalChannelRegistry"

cescoffier:
Reverts smallrye/smallrye-reactive-messaging#2209

We need to tidy-up the commit.


whitingjr:
@cescoffier thank you for that. Any chance this commit can get into the next 3.x tag ?

Issue #2207, opened 2023-06-29
Locking contention when lookup of Emitter by name.


whitingjr:
Hi,
 I have an application that is using smallrye in EAP7.4.0 patched with XP4.0.0. This includes Smallrye Reactive Messaging library 1.13.0.
`jboss-eap-7.4/modules/system/layers/microprofile/.overlays/layer-jboss-eap-xp-4.0.0.CP/io/smallrye/reactive/messaging/main/smallrye-reactive-messaging-provider-3.13.0.redhat-00001.jar`
Running an application called [First Responder](https://github.com/hpehl/first-responder-demo) I am seeing on a hot code path the lookup of an Emitter is causing contention on a synchronized method.
Emitter [io.smallrye.reactive.messaging.providers.impl.InternalChannelRegistry$Proxy$_$$_WeldClientProxy.getEmitter(String)](https://github.com/smallrye/smallrye-reactive-messaging/blob/3.13.0/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/impl/InternalChannelRegistry.java#L69-L73)

This is revealed by Java Mission Control profile when performance testing. This image shows the contention in the Lock Instances view.
![Screenshot from 2023-06-29 21-42-16](https://github.com/smallrye/smallrye-reactive-messaging/assets/708428/39b0101c-b956-4e82-8f2f-1e29f4965283)

In the critical section there is no concurrent access to data structures. If concerned with concurrent modification then surely making the field InternalChannelRegistry.emitters a CHM would add the protection necessary. I may be overlooking a compelling reason to make this method synchronized of course.
 This code [branch](https://github.com/whitingjr/smallrye-reactive-messaging/blob/improved-concurrent-emitter-lookup/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/impl/InternalChannelRegistry.java) shows what I am suggesting.

 Is there a compelling reason for the synchronized method ?

Testing is on a bare metal machine.
java -version
openjdk version "11.0.19" 2023-04-18 LTS
OpenJDK Runtime Environment (Red_Hat-11.0.19.0.7-2) (build 11.0.19+7-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-11.0.19.0.7-2) (build 11.0.19+7-LTS, mixed mode, sharing)

cescoffier:
[...]

PR #2212, opened 2023-07-05
Fix the synchronization protocol of the InternalChannelRegistry

cescoffier:
Fix https://github.com/smallrye/smallrye-reactive-messaging/issues/2207:

- Switched to CHM and CopyOnWriteArrayList instead of synchronized methods.

Based on this initial commit:
https://github.com/whitingjr/smallrye-reactive-messaging/commit/f9c5def7b71401d177b8b6d7d7c78b1d63245487


cescoffier:
@whitingjr are you ok with this one?

whitingjr:
@cescoffier yes that is fine. ty

ozangunalp (approved):

Issue #2198, opened 2023-06-15
Unordered blocking consumer requests too many messages


AndreasPetersen:
# Describe the bug
I made a [discussion post](https://github.com/smallrye/smallrye-reactive-messaging/discussions/2196) describing the problem:

<div type='discussions-op-text'>

<sup>Originally posted by **AndreasPetersen** June 14, 2023</sup>
I'm implementing a connector to Azure Service Bus using SmallRye Reactive Messaging 3.22.1. I want to be able to concurrently process a number of messages. I can do this by:

```
@Incoming("in-history")
@Blocking(ordered = false, value = "in-history")
public CompletionStage<Void> handleMessageFromAzure(Message<History> message) {
    try {
        logger.info("Processing message: " + message.getMetadata(MessageId.class).get().id()); // My connector adds a Azure Service Bus message id as metadata
        Thread.sleep(1000); // Do some blocking work
        return message.ack();
    } catch (InterruptedException e) {
        return message.nack(e);
    }
}
```

with the config:

```
smallrye.messaging.worker.in-history.max-concurrency=2
```

This works almost as expected. Two messages are indeed processed concurrently, but requests are constantly being made to the publisher. My connector looks like this:

```
    @Override
    public PublisherBuilder<? extends Message<?>> getPublisherBuilder(Config config) {
        ServiceBusReceiverAsyncClient serviceBusReceiverAsyncClient = getServiceBusReceiverAsyncClient(config); // This gives us an Azure Service Bus client that we can use to received messages from Azure
        Flux<ServiceBusReceivedMessage> flux = serviceBusReceiverAsyncClient.receiveMessages();

        var multi = Multi.createFrom().publisher(flux)
                .log() // I've added a log to see what's going on
                .onItem().invoke(serviceBusReceivedMessage -> logger.info("Received {}", serviceBusReceivedMessage.getMessageId()))                     
                .onItem().transform(azureMessage -> mapToReactiveMessage(azureMessage))

        return ReactiveStreams.fromPublisher(multi);
    }
```

When starting my Quarkus service I can see in the logs:

```
2023-06-14 17:27:50,196 INFO  [dk.ban.qua.rea.mes.asb.AzureServiceBusConnector] (boundedElastic-1) {} Received 9500cb1f42ed4b9a860ec35c4c89b172
[...]

PR #2219, opened 2023-07-10
Blocking improvements max-concurrency

ozangunalp:
Draft PR for discussions

Currently, Reactive Messaging controls the `@Blocking` method max concurrency level using the pool size attribute of the vertx worker pool created for the blocking execution.

However, there is a second aspect that affects the max concurrency which is the mutiny concurrency level when merging multiple Uni's into a Multi: https://smallrye.io/smallrye-mutiny/2.3.1/tutorials/transforming-items-asynchronously/#transforming-items-from-multi-the-merge-vs-concatenate-dilemma

This last control how many items will be requested upstream in parallel and effective only when blocking with ordered=false.

This PR changes how the Mutiny merge concurrency is configured : 
- When available the `max-concurrency` is used to configure the Mutiny merge concurrency.
- If not it falls back to the current default value of `Queues.BUFFER_S`.

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
@@ -35,8 +35,8 @@ import io.vertx.mutiny.core.WorkerExecutor;
 
 @ApplicationScoped
 public class WorkerPoolRegistry {
-    private static final String WORKER_CONFIG_PREFIX = "smallrye.messaging.worker";
-    private static final String WORKER_CONCURRENCY = "max-concurrency";
+    public static final String WORKER_CONFIG_PREFIX = "smallrye.messaging.worker";
+    public static final String WORKER_CONCURRENCY = "max-concurrency";
 
     @Inject
     Instance<ExecutionHolder> executionHolder;
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/AbstractMediator.java
@@ -44,6 +44,7 @@ public abstract class AbstractMediator {
     protected HealthCenter health;
     private Instance<MessageConverter> converters;
     private Instance<KeyValueExtractor> extractors;
+    private int maxConcurrency;
 
     public AbstractMediator(MediatorConfiguration configuration) {
         this.configuration = configuration;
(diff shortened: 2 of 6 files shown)


ozangunalp:
Fixes https://github.com/smallrye/smallrye-reactive-messaging/issues/2198 and related to  https://github.com/smallrye/smallrye-reactive-messaging/discussions/2196

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -36,7 +38,10 @@
 @ApplicationScoped
 public class WorkerPoolRegistry {
     private static final String WORKER_CONFIG_PREFIX = "smallrye.messaging.worker";
+    @Deprecated(since = "4.8.0")
     private static final String WORKER_CONCURRENCY = "max-concurrency";
+    private static final String WORKER_POOL_SIZE = "pool-size";
```
cescoffier, line 43: actually, max-concurrency was better.
virtual threads are not pooled, so pool size does not make sense.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/i18n/ProviderLogging.java
```diff
@@ -135,4 +135,8 @@ public interface ProviderLogging extends BasicLogger {
 
     @Message(id = 240, value = "Could not find an SSLContext bean with the @Identifier=%s")
     IllegalStateException couldFindSslContextWithIdentifier(String sslContextIdentifier);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 241, value = "Use of `smallrye.messaging.worker.{pool-name}.max-concurrency` is deprecated, instead use `@Blocking(maxConcurrency)` to adjust the maximum concurrency level and `smallrye.messaging.worker.{pool-name}.pool-size` for worker pool size")
```
cescoffier, line 140: `(maxConcurrency=...)`

cescoffier (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
ozangunalp: I wasn't clear enough. With this change, there wouldn't be a need for a `pool-size` or `max-concurrency` config to limit the concurrency with virtual threads. 

For example:
```java
@RunOnVirtualThread
@Blocking(maxConcurrency=1000)
@Incoming("channel")
void process(String payload) {
    //
}
```

For Vert.x workers:
```java
@Blocking(value="my-custom-pool", maxConcurrency=10, ordered=false)
@Incoming("channel")
void process(String payload) {
    //
}
```
with
```properties
smallrye.messaging.worker.my-custom-pool.pool-size=10
smallrye.messaging.worker.my-custom-pool.max-execute-time=30s
```

cescoffier (approved):

PR #2223, opened 2023-07-13
Add beans.xml to the inmemory connector

kabir:
We're trying to include this in WildFly, and are seeing some problems with trying to add a beans.xml on our side.


ozangunalp:
@cescoffier any objections to this?

cescoffier:
No, it should be fine.

ozangunalp (approved):

PR #2231, opened 2023-07-25
Provide a utility method to coordinate (negative) acknowledgement when multiple messages are created from one

### LINKED ISSUE #2191: Provide utility methods to coordinate (negative) acknowledgement when multiple messages is created from a single one

cescoffier:
Fix #2191

### DIFF
--- a/api/src/main/java/io/smallrye/reactive/messaging/Messages.java
+++ b/api/src/main/java/io/smallrye/reactive/messaging/Messages.java
@@ -1,11 +1,14 @@
 package io.smallrye.reactive.messaging;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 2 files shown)


cescoffier:
@ozangunalp that will be handy to handle @Outgoings.
I plan to use it to coordinate acknowledgement when doing broadcast.

api/src/main/java/io/smallrye/reactive/messaging/Messages.java
```diff
@@ -121,4 +174,103 @@ private static Metadata merge(Metadata first, Metadata second) {
...
+         * Passed the chained messages.
+         * The messages are not modified, but should not be used afterward, and should be replaced by the messages contained
+         * in the returned list.
+         * This method preserve the order. So, the first message corresponds to the first message in the returned list.
+         * The message from the returned list have the necessary logic to chain the ack/nack signals and the copied metadata.
+         *
+         * @param messages the chained messages, must not be empty, must not be null, must not contain null
+         * @return the list of modified messages
+         */
+        public List<Message<?>> with(Message<?>... messages) {
```
ozangunalp, line 242: We may think about wrapping the return into a marker type to integrate with `@Outgoings`. 

If multiple messages are returned we redirect them by order of declaration. If a single message is returned it'd be considered as broadcast.

ozangunalp (approved):

PR #2239, opened 2023-08-01
RabbitMQ requeue with additional nack metadata

### LINKED ISSUE #2200: RabbitMQ - Missing 'requeue' strategy on rejection/failure of rabbitmq message consumer
**Introduction:** 

As per the RabbitMQ specifications (as mentioned on https://www.rabbitmq.com/nack.html),

> The [AMQP 0-9-1 specification](https://www.rabbitmq.com/specification.html) defines the basic.reject method that allows clients to reject individual, delivered messages, instructing the broker to either discard them or **requeue them**.

The example also demonstrates the plain Java way of requeuing the consumed message with negative delivery to the broker.

> GetResponse gr = channel.basicGet("some.queue", false);
channel.basicNack(gr.getEnvelope().getDeliveryTag(), false, true);

**Problem Statement:**

The current implementation of by RabbitMQ broker by this project does not honor the requeue strategy.

The `failure-strategy` supports only 3 modes: (https://quarkus.io/guides/rabbitmq-reference)

> The failure strategy to apply when a RabbitMQ message is nacked. Accepted values are fail, accept, reject (default).

While digging down more into source code, I found:

````  
    /**
     * Rejects the message by nack'ing with requeue=false; this will either discard the message for good or
     * (if a DLQ has been set up) send it to the DLQ.
     *
     * @param reason the cause of the rejection, which must not be null
     */
    public void rejectMessage(Throwable reason) {
        holder.getNack(this.deliveryTag, false).apply(reason);
    }
````

The above implementer only encouraged the Dead Letter Queue routing for redelivery, but that is not the thing that everyone might want.

**Proposals:**

Can we introduce one more failure-strategy > requeue?

That can do almost exact same thing but instead of passing the `requeue=false`, it will send `requeue=true`.

Or please suggest any other approach to overcome this?

michaelkoetter:
Supports specifying the 'requeue' flag using additional nack metadata. Should fix #2200 .

### DIFF
--- a/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
+++ b/documentation/src/main/docs/rabbitmq/receiving-messages-from-rabbitmq.md
@@ -173,12 +173,31 @@ controlled by the `failure-strategy` channel setting:
 -   `reject` - this strategy marks the RabbitMQ message as rejected
     (default). The processing continues with the next message.
 
+-   `requeue` - this strategy marks the RabbitMQ message as rejected
+    with requeue flag to true. The processing continues with the next message,
+    but the requeued message will be redelivered to the consumer.
+
+When using `dead-letter-queue`, it is also possible to change some
+metadata of the record that is sent to the dead letter topic. To do
+that, use the `Message.nack(Throwable, Metadata)` method:
+
+The RabbitMQ reject `requeue` flag can be controlled on different failure strategies
+using the {{ javadoc('io.smallrye.reactive.messaging.rabbitmq.RabbitMQRejectMetadata') }}.
+To do that, use the `Message.nack(Throwable, Metadata)` method by including the
+`RabbitMQRejectMetadata` metadata with `requeue` to `true`.
+
+``` java
+{{ insert('rabbitmq/inbound/RabbitMQRejectMetadataExample.java', 'code') }}
+```
+
 !!!warning "Experimental"
     `RabbitMQFailureHandler` is experimental and APIs are subject to change in the future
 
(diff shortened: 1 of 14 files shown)


cescoffier (changes requested):
LGTM

It needs a bit of documentation.
Also, did you try accessing the metadata from the message itself instead of passing it around?

michaelkoetter:
> It needs a bit of documentation. 

You mean in the markdown docs?

> Also, did you try accessing the metadata from the message itself instead of passing it around?

I figured the purpose of `nack(Throwable reason, Metadata metadata)` was to pass metadata that's relevant for nack'ing the message. From javadoc:
> Additional metadata may be provided that the connector can use when nacking the message. The interpretation of metadata is connector-specific.

Sure we could simply add the additional metadata to the message, but then what's the point in having that extra signature for `nack` 😄

ozangunalp (changes requested):
Thanks for this! It looks good to me too. It'd be great to add some explanation to the `receiving-messages-from-rabbitmq.md#failure-management`

On the metadata stuff, we've been kind of entertaining the idea of reworking the message composability and maybe deprecating that method `nack(Throwable, Metadata)`.
 
The current issue is if the message is intercepted in a way to add metadata etc. nack(Throwable, Metadata) method wasn't really composable, and passed metadata would be completely ignored.

It turns out you've discovered that it is not that easy there are already other problems to achieve this.

ozangunalp:
Rebased to main, reviewed the test case and added documentation

cescoffier (approved):

cescoffier:
It seems that the CI is unhappy.

ozangunalp (approved):

PR #2242, opened 2023-08-02
Generate `smallrye-reactive-messaging-bom`

### LINKED ISSUE #2241: Provide a BOM
It would be nice if we had a `smallrye-reactive-messaging-bom` containing all modules to avoid redeclaring the same version in multiple dependencies

Implementation idea: https://github.com/smallrye/smallrye-common/blob/78d5c1afd60472d97f31bd1f3a96eef41f9b0a6f/pom.xml#L188-L212

gastaldi:
- Fixes #2241

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,7 @@
     <yasson.version>2.0.3</yasson.version>
 
     <jandex-maven-plugin.version>1.2.3</jandex-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


ozangunalp:
Nice one! I didn't know we used it in smallrye-common. How does it integrate with deploy?

gastaldi:
It generates the artifact during the build and adds to the reactor as a new artifact

ozangunalp (approved):
LGTM!

PR #2244, opened 2023-08-03
Message metadata propagation with ack and nack

### LINKED ISSUE #334: Allow observing message acknowledgment
This would allow decorating a message and being aware when the message is acknowledged.

ozangunalp:
Currently, message composition does not propagate message metadata. 
This makes message interception (through decorators etc.) practically unfeasible.

With this change `Message` acks and nacks receive an additional `Metadata` parameter.
Existing methods for acks/nacks and composition continue to work by delegating to the new ack/nack functions.
The `Message` interface from the `api` package is still compatible with the MP Messaging Spec `Message` interface, verified by the `MessageSpecCompatibilityTest` test.

Other changes related to message ack/nack composition :
- Deprecate `PublisherDecorator#decorate(String, String, boolean)`
- Introduce `IncomingInterceptor`
- Change message interceptor method names: 
  - `OutgoingInterceptor#onMessage` -> `beforeMessageSend`
  - `IncomingInterceptor#afterMessageReceive`
- Message Observation API: Allows a `MessageObservationCollector` to observe message creation, acks, nacks, and message processing duration. 

Resolves #334

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -49,6 +49,31 @@
       </dependency>
     </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>process-test-resources</phase>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.eclipse.microprofile.reactive.messaging</groupId>
+                  <artifactId>microprofile-reactive-messaging-api</artifactId>
+                </artifactItem>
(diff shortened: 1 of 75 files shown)


cescoffier (approved):
A few words of caution:

- We need to make sure that the previous methods (where the ack is a function taking no parameter is still supported)
- We need to make sure we pass the TCK.

ozangunalp:
This is the reason I didn't push this faster. 
TCK pass, and existing Message tests as well. But I want to take some time to check the test coverage and see which cases are covered.

cescoffier:
The commits might need to be refactored a bit.

cescoffier (approved):

PR #2272, opened 2023-08-22
Fix the split package between the smallrye-reactive-messaging-jackson and smallrye-reactive-messaging-jsonb modules

### LINKED ISSUE #2261: Split package warning with -jackson or -jsonb libraries with Quarkus
When using `smallrye-reactive-messaging-jackson` or `smallrye-reactive-messaging-jsonb` with Quarkus you get a split package warning because both `JacksonMapping` and/or `JsonBMapping` are in the same package `io.smallrye.reactive.messaging.json` as `JsonMapping` but in different artifacts.

cescoffier:
Fix #2261


cescoffier:
As it's not a public API, it's not a breaking change. More a cosmetic thing.

am-tribia:
Why is smallrye-reactive-messaging-jackson.jar missing /META-INF/beans.xml file? I get the following error when I use 
smallrye-reactive-messaging-jms:4.10.0 and smallrye-reactive-messaging-jackson:4.10.0

```ERROR [io.sma.rea.mes.jms] (executor-thread-1) SRMSG15800: Unable to send message to JMS: java.lang.NullPointerException: Cannot invoke "io.smallrye.reactive.messaging.json.JsonMapping.toJson(Object)" because "this.jsonMapping" is null```

PR #2323, opened 2023-10-11
Fix PauseResumeTest that became flaky because starting from Kafka 3.6.0 

### LINKED ISSUE #2322: `PauseResumeTest.testRebalanceDuringPaused` is failing in CI since Kafka 3.6.0 update
Seems to always fail on CI and flaky locally.

ozangunalp:
MockConsumer calls rebalance listeners

Fixes #2322


cescoffier:
Since when does the mock consumer call the rebalance listeners? That was one of their big limitations. 
It seems that we can now improve how we test our commit strategies.

cescoffier (approved):

ozangunalp:
It is since 3.6.0. We already called rebalance listeners manually for some test cases using MockConsumer.
It is still not easy to use MockConsumer, for example, it doesn't take into account consumer configs.

Issue #2332, opened 2023-10-18
Missing metrics for channels associated with Emitters


abutch3r:
We have created an emitter that is sending messages to a kafka deployment via our own connector and we are pulling messages from Kafka on a separate channel. When we look at the metrics that are being produced. We have a single entry for mp_messaging that is for the incoming channel. But we are not seeing an entry at all the emitter channel.

When reviewing the [specs](https://download.eclipse.org/microprofile/microprofile-reactive-messaging-3.0/microprofile-reactive-messaging-spec-3.0.html#_metrics) for metrics. This part: `metrics are produced for each channel declared by the application`. There is no statement about how a channel associated with an emitter.

Our scenario can be seen as the following simple flows:
Emitter -> channelA -> Kafka connector -> Kafka
Kafka -> Kafka Connector -> channelB -> `@Incoming`

We are seeing metrics only for channelB.

When we used an `@Outgoing` and `@Incoming` we got metrics to both. so would indicate that it is the fact we had the xhannel  associated with an Emitter that meant we were not seeing any metrics.

cescoffier:
That's a bug, we should report these messages too.

PR #2336, opened 2023-10-23
Let emitters report message metrics

cescoffier:
Fix https://github.com/smallrye/smallrye-reactive-messaging/issues/2332

I've decided to use the beans directly and not apply all the decorators. The reason is that I think we need to change the decorator interface to have an `isEmitter` flag so the decorator knows if it's an emitter or not, like for connectors.


cescoffier:
Superseded by https://github.com/smallrye/smallrye-reactive-messaging/pull/2342.

PR #2339, opened 2023-10-23
Include correct metadata object in IncomingRabbitMQMessage injectMetadata

### LINKED ISSUE #2338: RabbitMQ connector no longer propagates tracing
The culprit is f6cfcc8caad47cfd6c926e971aaf4f8d76696362 and on main since 4.6.0

ozangunalp:
Fixes #2338

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
@@ -264,7 +264,7 @@ public class IncomingRabbitMQMessage<T> implements ContextAwareMessage<T>, Metad
 
     @Override
     public synchronized void injectMetadata(Object metadataObject) {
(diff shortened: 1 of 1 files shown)


cescoffier (approved):

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMessage.java
```diff
@@ -263,7 +263,7 @@ public io.vertx.mutiny.rabbitmq.RabbitMQMessage getRabbitMQMessage() {
     }
 
     @Override
-    public synchronized void injectMetadata(Object metadataObject) {
+    public synchronized void injectMetadata(Object metadata) {
```
wabrit, line 266: Wouldn't be better to retain the name of the existing parameter (`metadataObject`) and instead change the body i.e.

```java
public synchronized void injectMetadata(Object metadataObject) {
  this.metadata = this.metadata.with(metadataObject);
}
```

This avoids having a parameter name that shadows that of a class member.

PR #2342, opened 2023-10-23
Decorate emitters with PublisherDecorators

### LINKED ISSUE #2332: Missing metrics for channels associated with Emitters

ozangunalp:
Resolves #2332 
May supersede #2336

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
@@ -22,6 +22,7 @@ import io.smallrye.reactive.messaging.EmitterConfiguration;
 import io.smallrye.reactive.messaging.EmitterFactory;
 import io.smallrye.reactive.messaging.MediatorConfiguration;
 import io.smallrye.reactive.messaging.MessagePublisherProvider;
+import io.smallrye.reactive.messaging.PublisherDecorator;
 import io.smallrye.reactive.messaging.SubscriberDecorator;
 import io.smallrye.reactive.messaging.annotations.EmitterFactoryFor;
 import io.smallrye.reactive.messaging.annotations.Merge;
@@ -29,6 +30,7 @@ import io.smallrye.reactive.messaging.providers.AbstractMediator;
 import io.smallrye.reactive.messaging.providers.extension.*;
 import io.smallrye.reactive.messaging.providers.helpers.MultiUtils;
 import io.smallrye.reactive.messaging.providers.i18n.ProviderLogging;
+import io.smallrye.reactive.messaging.providers.locals.ContextDecorator;
 
 @ApplicationScoped
 public class Wiring {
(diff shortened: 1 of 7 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
```diff
@@ -504,7 +515,13 @@ public void materialize(ChannelRegistry registry) {
         private <T extends MessagePublisherProvider<?>> void registerEmitter(ChannelRegistry registry, int def) {
             EmitterFactory<?> emitterFactory = getEmitterFactory(configuration.emitterType());
             T emitter = (T) emitterFactory.createEmitter(configuration, def);
-            Publisher<? extends Message<?>> publisher = emitter.getPublisher();
+            Multi<? extends Message<?>> publisher = Multi.createFrom().publisher(emitter.getPublisher());
+            for (PublisherDecorator decorator : getSortedInstances(decorators)) {
+                // TODO do not apply ContextDecorator here, it messes with custom emitter transactions
```
cescoffier, line 520: Any reason why?

cescoffier: If we do not apply it, how to we associate the context to the message?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
```diff
@@ -504,7 +515,13 @@ public void materialize(ChannelRegistry registry) {
...
+                if (!(decorator instanceof ContextDecorator)) {
+                    publisher = decorator.decorate(publisher, configuration.name(), false);
```
cescoffier, line 522: So, you consider the emitter as a "not connector", which is exact. However, I was wondering if we need to be specific in this case and introduce a second flag (I'm not a big fan of signatures containing two exclusive flags, having an enum would be better).

However, do we need to know when it's an emitter?

cescoffier (changes requested):

ozangunalp:
@cescoffier pushed a fix for Kafka transactions. It appears that applying ContextDecorator there changes the tx flow order, flushing and commiting the Tx before records are sent. Now it captures sent record Uni's and makes sure that all of them are sent before flusing and committing.

ozangunalp:
@cescoffier shall we get this in ?

cescoffier:
Yes!

PR #2356, opened 2023-10-31
added in the metadata the input received and not itself

Francesco2498:


Francesco2498:
The injectMetadata method was wrongly adding metadata to itself. Now with the fix the object received as input is added to the metadata.

cescoffier:
@ozangunalp don't we have a fix for this already?

ozangunalp:
Duplicate of #2339

PR #2357, opened 2023-10-31
Seperate afterDeploymentValidation into two submethods

benjamin-confino:
Hello

I have been working on getting reactive messaging to work with [InstantOn](https://openliberty.io/docs/latest/instanton.html) in open liberty. I have a proof of concept and have determined that if I prevent the line `mediatorManager.start()` in `ReactiveMessagingExtension` from running before the checkpoint, and instead run it after the restore, everything just works.

However because `mediatorManager.start()` is invoked as part of a larger method containing code that ideally will run before the checkpoint I had to modify that class locally. I would like to submit a simple change that makes  `ReactiveMessagingExtension.afterDeploymentValidation` call two subordinate methods. Then in my code I can simply extend, rather than modify, `ReactiveMessagingExtension` and call the subordinate methods directly at the appropriate times.

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/ReactiveMessagingExtension.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/extension/ReactiveMessagingExtension.java
@@ -105,7 +105,12 @@ public class ReactiveMessagingExtension implements Extension {
     }
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
-    void afterDeploymentValidation(@Observes AfterDeploymentValidation done, BeanManager beanManager) {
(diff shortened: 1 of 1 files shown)


cescoffier (approved):
The change looks good, but before merging, we need to see if it does change something in Quarkus. I don't believe it does (as we do not use this method if I remember correctly).

@ozangunalp do you know?

Ladicek:
I'm pretty sure the entire `ReactiveMessagingExtension` is ignored in Quarkus, because it's a CDI portable extension. The Quarkus extension for Reactive Messaging has to implement the entire logic on its own.

ozangunalp (approved):
This looks good to me too. 
As long as MediatorManager start method is there there isn't any impact on Quarkus integration. Indeed the CDI Extension is not used in Quarkus

PR #2366, opened 2023-11-06
Fix the synchronization protocol of the InternalChannelRegistry

xstefank:
Switched to CHM and CopyOnWriteArrayList instead of synchronized methods.

Based on this initial commit:
https://github.com/whitingjr/smallrye-reactive-messaging/commit/f9c5def7b71401d177b8b6d7d7c78b1d63245487


ozangunalp:
Backport to 3.13.x

ozangunalp (approved):

PR #2370, opened 2023-11-07
Run emitter publisher decorators after registering the emitter to channel registry

ozangunalp:
Fix after #2342

This way decorators can access channel registry to check if the channel is an emitter or not.
Additional changes :
- PublisherDecorator calls are made to the `decorate(Multi<Message>, List<String>, boolean)` method with list of channel names.
- `EmitterImpl#send(Message)` method now captures the context metadata – Don't know why it wasn't done before. Looks like it was forgotten.


ozangunalp:
Blocker for OTel 1.31 integration for Quarkus

cescoffier (approved):

PR #2382, opened 2023-11-20
Incoming channel concurrency

ozangunalp:
Create multiple copies of the configured channel using, ex.
`mp.messaging.incoming.data.concurrency=4` config,
in order to increase the processing concurrency on non-blocking processing

Attributes per copy-channels can be configured using the channel name with `$` separator, ex. `mp.messaging.incoming.data$1.attribute=...`
Each inbound connector channel would dispatch messages accordingly. Connectors based on Vert.x will create separate event loops per channel to dispatch messages. Without the `concurrency` attribute the current behaviour won't change.


cescoffier:
I had a quick look - it looks good.

documentation/mkdocs.yml
```diff
@@ -102,6 +102,7 @@ plugins:
         version_selector: true
         css_dir: css
         javascript_dir: javascript
+  - print-site
```
cescoffier, line 105: what does it do?

ozangunalp: It creates a printable single page doc in path `/print_page`

cescoffier (approved):

PR #2404, opened 2023-12-08
Bump to Mutiny 2.5.2

jponge:


jponge:
Not sure about these errors, I'll let people here have a look wether Mutiny introduces bugs or not 😉

cescoffier:
@jponge, I will ask @ozangunalp to check, but yes, something changed. I got the same tests fail twice.

ozangunalp:
Still continuing investigation, I can definitely reproduce breaking tests but don't have any leads yet.

cescoffier:
Seems to be related to `transformToUniAndConcatenate`.

jponge:
2.5.3 coming with fixes

PR #2409, opened 2023-12-12
Kafka Request Reply emitter and manual partition assignment

### LINKED ISSUE #988: Request for Synchronous KafkaResponse
With Message Communication we want to prevent direct communication between the services.
Sometimes there is the case where we need inforamtion from a service and need to continue working on with the response.

Would sth. similar to the KafkaReplyTemplate be an option to be part of Smallrye Reactive Messaging?

ozangunalp:
On #2406 

Kafka connector `assign-seek` attribute for manual partition assignment of Kafka consumers. 
Optionally receives the offset portion to seek the partition to an offset after assignment.

Kafka Request/Reply emitter is a custom emitter that allows sending request messages to a Kafka topic and receiving back a reply message. It builds on top of a regular outgoing channel with a custom emitter that creates an internal Kafka consumer (Kafka incoming channel through KafkaSource). The outgoing record headers contain the **correlation id** against which the reply record will be matched, the **reply topic** and optionally **reply partition** from where replies will be expected.

A summary of features implemented:
- Default headers work with the default outgoing channel, therefore implementing a server is simplified
- Headers are configurable
- Default reply timeout is set to 5 seconds and configurable.
- When the internal consumer `auto.offset.reset` is set to `latest` (the default) the emitter subscription waits on assignments of topic-partitions expected to be listened, in order to not miss any replies.
- Correlation id handling is configurable through a bean configuration
- Reply error mapping is configurable through a bean configuration
- Incoming reply payloads are converted using `MessageConverter`s based on the injected emitter payload type. A limitation of this is that the same emitter channel injected to different signatures would not work.
- If the graceful shutdown is configured (default true) the closing waits for pending replies to be returned.

Closes #988


documentation/src/main/docs/kafka/receiving-kafka-records.md
```diff
@@ -618,6 +618,38 @@ Conversely, if the processing throws an exception, all messages are
...
+
+assigns the consumer to:
+- Partition 0 of topic 'topic1', setting the initial position at offset 10.
+- Partition 1 of topic 'topic2', setting the initial position at offset 20.
+
+The topic, partition, and offset in each triplet can have the following variations:
+- If the topic is omitted, the configured `topic` will be used.
+- If the offset is omitted, partitions are assigned to the consumer but won't be seeked to offset.
+- If offset is 0, it seeks to the beginning of the topic-partition.
+- If offset is -1, it seeks to the end of the topic-partition.
```
cescoffier, line 650: I was wondering if omitting the offset would mean 'last'. 

Actually it could mean first too.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
```diff
@@ -63,6 +63,7 @@
 
 @ConnectorAttribute(name = "topics", type = "string", direction = Direction.INCOMING, description = "A comma-separating list of topics to be consumed. Cannot be used with the `topic` or `pattern` properties")
 @ConnectorAttribute(name = "pattern", type = "boolean", direction = Direction.INCOMING, description = "Indicate that the `topic` property is a regular expression. Must be used with the `topic` property. Cannot be used with the `topics` property", defaultValue = "false")
+@ConnectorAttribute(name = "assign-seek", type = "string", direction = Direction.INCOMING, description = "Assign partitions and optionally seek to offsets, instead of subscribing to topics. A comma-separating list of triplets in form of `<topic>:|<partition>|:<offset>` to assign statically to the consumer and seek to the given offsets. Offset `0` seeks to beginning and offset `-1` seeks to the end of the topic-partition. If the topic is omitted the configured topic will be used. If the offset is omitted partitions are assigned to the consumer but won't be seeked to offset.")
```
cescoffier, line 66: We need to remove that key when we cleanup the conf (before configuring the Kafka objects)

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaConnector.java
ozangunalp: Indeed, I removed all producer config starting with `reply` as well.

documentation/src/main/docs/kafka/receiving-kafka-records.md
ozangunalp: If the offset is omitted, seek is not called, so `auto.offset.reset` is fully operational.

cescoffier: Cool!

cescoffier (approved):

PR #2485, opened 2024-02-13
Add build workflow to main branch to run tests with podman

ozangunalp:
Configures podman with `service_timeout=0` to avoid testcontainer failures

### DIFF
--- a/.github/workflows/build-podman.yml
+++ b/.github/workflows/build-podman.yml
@@ -1,0 +1,76 @@
+---
+name: Continuous Integration Build – Podman
+
+on:
+    push:
+        branches:
+            - main
+        paths-ignore:
+            - '.github/**'
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        continue-on-error: ${{ matrix.java.continue-on-error == true }}
+        strategy:
+            matrix:
+                java: [
+                    {
+                        'name': '',
+                        'version': '11',
+                        'build_opts': '-Drevapi.skip=true'
+                    },
+                    {
+                        'name': '',
+                        'version': '17',
+                    },
+                    {
+                        'name': '',
+                        'version': '21',
+                    },
+                    {
+                        'version': '11',
+                        'build_opts': '-Pslow-flaky-tests -Drevapi.skip=true',
+                        'name': 'Slow and Flaky tests',
+                        'continue-on-error': true
+                    }
+                ]
+        name: Build with Java ${{ matrix.java.version }} ${{ matrix.java.name }}
+        steps:
+            -   name: Setup docker host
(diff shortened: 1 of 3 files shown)


ozangunalp:
Run here: https://github.com/ozangunalp/smallrye-reactive-messaging/actions/runs/7889471670

.github/workflows/build-podman.yml
```diff
@@ -0,0 +1,76 @@
...
+                    sudo apt-get -y purge podman
+                    sudo mkdir -p /etc/apt/keyrings
+                    curl -fsSL https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key | gpg --dearmor | sudo tee /etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg > /dev/null
+                    echo \
+                      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg] \
+                      https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
+                      | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
+                    sudo apt-get update -qq
+                    sudo apt-get -qq -y install podman
+                    sudo bash -c "echo -e '[engine]\nservice_timeout=0' >> /etc/containers/containers.conf"
```
maxandersen, line 53: what does that service_timeout change? does it mean wait forever? 

if it does - is the reason we need to do this that podman is slower or that the timeouts is lower in podman than in docker?

ozangunalp: This is essentially a workaround for a bug in the docker java client: https://github.com/testcontainers/testcontainers-java/issues/7310
This config controls the timeout to drop the connection between the client and the server, when set to 0 podman never drops the connection. Like kroxylicious folks we also experience the issue most commonly on network create: https://github.com/containers/podman/issues/17640

maxandersen: gotcha - reading the issue linked sounds to me one could argue docker java client could handle both better but could also say podman could have a different timeout for the docker interface to be more compatible. 

anyhow would setting this timeout to 3 minutes not be better to avoid any continued hanging but still effectively result in the setup/test to pass? 

Asking this because great we can get it to pass here but how about users of podman desktop using podman as users - should we recommend the podman desktop to set this timeout to make things work for not just Quarkus but any testcontainer users ?

cc @n1hility

ozangunalp: I guess what this will effectively do is keep the podman process running even though a client holds an idle connection to the socket. What I can't explain is, on podman logs I see pings from connected clients, at which point the ping no longer works to keep the connection alive?

jponge (approved):

Issue #2486, opened 2024-02-14
NullPointerException when a non recoverable KafkaException is thrown from the Kafka producer


vietk:
Hello

I wanted to test the [ProducerInterceptor](https://smallrye.io/smallrye-reactive-messaging/latest/kafka/writing-kafka-records/#producer-interceptors) feature. 
When I tried a message that exceeds the topic limit (generate a KafkaException) I have the following error
```shell
22:51:32,308 TRACE [io.smallrye.reactive.messaging.kafka] - SRMSG18275: Error caught in producer interceptor `onAcknowledge` for channel output-record
java.lang.NullPointerException
	at io.smallrye.reactive.messaging.kafka.impl.ReactiveKafkaProducer.getRecordMetadataForFailure(ReactiveKafkaProducer.java:360)
	at io.smallrye.reactive.messaging.kafka.impl.ReactiveKafkaProducer.interceptOnAcknowledge(ReactiveKafkaProducer.java:351)
	at io.smallrye.reactive.messaging.kafka.impl.ReactiveKafkaProducer.lambda$send$3(ReactiveKafkaProducer.java:165)
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:1062)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:947)
	at io.smallrye.reactive.messaging.kafka.impl.ReactiveKafkaProducer.lambda$send$4(ReactiveKafkaProducer.java:164)
	at io.smallrye.mutiny.operators.uni.builders.UniCreateWithEmitter.subscribe(UniCreateWithEmitter.java:22)
	at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
	at io.smallrye.mutiny.operators.uni.UniOnItemTransformToUni$UniOnItemTransformToUniProcessor.performInnerSubscription(UniOnItemTransformToUni.java:81)
	at io.smallrye.mutiny.operators.uni.UniOnItemTransformToUni$UniOnItemTransformToUniProcessor.onItem(UniOnItemTransformToUni.java:57)
	at io.smallrye.mutiny.operators.uni.UniOperatorProcessor.onItem(UniOperatorProcessor.java:47)
	at io.smallrye.mutiny.operators.uni.UniMemoizeOp.forwardTo(UniMemoizeOp.java:123)
[...]

PR #2487, opened 2024-02-14
fix: failure metadata creation when partition is null

vietk:
Fix for #2486 

Prevent autoboxing failure when a Kafka exception is thrown from the internal producer and the ProducerRecord partition is null.
Change the log level of onAckownledgement error to INFO rather than TRACE


ozangunalp (approved):
Nice catch! Thank you for the fix

PR #2495, opened 2024-02-22
Kafka request reply same topic

ozangunalp:


ozangunalp:
@iLem0n would have the means to test this solution?

ozangunalp:
@iLem0n would you have the time to check if this can work for you ?

iLem0n:
@ozangunalp Sorry for the late response. This seems to work well on my side.

ozangunalp:
Thanks!

PR #2514, opened 2024-03-07
Take 2 : Pulsar force schema fetch on pulsar-client-thread for all schema using AbstractMultiVersionReader

### LINKED ISSUE #2513: Deadlock when using Avro schema with pulsar
I reported the issue https://github.com/smallrye/smallrye-reactive-messaging/issues/2493#issue-2146960922

which has been fixed by changing the condition to fetch schema that requires fetchnig. Unfotunately AvroSchema does not fulfill this condition and the issue still occurs.

ozangunalp:
Fixes #2513

### DIFF
--- a/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/PulsarIncomingChannel.java
+++ b/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/PulsarIncomingChannel.java
@@ -13,6 +13,10 @@ import java.util.stream.Collectors;
 import org.apache.pulsar.client.api.*;
 import org.apache.pulsar.client.impl.MultiplierRedeliveryBackoff;
 import org.apache.pulsar.client.impl.conf.ConsumerConfigurationData;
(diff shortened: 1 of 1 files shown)


smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/PulsarIncomingChannel.java
```diff
@@ -112,7 +116,7 @@ public PulsarIncomingChannel(PulsarClient client, Vertx vertx, Schema<T> schema,
                     .until(m -> closed.get())
                     .plug(msgMulti -> {
                         // Calling getValue on the pulsar-client-internal thread to make sure the SchemaInfo is fetched
-                        if (schema.requireFetchingSchemaInfo()) {
+                        if (schemaRequiresBLockingFetch(schema)) {
```
michalcukierman, line 119: Looks like a typo in `BLocking`

ozangunalp: Indeed 😞 
Thanks!

PR #2515, opened 2024-03-07
GitHub Actions release workflow

ozangunalp:
Inspired by the Mutiny release workflow using justfile

- Adds maven wrapper
- JReleaser for release notes
- Changes the development version to 999-SNAPSHOT


cescoffier (approved):
LGTM - the first release is going to be fun :-D

pom.xml
```diff
@@ -837,5 +844,48 @@
...
+            <configuration>
+              <projectVersionPattern>SEMVER</projectVersionPattern>
+              <jreleaser>
+                <release>
+                  <github>
+                    <owner>smallrye</owner>
+                    <name>smallrye-reactive-messaging</name>
+                    <overwrite>true</overwrite>
+                    <changelog>
+                      <preset>conventional-commits</preset>
```
jponge, line 865: You might want to check the changelog output `jreleaser:changelog`). Another option is to adopt conventional commits going forward.

ozangunalp: Even without adopting conventional commits it seemed ok. But yes that's an option.

ozangunalp:
> LGTM - the first release is going to be fun :-D

Piece of cake :)

jponge (dismissed):
See my comment on the changelog generation.

Other than that I recommend testing this on release candidates, because I also predict a few fun issues 🤣  (sadly, there's no proper way to test GH Actions, especially those who perform release tasks)

jponge:
> > LGTM - the first release is going to be fun :-D
> 
> Piece of cake :)

Famous last words

xstefank (approved):
LGTM, but note we don't use jreleaser yet in health. But all release steps are there.

Issue #2519, opened 2024-03-09
Pulsar Traces are not always propagated 


michalcukierman:
This is a regression introduced by #2456.
Tracing properties from incoming Pulsar messages are not propagated to outgoing Pulsar messages.

The change is in PulsarTrace.java:

before:
```java        
        public Builder withProperties(Map<String, String> properties) {
            this.properties = properties;
            return this;
        }
```

after:
```java        
        public Builder withProperties(Map<String, String> properties) {
            if (properties != null) {
                this.properties = new HashMap<>(properties);
            }
            return this;
        }
```

It was introduced to ensure that the map is modifiable because it's required during producing outgoing traces.

Unfortunately, it introduced a regression in PulsarOutgoingChannel. We expect that the changes in properties will be propagated into the state of the trace:

```java
            if (tracingEnabled) {
                Map<String, String> properties = new HashMap<>(); // Single instance has to be used in Trace and MessageBuilder
                TracingUtils.traceOutgoing(instrumenter, message, new PulsarTrace.Builder()
                        .withProperties(properties) // REGRESSION HERE, we create a new instance of a HashMap
                        .withTopic(producer.getTopic())
                        .build());
                messageBuilder.properties(properties);
            }
```

I think that the correct solution could look like:

```java
            if (tracingEnabled) {
                PulsarTrace trace = new PulsarTrace.Builder() // Builder initializes properties with new HashMap
                        .withTopic(producer.getTopic())
                        .build();
                TracingUtils.traceOutgoing(instrumenter, message, trace);
                messageBuilder.properties(trace.getProperties()); // Reuse the same instance of the HashMap
            }
```

This happens only when PulsarOutgoingMessageMetadata is not used.

michalcukierman:
During the tests, I discovered that more changes are required. Pushed the second commit. In the other case it's also about re-using the same instance of map between the PulsarTrace and MessageBuilder. 

Please feel free to reject this PR and propose other solutions that would work.

PR #2520, opened 2024-03-09
#2519 Fix: PulsarOutgoingChannel tracing properties propagation

michalcukierman:


smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/PulsarOutgoingChannel.java
```diff
@@ -166,12 +165,11 @@ private TypedMessageBuilder<T> toMessageBuilder(Message<?> message, Producer<T>
...
-                Map<String, String> properties = new HashMap<>();
-                TracingUtils.traceOutgoing(instrumenter, message, new PulsarTrace.Builder()
-                        .withProperties(properties)
+                PulsarTrace trace = new PulsarTrace.Builder()
                         .withTopic(producer.getTopic())
-                        .build());
-                messageBuilder.properties(properties);
+                        .build();
+                TracingUtils.traceOutgoing(instrumenter, message, trace);
+                messageBuilder.properties(trace.getProperties());
```
michalcukierman, line 172: We need to reuse the same instance of a map.

ozangunalp:
@michalcukierman Thank you for this! We better check this in `io.smallrye.reactive.messaging.pulsar.tracing.TracingPropagationTest`.
I suggest we add an assertion in `testFromPulsarToAppToPulsar` to check for "traceparent" property in the received message properties.

ozangunalp:
@michalcukierman Here's my suggestion : https://github.com/smallrye/smallrye-reactive-messaging/commit/fa5099b063ac9a0b9202bf995a5a67011c17ebca

michalcukierman:
Thanks @ozangunalp, thank you for the test I should have add it at first. It looks much better now.
And sorry for the late update, we are going to launch the product next week, so I have a lot on my plate.

michalcukierman:
I've added one more test for messages with PulsarOutgoingMessageMetadata, to cover this change:
https://github.com/smallrye/smallrye-reactive-messaging/pull/2520/files#diff-b9a66ad40b9d795a40fff17754b68b2bf3da86869a99eebc2581100198500a6aR128-R136

ozangunalp (approved):

PR #2543, opened 2024-03-18
Upgrade to Mutiny 2.6.0 with JCTools

jponge:


jponge:
@ozangunalp @cescoffier fear the upgrade, anything can go wrong 🤣

ozangunalp:
@jponge well it is green

jponge:
> @jponge well it is green

I'm scared it passes all 4 tests

cescoffier:
It passes even the flaky tests...

cescoffier (approved):

cescoffier:
@ozangunalp this would need to be in a new minor, as it changes quite some stuff internally.

ozangunalp:
@cescoffier I've assigned it to 4.20.0 so the release after the next

jponge:
Do you have any idea of the timeframe? (to plan the Quarkus update)

cescoffier:
I think the timing is good, I do not see why we should delay it. The only consideration was that we need a way to release previous version to support the Quarkus LTS.

cescoffier (approved):

ozangunalp (approved):

jponge:
:part

PR #2547, opened 2024-03-20
Must be a QuarkusTest to use inject

bechhansen:
This test need to be annotated with @QuarkusTest for Injection to work right?


cescoffier:
I don't mind adding @QuarkusTest, but I would do something like:

```
@QuarkusTest // Or the Junit 5 extension that allows injection in tests
```

ozangunalp:
Yes it could be Arquillian for example

ozangunalp:
I've pushed a simple comment as I am against adding Quarkus dependency to the project.

PR #2580, opened 2024-04-12
Corrected message.release() in PulsarMessageAck.java

michalcukierman:
Message is already released here:
https://github.com/smallrye/smallrye-reactive-messaging/blob/a9719f018ad34f8140d1a0fd5b5f6537b5a1859a/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/ack/PulsarMessageAck.java#L40

### DIFF
--- a/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/ack/PulsarMessageAck.java
+++ b/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/ack/PulsarMessageAck.java
@@ -41,7 +41,6 @@ public class PulsarMessageAck implements PulsarAckHandler {
         return Uni.createFrom().completionStage(() -> {
             var txnMetadata = message.getMetadata(PulsarTransactionMetadata.class);
             if (txnMetadata.isPresent()) {
-                message.unwrap().release();
                 return consumer.acknowledgeAsync(messageId, txnMetadata.get().getTransaction());
             } else {
                 return consumer.acknowledgeAsync(messageId);


michalcukierman:
@ozangunalp What do you think about enabling the pooling by default?

What's more, we should probably release the messages on FailureHandlers which are not performing ACK.

https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarContinue.java#L42
https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarNack.java#L45

smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/ack/PulsarMessageAck.java
```diff
@@ -41,7 +41,6 @@ public Uni<Void> handle(PulsarIncomingMessage<?> message) {
         return Uni.createFrom().completionStage(() -> {
             var txnMetadata = message.getMetadata(PulsarTransactionMetadata.class);
             if (txnMetadata.isPresent()) {
```
michalcukierman, line 43: It's already released in line #40

ozangunalp:
> What's more, we should probably release the messages on FailureHandlers which are not performing ACK.

Ok for releasing the message on all failure handlers.

But I think we should keep the message pooling to the Pulsar producer default. I didn't hesitate to add `release` calls because when the message is not pooled it is a no-op.

ozangunalp (approved):

PR #2584, opened 2024-04-15
feature: Pulsar pool messages support in failure handlers

michalcukierman:
Added messages release in failure handler, What do you think @ozangunalp @marekczajkowski?

### DIFF
--- a/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
+++ b/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
@@ -42,6 +42,7 @@ public class PulsarIgnore implements PulsarFailureHandler {
     public Uni<Void> handle(PulsarIncomingMessage<?> message, Throwable reason, Metadata metadata) {
         log.messageFailureIgnored(channel, reason.getMessage());
         log.messageFailureFullCause(reason);
+        message.unwrap().release();
         return Uni.createFrom().completionStage(message.ack())
                 .emitOn(message::runOnMessageContext);
     }
--- a/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
+++ b/smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
@@ -42,7 +42,6 @@ public class PulsarIgnore implements PulsarFailureHandler {
     public Uni<Void> handle(PulsarIncomingMessage<?> message, Throwable reason, Metadata metadata) {
         log.messageFailureIgnored(channel, reason.getMessage());
         log.messageFailureFullCause(reason);
-        message.unwrap().release();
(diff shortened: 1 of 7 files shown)


smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
```diff
@@ -42,6 +42,7 @@ public PulsarIgnore(String channel) {
     public Uni<Void> handle(PulsarIncomingMessage<?> message, Throwable reason, Metadata metadata) {
         log.messageFailureIgnored(channel, reason.getMessage());
         log.messageFailureFullCause(reason);
+        message.unwrap().release();
```
ozangunalp, line 45: This would not work as the ack needs to access messageid.

ozangunalp (changes requested):
I'll add a change to run ack/nack tests with poolMessages to true.

smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/fault/PulsarIgnore.java
michalcukierman: @ozangunalp 
the release method:
```  
public void release() {
    if (this.poolMessage) {
      ReferenceCountUtil.safeRelease(this.payload);
      this.recycle();
    }

  }
```
works only on payload, not the messageId. MessageId is not a ByteBuf, but a standard Java class (not a direct memory).

michalcukierman: Ok your are right, the recycle method sets the messageId to null. I'll fix that.

ozangunalp: @michalcukierman I just pushed a change to the branch with some tests. The final change only calls the release on continue and failstop.

michalcukierman: Thank you @ozangunalp and sorry for incomplete PR, I believe it’s OK now. We’ll probably enable pooling, so if anything comes out, we will let you know.

kamilchociej: @ozangunalp It looks like the [message release](https://github.com/smallrye/smallrye-reactive-messaging/pull/2584/commits/02538ac0ce8cadec3ecb274911674d39b8afa8aa#diff-a284b5a2429cc7c1a84b5c55ef324b1d3dcaaceb944cd459a700949ebc3817ebR48) introduced by @michalcukierman is required to avoid a memory leak. While this change resolves the direct memory issue, it breaks DLQ functionality due to a [bug in Pulsar](https://github.com/apache/pulsar/issues/13269). I have created an [issue](https://github.com/smallrye/smallrye-reactive-messaging/issues/2698) to track this problem.

PR #2599, opened 2024-04-30
JMS: WIP: Initial commit for fault strategies

dankristensen:
WIP: Initial commit with 2 simple fault strategies for jms.

I need a little help, creating tests for these. I looked into how it is done for kafka, but it does not look like it can be done the same way here.

Please give me some feedback/hints for this.


dankristensen:
The build keeps failing here. It works fine when i build it locally. Does anybody have any clues?

ozangunalp:
I've only skimmed the change, it looks good. But I need to look into something with how messages are dispatched for JMS connector. I am looking at the CI issues, I hope I found the fix. After that I'll look into this PR.

dankristensen:
Finally i have had time to look further into this. I have a problem to get the test LocalPropagationAckTest.testChannelWithNackOnMessageContextDlq to work correctly. It is still returning 2,3,3,5,6 and should "only" return 2,3,5,6.

I had to pass the JmsConnector into the failure strategy. I am not happy about this, but did not find any other way to do this. Hopefully you guys have some pointers for me, so i can avoid this.

PR #2604, opened 2024-05-03
Set global junit timout to 5 minutes for rabbitmq

ozangunalp:


smallrye-reactive-messaging-rabbitmq/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1 @@
+junit.jupiter.execution.timeout.default=5 m
```
cescoffier, line 1: It is per test or for the whole test suite? 

If it's per test that's awesome!

cescoffier (dismissed):

smallrye-reactive-messaging-rabbitmq/src/test/resources/junit-platform.properties
ozangunalp: The doc says "Default timeout for all testable and lifecycle methods". 
it is interesting because rabbitmq container startup happens on beforeall lifecycle method.

Now that we know it is the problem, we can apply a startup timeout with a retry. 
I'll need to check out if downgrading to the 3.11 image really resolves the issue. 
And then understand what is the issue with the newest rabbitmq images.

cescoffier (approved):

cescoffier (approved):

cescoffier:
@ozangunalp I let you merge when you're ready. We might need to rebase a few PRs once in.

ozangunalp:
I'll rerun again it seems good. 
For the record I had to combine two changes : 
- I've rolled back on to 3.11 for the rabbitmq image.
- Removed await indefinitely from RabbitMQ channel termination.

Issue #2625, opened 2024-05-20
IncomingJmsMessage assumes message type it won't be empty


neazevedo:
When used to connect with (legacy activemq server) activemq-client v6.1.1, sending a JMSMessage by default type is "" (empty). 

<img width="638" alt="image" src="https://github.com/smallrye/smallrye-reactive-messaging/assets/16721777/832dedb4-1ad2-42fd-ad3d-7d314878ced0">

![image](https://github.com/smallrye/smallrye-reactive-messaging/assets/16721777/43abcb0a-389f-4e2b-a240-09a1fa9547bb)

After consuming a JMSMessage JMSType came empty and we try to load type "" from the classloader.  

Can we add extra isNotEmpty() validation?
Thanks

neazevedo:
A possible solution for this issue:

` IncomingJmsMessage(Message message, Executor executor, JsonMapping jsonMapping) {
        this.delegate = message;
        this.jsonMapping = jsonMapping;
        this.executor = executor;
        String cn = null;
        try {
            cn = message.getStringProperty("_classname");
            if (cn == null) {
                cn = message.getJMSType();
            }
        } catch (JMSException e) {
            // ignore it
        }
        try {
            this.clazz = cn != null && !cn.isEmpty() ? load(cn) : null;
        } catch (ClassNotFoundException e) {
            throw ex.illegalStateUnableToLoadClass(e);
        }

        this.jmsMetadata = new IncomingJmsMessageMetadata(message);
        this.metadata = Metadata.of(this.jmsMetadata);
    }`

neazevedo:
Possible solution: 

```java
 IncomingJmsMessage(Message message, Executor executor, JsonMapping jsonMapping) {
        this.delegate = message;
        this.jsonMapping = jsonMapping;
        this.executor = executor;
        String cn = null;
        try {
            cn = message.getStringProperty("_classname");
            if (cn == null) {
                cn = message.getJMSType();
            }
        } catch (JMSException e) {
            // ignore it
        }
        try {
            this.clazz = cn != null && !cn.isEmpty() ? load(cn) : null;
        } catch (ClassNotFoundException e) {
            throw ex.illegalStateUnableToLoadClass(e);
        }

        this.jmsMetadata = new IncomingJmsMessageMetadata(message);
        this.metadata = Metadata.of(this.jmsMetadata);
    }
```

cescoffier:
Looks good! Fancy a PR?

[...]

PR #2630, opened 2024-05-21
Fix empty JMSType for legacy connector

neazevedo:
This PR fix: https://github.com/smallrye/smallrye-reactive-messaging/issues/2625


ozangunalp:
@neazevedo Thanks for this!

ozangunalp (approved):

Issue #2663, opened 2024-06-21
Add configuration to outgoing MQTT connector to set retain property


diversit:
# Issue

With the MQTT Connector,
it is currently not possible to set the 'retain' property just by configuration for an outgoing MQTT channel.

Now, instead of just returning or emitting the payload for a message, an instance of `MqttMessage` must be returned to allow to set the `retain` property to `True`.

Example:
```
MqttMessage.of("a-topic", myValue, MqttQoS.AT_LEAST_ONCE, true))
```

The `topic` and `qos` settings __can__ (need to ?) be set in configuration, but retain property cannot.
It's unfortunate that, when the message must be retained, the `topic` and `qos` values now also must be set in code.

I assume that the MQTT Connector only uses those configuration values to construct a `MqttMessage` instance
and that by creating a `MqttMessage` in code, this take priority over the configuration values. 

# Proposal

Add a 'retain' property for outgoing MQTT channel to set the retain property of a outgoing MQTT message.

Example:
```
mp.messaging.outgoing.my-channel.connector=smallrye-mqtt
mp.messaging.outgoing.my-channel.retain=true
```

cescoffier:
Fancy a PR? It should be easy to add, as you need:

1. Add a connector attribute
2. Pass the configuration of the MQTT client

Of course, we can guide you.

diversit:
I made a PR: #2665 

I was already looking at it Friday. The changes did not seem that difficult indeed.
I made the necessary changes including a test case, but the test case is failing and I can't figure out why. It seems to go wrong in the MQTT client code. While debugging, the retain property is set when it's being serialised to a `byte[]` but when deserialising the retain values seems to have been lost.

Any idea how to fix it? Or is my test case wrong?

ozangunalp:
Closed by #2665

PR #2665, opened 2024-06-24
Add configuration 'retain' property for outgoing MQTT message

diversit:
Fixed issue #2663 

I did not find any documentation or example to update manually.
As far as I can see, the documentation gets updated automagically with the new connector attribute for outgoing messages.

Added a test to verify the send message contains the 'retain' property set to True, but cannot get this test to succeed.
It seems to go wrong somewhere in the Mqtt client code.
While debugging, I see the retain property is correctly set to true when the message is transformed into a `byte[]`.
But when reconstructing the message from `byte[]` the retain setting seems to have been lost and therefore the value is False instead of the expected True.

Any idea how to fix this? I'm clueless.


ozangunalp:
For the test, you need to start the consumer after the message has been produced. The best way to ensure that is to consume it once and check that retained flag is false, and then consume it again to check that retained flag is true.

cescoffier:
The code change looks good, but yes, the test might be a bit brittle.

ozangunalp:
@diversit I pushed a fix to the test you've added. Note that only the last message is retained with mqtt.

ozangunalp (approved):

PR #2677, opened 2024-07-03
Bump OTel instrumentation to 2.5.0

brunobat:
**Breaking changes.** 
For details see: https://github.com/quarkusio/quarkus/pull/41521

### DIFF
--- a/documentation/src/main/java/connectors/tracing/MyAttributesExtractor.java
+++ b/documentation/src/main/java/connectors/tracing/MyAttributesExtractor.java
@@ -1,9 +1,12 @@
 package connectors.tracing;
 
+import java.util.Collections;
+import java.util.List;
+
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.incubator.semconv.messaging.MessagingAttributesGetter;
 import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
-import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesGetter;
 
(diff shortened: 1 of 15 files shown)


brunobat:
@ozangunalp leaving this draft here because I was having problems with the pulsar image locally and also I get a test error that looks unrelated:
`SqsClientConfigTest.testRegionConfigMissing:22 
Expecting code to raise a throwable.
`

brunobat:
@ozangunalp I have no idea if these errors are relevant for the PR.

ozangunalp:
That was just the documentation module which provides a sample connector with tracing integration. I've pushed a fix.

brunobat:
@ozangunalp do you think this is good to go?

ozangunalp:
Is it ok if we release 4.22.0 without this one and include it in 4.23 ?

brunobat:
It depends when you want to have this merged on Quarkus... I'm not sure either.
When do you think 4.23 can go in?

ozangunalp:
Ideally, we can cut 4.23 just after with only the OTel update.

brunobat:
When do you think that can happen?

gsmet:
Ideally, I think we would need this upgrade merged in Quarkus by next Tuesday.

Which means we need the release to happen very very soon (release time + CI time).

ozangunalp:
Rebased this on top the 4.22.0 release. 

If I am not mistaken this means RM will be incompatible with OTel 1.x. 
I'll release 4.23.0 to include only this PR.

ozangunalp (approved):

brunobat:
> Rebased this on top the 4.22.0 release.
> 
> If I am not mistaken this means RM will be incompatible with OTel 1.x. I'll release 4.23.0 to include only this PR.

Yes, this is a Breaking Change for the OTel side. IT should be on the release notes.

PR #2692, opened 2024-07-16
Improve failure handling for Kafka transactions

ozangunalp:
- Abort the Kafka tx when the commit fails
- Cache ConsumerGroup metadata on the consumer in rebalance handler: Enables accessing the metadata without the need for polling thread
- Make sure EOP transactions are committed without changing to the producer thread.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
@@ -4,6 +4,8 @@ import java.util.Arrays;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.clients.consumer.ConsumerGroupMetadata;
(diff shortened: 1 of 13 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
```diff
@@ -14,6 +16,8 @@
 
 public class KafkaRecordHelper {
 
+    public static final String SMALLRYE_GENERATION_ID = "__smallrye_generationId";
```
cescoffier, line 19: Does it mean it's SmallRye specific?

cescoffier (approved):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordHelper.java
ozangunalp: Yes, to be clear, this should never make it to the user code. It is added on poller thread and removed after.

PR #2727, opened 2024-08-16
Added tracing using opentelemetry for jms like kafka does.

dankristensen:
JMS Connector did not have tracing capabilities like kafka did.

Here is a PR, that contains this functionality.

Please review and give me feedback on this

### DIFF
--- a/smallrye-reactive-messaging-jms/pom.xml
+++ b/smallrye-reactive-messaging-jms/pom.xml
@@ -107,6 +107,16 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-trace</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-testing</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>io.smallrye.reactive</groupId>
       <artifactId>smallrye-reactive-messaging-provider</artifactId>
       <version>${project.version}</version>
@@ -116,6 +126,20 @@
       <artifactId>smallrye-connector-attribute-processor</artifactId>
       <version>${project.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-instrumentation-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-instrumentation-api-incubator</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-otel</artifactId>
+      <version>999-SNAPSHOT</version>
+      <scope>compile</scope>
     </dependency>
 
   </dependencies>
--- a/smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/tracing/JmsAttributesExtractor.java
+++ b/smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/tracing/JmsAttributesExtractor.java
@@ -1,0 +1,114 @@
+package io.smallrye.reactive.messaging.jms.tracing;
+
+import static io.opentelemetry.semconv.SemanticAttributes.*;
+
+import java.util.Collections;
+import java.util.List;
+
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.incubator.semconv.messaging.MessagingAttributesGetter;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
+
+public class JmsAttributesExtractor implements AttributesExtractor<JmsTrace, Void> {
+    private final MessagingAttributesGetter<JmsTrace, Void> messagingAttributesGetter;
+
+    public JmsAttributesExtractor() {
+        this.messagingAttributesGetter = new JmsMessagingAttributesGetter();
+    }
+
+    @Override
+    public void onStart(final AttributesBuilder attributes, final Context parentContext, final JmsTrace jmsTrace) {
+        String groupId = jmsTrace.getGroupId();
+        String clientId = jmsTrace.getClientId();
+        if (groupId != null && clientId != null) {
+            String consumerId = groupId;
+            if (!clientId.isEmpty()) {
+                consumerId += " - " + clientId;
+            }
+            attributes.put(MESSAGING_CONSUMER_ID, consumerId);
+        }
+        if (groupId != null) {
+            attributes.put(MESSAGING_KAFKA_CONSUMER_GROUP, groupId);
+        }
+    }
+
+    @Override
+    public void onEnd(
+            final AttributesBuilder attributes,
+            final Context context,
+            final JmsTrace kafkaTrace,
+            final Void unused,
+            final Throwable error) {
+
+    }
+
+    public MessagingAttributesGetter<JmsTrace, Void> getMessagingAttributesGetter() {
+        return messagingAttributesGetter;
+    }
+
+    private static final class JmsMessagingAttributesGetter implements MessagingAttributesGetter<JmsTrace, Void> {
+        @Override
+        public String getSystem(final JmsTrace jmsTrace) {
+            return "jms";
+        }
+
+        @Override
+        public String getDestination(final JmsTrace jmsTrace) {
+            return jmsTrace.getQueue();
+        }
+
+        @Override
+        public boolean isTemporaryDestination(final JmsTrace jmsTrace) {
+            return false;
+        }
+
+        @Override
+        public String getConversationId(final JmsTrace jmsTrace) {
+            return null;
+        }
+
+        @Override
+        public String getMessageId(final JmsTrace jmsTrace, final Void unused) {
+            return null;
+        }
+
+        @Override
+        public List<String> getMessageHeader(JmsTrace jmsTrace, String name) {
+            return Collections.emptyList();
+        }
+
+        @Override
+        public String getDestinationTemplate(JmsTrace jmsTrace) {
+            return null;
+        }
+
+        @Override
+        public boolean isAnonymousDestination(JmsTrace jmsTrace) {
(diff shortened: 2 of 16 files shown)


smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/tracing/JmsAttributesExtractor.java
```diff
@@ -0,0 +1,114 @@
...
+        String groupId = jmsTrace.getGroupId();
+        String clientId = jmsTrace.getClientId();
+        if (groupId != null && clientId != null) {
+            String consumerId = groupId;
+            if (!clientId.isEmpty()) {
+                consumerId += " - " + clientId;
+            }
+            attributes.put(MESSAGING_CONSUMER_ID, consumerId);
+        }
+        if (groupId != null) {
```
ozangunalp, line 31: We can remove this one.

ozangunalp (changes requested):
Thanks for this! Looks good overall. I've noted some small changes that are required.
We can also rename variables `kafkaTrace` to `jmsTrace`.

There are some disabled tests. Do you need any help with those?

smallrye-reactive-messaging-jms/pom.xml
```diff
@@ -117,6 +127,20 @@
...
+    <dependency>
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-instrumentation-api-incubator</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-otel</artifactId>
+      <version>999-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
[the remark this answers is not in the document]
dankristensen: I have renamed all places where kafkaTrace is used to jmsTrace

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/tracing/JmsAttributesExtractor.java
dankristensen: I am not quite sure i understand this. Do you think we can remove groupId and clientId from the JmsTrace.Builder, so we only have queue and properties? And then cleanup?

smallrye-reactive-messaging-jms/pom.xml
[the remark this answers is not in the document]
ozangunalp: The dependency `io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-incubator` is already pulled by the `smallrye-reactive-messaging-otel`. Maybe you are adding it as a test dependency ?

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/tracing/JmsAttributesExtractor.java
ozangunalp: I meant to suggest removing `attributes.put(MESSAGING_KAFKA_CONSUMER_GROUP, groupId);` because that key is for kafka.

dankristensen: I have removed the use of MESSAGING_KAFKA_CONSUMER_GROUP. But i could use your help to get the tests in TestPropagationTest working.

smallrye-reactive-messaging-jms/pom.xml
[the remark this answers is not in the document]
ozangunalp: Could you take a look at the last change I pushed. I am looking at the tests. I think we don't need batch and regular tests for Jms.

ozangunalp:
@dankristensen I've fixed the tests and removed completely client id and group id from the JmsTrace, I didn't know how to fill them.

dankristensen:
@ozangunalp That is great. I have time to look at this tommorow, I will try to get it running locally, and see that i actually get these telemetries shown correctly in jaeger.

ozangunalp:
@dankristensen did you have any time to check this ?

dankristensen:
@ozangunalp Sorry for not getting back to you. I got serious sick, but is back to work again now. I will try to build one, so i can verify if it works correctly

dankristensen:
I have tried, and can see, that i MUST have this dependency available: smallrye-reactive-messaging-otel. This is because it is used in JmsOpenTelemetryInstrumenter. Is this as wanted? Apparently it looks like kafka has same issue. I can see from the documentation, that you must include this. But the error is a little confusing, because tracing is per default enabled, and therefore i get this error. `iled to start application: java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
        at io.quarkus.runtime.Application.start(Application.java:101)
        at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:119)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:71)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
        at io.quarkus.runner.GeneratedMain.main(Unknown Source)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:116)
        at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.NoClassDefFoundError: io/smallrye/reactive/messaging/tracing/TracingUtils
        at io.smallrye.reactive.messaging.jms.tracing.JmsOpenTelemetryInstrumenter.createForSource(JmsOpenTelemetryInstrumenter.java:29)
`

What do you think?

ozangunalp:
> I can confirm, that i get traces into jaeger now. But i am a little in doubt about to to pass span's across jms messages. Do you have any suggestions for this?

@dankristensen indeed for trace propagation, we needed to make sure that trace information (traceparent) was set on the outgoing message properties. It was missing. I've fixed it.

I think this is good to go. Thanks for your contribution!

dankristensen:
@ozangunalp i have just built a snapshot and tried it. It helped with the traceparent, but it is not ALL of them that is being propagated correctly. I now see a rest POST call, that is publishing a message, and then receiving the message again, in the same span. When recieving the message again, it is doing some operation, and then publishing a new message, which is not aggregated in the same span. Should that not be the case, with the change you made to traceparent, or do i need to adjust something in my app for this?

Furthermore i can see, that it would be good, if we could get the channel name as part of the tags we publish. I have looked a bit into this, but i cannot see an obvious place to handle this. Do you have any suggestions for this?

ozangunalp:
> i have just built a snapshot and tried it. It helped with the traceparent, but it is not ALL of them that is being propagated correctly. I now see a rest POST call, that is publishing a message, and then receiving the message again, in the same span. When recieving the message again, it is doing some operation, and then publishing a new message, which is not aggregated in the same span. Should that not be the case, with the change you made to traceparent, or do i need to adjust something in my app for this?

So the app does the following ? 
receive POST -> publish message -> JMS -> consume message -> publish message -> JMS

> Furthermore i can see, that it would be good, if we could get the channel name as part of the tags we publish. I have looked a bit into this, but i cannot see an obvious place to handle this. Do you have any suggestions for this?

I don't think we do this for other connectors supporting tracing, if you can come up with a proposition that would be awesome.

dankristensen:
That is correct @ozangunalp .

receive POST -> publish message -> JMS -> consume message -> publish message -> JMS

And the publish/consume is actually done multiple times in the same microservice, before leaving to another microservice. So i guess this should be in the same span, as the original receive POST. Do you not agree in this?

ozangunalp:
@dankristensen Can you push your test to a repository so that I can take a look?

dankristensen:
@ozangunalp Here is a reproducer: https://github.com/dankristensen/srmj-trace-reproducer.
Since my company is running behind a proxy, i have adopted build.gradle/settings.gradle and jaeger configuration to my best effort. Let me know if you have any issues

ozangunalp:
@dankristensen Thanks for the reproducer. 
Indeed, we need to make a little more effort to make the automatic tracing propagation work on Quarkus for the JMS connector. 

But it changes completely how messages are dispatched:
Most of the other connectors follow this pattern to dispatch each message on a [distinct message Vert.x context](https://smallrye.io/smallrye-reactive-messaging/latest/concepts/message-context/). JMS doesn't adhere to that and dispatch messages on the poller thread pool.

Note that changing that may break some assumptions from app developers, but context propagation and different execution modes (event-loop, blocking, VT). will work better on Quarkus.

dankristensen:
I can see, that you have pushed changes @ozangunalp. Do i need to try it again in a real world example?

ozangunalp:
@dankristensen if you do that'll be plus yes!

ozangunalp (approved):

ozangunalp:
@dankristensen I am in favor of getting this in. wdyt ?

dankristensen:
Give me an hour, i am currently testing in a real world scenario. Will get back to you @ozangunalp

dankristensen:
I see a problem, that i cannot fix. In JmsOpenTelemetryInstrumenter line 38, we use MessageOperation, that is from an incubator package. Is it possible, to avoid using that one in any way?

dankristensen:
I currently have a problem in my service, and do not get the multi propagation executed correctly. It could be a mistake on my end, since we have also done some migration to github of the service i am working on. I will get back to you tommorow, where i have found a solution for this. Hopefully that is ok @ozangunalp ?

ozangunalp:
Thanks for looking to this. I am thinking that this is already fine for this iteration. We can merge this and iron out the edges on the next.

dankristensen:
I can confirm, that the trace is now working across multiple publish / receive events. But i had to change something in my service, to get it working. In particular i could not use MutinyEmitter.sendAndAwait, but had to use MutinyEmitter.sendAndForget instead. I am using this in application.properties: quarkus.messaging.blocking.signatures.execution.mode=event-loop
I also had to add @Blocking to some of the incoming messages

dankristensen:
This is great @ozangunalp . Thanks alot for your cooperation with this PR. That is fantastic. Looking forward to the release.

PR #2729, opened 2024-08-19
introduce retry parameters for publish-confirm

### LINKED ISSUE #2726: RabbitMQ should allow  configure explicitly Retry Failure for nack when using Publish Confirm
**Use case :**
Sending a message with rabbit mq ( with publish-confirm ) on a queue configured with `max-lenght` and `x-overflow : reject-publish'

**Expected**
As the queue is full, the message  is rejected and the error should get raise directly or quickly.

**Current Behaviour**
As the nack is considered as Failure,  the Retry on Failure came into action :

https://github.com/smallrye/smallrye-reactive-messaging/blob/c176b3ec6f4a65c456a9d17d4215316b60d37eb5/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/RabbitMQMessageSender.java#L285

But the Retry attemps / Retry interval are reusing the "ReconnectAttempts" and  "ReconnectInterval" :

https://github.com/smallrye/smallrye-reactive-messaging/blob/c176b3ec6f4a65c456a9d17d4215316b60d37eb5/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/RabbitMQMessageSender.java#L264-L265

The default configuration it quite high ( 100 tentative, delay 10 sec ==> 1000 seconds in total).
That fine to validate the connection itself, but a bit long for the nack check.

It think it should be great to have dedicated configuration options to be able to configure the retry on publish explicitly to shorter time.

**Sample project**
https://github.com/gcuisinier/quarkus-rabbitmq-tests/tree/rabbitOverflowNotNacked

gcuisinier:
Avoid reusing reconnect-attemps and reconnect interval to define explicits new parameters for the publish mode

Fix #2726


cescoffier (approved):

cescoffier:
thanks!

gcuisinier:
Happy to help :)

Issue #2714, opened 2024-08-04
KafkaRequestReply fails with  ` java.lang.IllegalStateException` if the partition assignment takes longer than `replyTimeout` at startup


Malandril:
# Context

When using quarkus 3.13.0 with a `KafkaRequestReply` at startup if the `KafkaRequestReply`'s `waitForAssignments` takes longer than the reply topic timeout, the `KafkaRequestReply` will always throw
```java
   java.lang.IllegalStateException: SRMSG00028: The subscription to question-out has been cancelled
        at io.smallrye.reactive.messaging.providers.extension.AbstractEmitter.verify(AbstractEmitter.java:196)
        at io.smallrye.reactive.messaging.providers.extension.AbstractEmitter.emit(AbstractEmitter.java:161)
        at io.smallrye.reactive.messaging.providers.extension.MutinyEmitterImpl.lambda$sendMessage$4(MutinyEmitterImpl.java:72)
        at io.smallrye.context.impl.wrappers.SlowContextualConsumer.accept(SlowContextualConsumer.java:21)
        at io.smallrye.mutiny.operators.uni.builders.UniCreateWithEmitter.subscribe(UniCreateWithEmitter.java:22)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniEmitOn.subscribe(UniEmitOn.java:22)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniOnItemConsume.subscribe(UniOnItemConsume.java:30)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniOnItemTransformToUni.subscribe(UniOnItemTransformToUni.java:25)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniOnItemOrFailureConsume.subscribe(UniOnItemOrFailureConsume.java:24)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniOnItemTransform.subscribe(UniOnItemTransform.java:22)
[...]

PR #2748, opened 2024-09-08
Add initial timeout property for `KafkaRequestReply` when using latest offset

Malandril:
This PR alleviates some a part of the issue of #2714 notably the fact that at start the `KafkaRequestReply` uses the same timeout to wait for assignments and to await replies when using the `latest` offset.

I didn't have any idea just named the property `reply.initial-assignment-timeout` you can propose a more suitable one if you think of one.


Malandril:
> @Malandril Thank you for this clean PR!
> 
> I'd like to push a change to disable the wait when `reply.initial-assignment-timeout` is `-1`. Would it be ok ?
> 
> Here is the proposition : [4940198](https://github.com/smallrye/smallrye-reactive-messaging/commit/494019874f69636e70aa18ae99c0a750bc86cd52)

Yeah it would be great ! Thanks!

ozangunalp (approved):

cescoffier (approved):

PR #2754, opened 2024-09-13
Added access to individual messages from incoming batch metadata for Kafka and Pulsar

ozangunalp:


cescoffier (approved):
Should we add something in the doc?

cescoffier (commented):
I'm wondering if we need to introduce the concept of "bag of messages", trying to generalize what's happening in Pulsar and Kafka batches. Should we define such an API and more particularly the semantic behind it (do not ack individual messages...)

ozangunalp:
Ack/nack semantics are particularly different in Kafka and Pulsar for batch messages (Pulsar keeps track of ack'ed messages in the batch)

The change brought here involves batch metadata, not the message types themselves.

cescoffier (approved):

PR #2757, opened 2024-09-16
Fix #2756 KafkaCompanion does not configure Serializer/Deserializer

### LINKED ISSUE #2756: KafkaCompanion does not configure Serializer/Deserializer
The `KafkaCompanion` does not properly configure a registered `Serde`, `Serializer` and/or `Deserialiser` which causes issues with certain implementations.

E.g. When using Confluents `io.confluent.kafka.streams.serdes.avro.SpecificAvroSerde` (and it's serializer/deserializer) not properly calling `configure` causes a test to fail with:
```
java.lang.AssertionError: Expected a completion event but got a failure: org.apache.kafka.common.errors.InvalidConfigurationException: You must configure() before serialize() or use serializer constructor with SchemaRegistryClient
``` 

This behaviour can be verified by adding a test case to `SerdesTest`:
```java
    @Test
    void testRegisteredSerdeShouldBeConfigured() {
        // Custom Serde which Serialize/Deserializer throw exception if not configured
        companion.registerSerde(Person.class, new Serde<>() {

            // Custom Deserializer which throws an exception if not configured
            private PersonDeserializer personDeserializer = new PersonDeserializer() {
                private boolean isConfigured = false;

                @Override
                public void configure(Map<String, ?> configs, boolean isKey) {
                    isConfigured = true;
                }

                @Override
                public Person deserialize(String s, byte[] bytes) {
                    if (isConfigured) {
                        return super.deserialize(s, bytes);
                    } else {
                        throw new SerializationException("Deserializer not configured");
                    }
                }
            };

            // Custom Serializer which throws an exception if not configured
            private PersonSerializer personSerializer = new PersonSerializer() {
                private boolean isConfigured = false;

                @Override
                public void configure(Map<String, ?> configs, boolean isKey) {
                    isConfigured = true;
                }

                @Override
                public byte[] serialize(String s, Person person) {
                    if (isConfigured) {
                        return super.serialize(s, person);
                    } else {
                        throw new SerializationException("Serializer not configured");
                    }
                }
            };

            @Override
            public Serializer<Person> serializer() {
                return personSerializer;
            }

            @Override
            public Deserializer<Person> deserializer() {
                return personDeserializer;
            }

            @Override
            public void configure(Map<String, ?> configs, boolean isKey) {
                // Configure the serializer and deserializer
                personSerializer.configure(configs, isKey);
                personDeserializer.configure(configs, isKey);
            }
        });

        companion.produce(Person.class).fromRecords(
                new ProducerRecord<>(topic, new Person("1", 30)),
                new ProducerRecord<>(topic, new Person("2", 25)),
                new ProducerRecord<>(topic, new Person("3", 18))).awaitCompletion();

        ConsumerBuilder<String, Person> consumer = companion.consumeWithDeserializers(PersonDeserializer.class.getName());
        ConsumerTask<String, Person> task = consumer.fromTopics(topic, 3).awaitCompletion();
        assertThat(task.getRecords()).hasSize(3);
    }
```
which will currently fail with
```
java.lang.AssertionError: Expected a completion event but got a failure: org.apache.kafka.common.errors.SerializationException: Serializer not configured
```

`KafkaCompanion` should be updated to call `configure` on the `Serde` so the Serializer and Deserialiser properly get configured __before__ a message is published or consumed.

Due to this issue, other serialisers like the `StringSerializer`, `UUIDSerializer`, `ListSerializer`, etc could also give unexpected results in tests.

Provided a fix in PR #2757

diversit:
Updated SerdesTest and added test case to verify KafkaCompanion does not configure the (de)serializers.

Updated ConsumerBuilder and ProducerBuilder to fix the issue.

### DIFF
--- a/smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ConsumerBuilder.java
+++ b/smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ConsumerBuilder.java
@@ -143,6 +143,11 @@ public class ConsumerBuilder<K, V> implements ConsumerRebalanceListener, Closeab
             Deserializer<K> keyDeser, Deserializer<V> valueDeser) {
         this.props = props;
         this.kafkaApiTimeout = kafkaApiTimeout;
+
+        // configure deserializers
+        keyDeser.configure(props, true);
+        valueDeser.configure(props, false);
+
         this.consumerCreator = p -> new KafkaConsumer<>(p, keyDeser, valueDeser);
     }
 
--- a/smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ProducerBuilder.java
+++ b/smallrye-reactive-messaging-kafka-test-companion/src/main/java/io/smallrye/reactive/messaging/kafka/companion/ProducerBuilder.java
@@ -131,6 +131,11 @@ public class ProducerBuilder<K, V> implements Closeable {
             Serializer<K> keySerializer, Serializer<V> valueSerializer) {
         this.props = props;
         this.kafkaApiTimeout = kafkaApiTimeout;
+
+        // configure serializers
+        keySerializer.configure(props, true);
+        valueSerializer.configure(props, false);
+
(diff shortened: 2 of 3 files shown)


diversit:
Not needed anymore as explained in #2756

PR #2772, opened 2024-09-27
Tries to inject Vertx instance from CDI using a qualifier

### LINKED ISSUE #2771: [Enhancement] Tries to inject Vertx instance from CDI using a qualifier
Currently the Vertx instance used in the reactive messaging comes from either CDI `Instance<Vertx>` or a new instance via `Vertx.vertx()`.

In some case, the Vertx instance is exposed with a CDI qualifier,  it would be good if we can only inject the instance with the qualifier.

More information can be found at the Zulip discussion: https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/vertx.20subsystem.20integration

gaol:
Fixes #2771

This PR tries to add 2 config items:
* `mp.messaging.connector.vertx.cdi.qualifier`, which is used to tries to  inject the Vertx instance with the specified qualifier if configured.
* `mp.messaging.connector.vertx.cdi.qualifier.required`, boolean, which is used to indicate if the qualifier is required in case it comes from CDI, otherwise it will create a new Vertx instance like what current does.

`ExecutionHolder` is used by both WildFly and Quarkus. And for WildFly, there is a coming vertx subsystem to be able to configure the VertxOptions following the WildFly management model style, which will expose a CDI bean with a qualifier so that other subsystems can select it by themselves. The 2nd config item can be included in the [integration code of WildFly](https://github.com/wildfly/wildfly/blob/34.0.0.Beta1/microprofile/reactive-messaging-smallrye/config/src/main/java/org/wildfly/microprofile/reactive/messaging/config/ReactiveMessagingConfigSource.java#L29) so that applications deployed to WildFly will inject only the instance with a qualifier, and leave the use case in Quarkus not changed at all.

### DIFF
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
@@ -7,9 +7,14 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.BeforeDestroyed;
 import jakarta.enterprise.event.Observes;
 import jakarta.enterprise.event.Reception;
+import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
 
+import org.eclipse.microprofile.config.Config;
+
+import io.smallrye.common.annotation.Identifier;
 import io.vertx.mutiny.core.Vertx;
 
 /**
@@ -18,6 +23,8 @@ import io.vertx.mutiny.core.Vertx;
  */
 @ApplicationScoped
 public class ExecutionHolder {
+
+    private static final String REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER = "mp.messaging.connector.vertx.cdi.identifier";
 
     private boolean internalVertxInstance = false;
     final Vertx vertx;
@@ -40,17 +47,34 @@ public class ExecutionHolder {
     }
 
     @Inject
-    public ExecutionHolder(Instance<Vertx> instanceOfVertx) {
-        if (instanceOfVertx == null || instanceOfVertx.isUnsatisfied()) {
+    public ExecutionHolder(@Any Instance<Vertx> instanceOfVertx, Instance<Config> config) {
+        String cdiQualifier = null;
+        if (config != null && !config.isUnsatisfied()) {
+            final Config theConfig = config.get();
+            cdiQualifier = theConfig.getConfigValue(REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER).getValue();
+        }
+        final Instance<Vertx> vertxInstance;
+        if (cdiQualifier != null && !cdiQualifier.isEmpty()) {
+            log.vertxFromCDIQualifier(cdiQualifier);
+            vertxInstance = instanceOfVertx.select(Identifier.Literal.of(cdiQualifier));
+        } else {
+            vertxInstance = instanceOfVertx.select(Default.Literal.INSTANCE);
+        }
+        if (vertxInstance == null || vertxInstance.isUnsatisfied()) {
             internalVertxInstance = true;
             this.vertx = Vertx.vertx();
             log.vertXInstanceCreated();
         } else {
-            this.vertx = instanceOfVertx.get();
+            this.vertx = vertxInstance.get();
         }
     }
 
     public Vertx vertx() {
         return vertx;
+    }
+
+    // this is used in the test
+    boolean isInternalVertxInstance() {
+        return internalVertxInstance;
     }
 }
--- a/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/i18n/ProviderLogging.java
+++ b/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/i18n/ProviderLogging.java
@@ -147,4 +147,8 @@ public interface ProviderLogging extends BasicLogger {
     @LogMessage(level = Logger.Level.WARN)
     @Message(id = 243, value = "Processing method '%s' annotated with @Acknowledgement(POST_PROCESSING), but may not be compatible with post-processing acknowledgement management. You may experience duplicate (negative-)acknowledgement of messages.")
     void postProcessingNotFullySupported(String methodAsString);
+
+    @LogMessage(level = Logger.Level.DEBUG)
+    @Message(id = 244, value = "Trying to get Vertx instance using Identifier qualifier: %s.")
+    void vertxFromCDIQualifier(String cdiQualifier);
 }
--- a/smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/WeldTestBaseWithoutTails.java
+++ b/smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/WeldTestBaseWithoutTails.java
@@ -4,10 +4,12 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
 
 import jakarta.enterprise.inject.se.SeContainer;
 import jakarta.enterprise.inject.se.SeContainerInitializer;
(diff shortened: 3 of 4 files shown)


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
```diff
@@ -40,13 +46,26 @@ public ExecutionHolder(Vertx vertx) {
...
+        if (config != null && !config.isUnsatisfied()) {
+            final Config theConfig = config.get();
+            cdiQualifier = theConfig.getConfigValue(REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER).getValue();
+            qualifierRequired = theConfig.getOptionalValue(REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER_REQUIRED, Boolean.class)
+                    .orElse(Boolean.FALSE);
+        }
+        Instance<Vertx> vertxInstance = instanceOfVertx;
+        if (cdiQualifier != null && !cdiQualifier.isEmpty()) {
+            log.vertxFromCDIQualifier(cdiQualifier);
+            vertxInstance = instanceOfVertx.select(NamedLiteral.of(cdiQualifier));
```
cescoffier, line 61: No Named, Identified if you want, but named has some hidden semantic you do not want here.

cescoffier (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
gaol: IIUC, that means we need to add `io.quarkus.arc:arc` as the dependency to the subsystems in WildFly, including smallrye-reactive-messaging, smallrye-opentelemetry and the coming vertx subsystem for WildFly, because [the same qualifier](https://github.com/quarkusio/quarkus/blob/3.15.1/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Identified.java) needs to be used in both exposing side and consuming side IMO. I am not sure if that is what we want.

OTOH, this will be primarily used in WildFly which has a vertx instance exposed with a cdi qualifier, so probably it won't get triggered in Quarkus.

Ladicek: Clement made a mistake, he meant `@Identifier`, which is present in SmallRye Common. ArC's `@Identified` is private and should not be used outside of ArC.

cescoffier: Ah yes, sorry.

ozangunalp: Yes, let's go for the `@io.smallrye.common.annotation.Identifier`.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
```diff
@@ -19,6 +22,9 @@
 @ApplicationScoped
 public class ExecutionHolder {
 
+    private static final String REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER = "mp.messaging.connector.vertx.cdi.qualifier";
+    private static final String REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER_REQUIRED = "mp.messaging.connector.vertx.cdi.qualifier.required";
```
ozangunalp, line 26: Do we really need a second property ? 
Can't we just look for "mp.messaging.connector.vertx.identifier" and if it is defined look up the instance with it?

ozangunalp (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
cescoffier: +1, just the identifier value should be fine.

gaol: Maybe not using Identifier, I will test to confirm that. The case it tries to cover is when there is a vertx instance exposed with a cdi qualifier(`Named("vertx")`) but the `Instance<Vertx> ist` in the ExecutionHolder, the `ist.isUnsatisfied()` is `false`, which forces sr rm to inject the instance.

BTW, I will be on holidays this week, will report back next week, thanks :)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
gaol: Done

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
gaol: Done by using only the first option and the `@io.smallrye.common.annotation.Identifier` qualifier only.

The vertx bean from the vertx subsystem is exposed with 2 qualifiers: `@Any` and `@Identifier`, so to select the instance only when the cdi qualifier is specified, I split the `@Default` out in the `else` check block.

gaol:
Thank you all for the reviews, I updated the PR, would you please review again ? thank you :)

gaol:
@cescoffier @ozangunalp would you please review again ? thanks :)

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
```diff
@@ -19,6 +24,8 @@
 @ApplicationScoped
 public class ExecutionHolder {
 
+    private static final String REACTIVE_MESSAGING_VERTX_CDI_QUALIFIER = "mp.messaging.connector.vertx.cdi.qualifier";
```
ozangunalp, line 27: Shouldn't we change the name of this property something like `mp.messaging.connector.vertx.cdi.identifier` ?

ozangunalp (changes requested):
A couple of tests in the provider module would be great!

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/ExecutionHolder.java
gaol: done

gaol:
@ozangunalp thank you for the reviews, I updated to add a test in the provider module, please let me know if there is anything needs to be fixed or improved. :)

gaol:
@cescoffier @ozangunalp would you please review again ? thanks

ozangunalp (approved):

ozangunalp:
@gaol looks good to me, I'll let @cescoffier to have another look.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/i18n/ProviderLogging.java
```diff
@@ -147,4 +147,8 @@ public interface ProviderLogging extends BasicLogger {
     @LogMessage(level = Logger.Level.WARN)
     @Message(id = 243, value = "Processing method '%s' annotated with @Acknowledgement(POST_PROCESSING), but may not be compatible with post-processing acknowledgement management. You may experience duplicate (negative-)acknowledgement of messages.")
     void postProcessingNotFullySupported(String methodAsString);
+
+    @LogMessage(level = Logger.Level.INFO)
```
cescoffier, line 151: Should be debug - not very informative for regular users

cescoffier (changes requested):
LGTM

I just had one minor comment about the log level.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/i18n/ProviderLogging.java
gaol: Done

gaol:
Thanks @ozangunalp and @cescoffier , updated again to change log level to DEBUG. :)

cescoffier (approved):
Thanks

PR #2825, opened 2024-11-19
Rabbitmq close consumer on stream termination

ozangunalp:
revert the stopAndForget client on termination.

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
@@ -261,7 +261,7 @@ public class RabbitMQConnector implements InboundConnector, OutboundConnector, H
             outgoing.terminate();
         }
 
-        clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndForget());
+        clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndAwait());
         clients.clear();
     }
 
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
@@ -239,21 +239,19 @@ public class IncomingRabbitMQChannel {
             RabbitMQAckHandler onAck) {
 
         final String queueName = RabbitMQClientHelper.getQueueName(ic);
-        final boolean isTracingEnabled = ic.getTracingEnabled();
         final String contentTypeOverride = ic.getContentTypeOverride().orElse(null);
         log.receiverListeningAddress(queueName);
 
-        if (isTracingEnabled) {
-            return receiver.toMulti()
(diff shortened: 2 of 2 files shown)


smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQConnector.java
```diff
@@ -261,7 +261,7 @@ public void terminate(
             outgoing.terminate();
         }
 
-        clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndForget());
+        clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndAwait());
```
cescoffier, line 264: Are you sure we are never called on the event loop? 

Maybe we need to be more defensive.

ozangunalp: For sure CDI events are called on worker threads. 

We used to await before actually. We had a CI issue which blocked on await. But I think it was because we didn't close the consumers.

bcluap:
When moving from Quarkus 3.17.X to 3.18.X, this change is incorporated and seems to prevent Quarkus from shutting down. Last log shows:
Stopping rabbitmq client
Main thread is stuck here:
`"main" #1 [137610] prio=5 os_prio=0 cpu=10538.03ms elapsed=144.24s tid=0x000078283402aac0 nid=137610 waiting on condition  [0x00007828399fc000]
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@23/Native Method)
        - parking to wait for  <0x00000000c75e0970> (a java.util.concurrent.CountDownLatch$Sync)
        at java.util.concurrent.locks.LockSupport.park(java.base@23/LockSupport.java:221)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@23/AbstractQueuedSynchronizer.java:754)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(java.base@23/AbstractQueuedSynchronizer.java:1099)
        at java.util.concurrent.CountDownLatch.await(java.base@23/CountDownLatch.java:230)
        at io.smallrye.mutiny.operators.uni.UniBlockingAwait.await(UniBlockingAwait.java:67)
        at io.smallrye.mutiny.groups.UniAwait.atMost(UniAwait.java:65)
        at io.smallrye.mutiny.groups.UniAwait.indefinitely(UniAwait.java:46)
        at io.vertx.mutiny.rabbitmq.RabbitMQClient.stopAndAwait(RabbitMQClient.java:1251)
   `
If i manually change to use 4.25.0 instead of 4.26.0, keeping Quarkus version at 3.18.1 then the problem goes away

bcluap:
The specific change is this one:
`
clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndForget());
-> clients.forEach((channel, rabbitMQClient) -> rabbitMQClient.stopAndAwait());
`
It looks like the vertx rabbit clients stop method does not pass a close timeout so it defaults to infinity. Perhaps with vertx having been shut down already, the close() just waits indefinitely 
See:
https://github.com/vert-x3/vertx-rabbitmq-client/blob/0128c6a628c403183ec18064a2704298e348f57f/src/main/java/io/vertx/rabbitmq/impl/RabbitMQClientImpl.java#L639

and 

https://github.com/rabbitmq/rabbitmq-java-client/blob/64fdf2d62b29dae5dfedb9230f4df5695a8a5141/src/main/java/com/rabbitmq/client/impl/AMQConnection.java#L1024

PR #2844, opened 2024-12-04
Disable context propagation for blocking message dispatch

ozangunalp:
Make sure to dispatch messages on the named worker thread pool with the message context


cescoffier (approved):
This deserves a small note in the doc, no?

PR #2852, opened 2024-12-12
New release process

ozangunalp:


cescoffier (approved):
LGTM - @jponge can you have a quick look?

ozangunalp:
Btw it is normal that the review release is failing, it doesn't have permission to add review to a PR coming from a fork.

cescoffier:
Yes, you need to push the branch to the main repo and not your fork.

jponge:
I've moved the Mutiny workflow to be local-centric with a `justfile`, and only have actions to push tags to Maven Central, and rebuild the website. I was having too many issues in running the whole workflow in CI, especially with JReleaser not getting the correct commit range.

I have had 0 issue on Mutiny Zero / Vert.x bindings with the standard SmallRye release process, so your mileage varies :-)

jponge:
(I was too quick to send)

The standard workflow is fine, but if it doesn't work and you don't want to spend hours debugging GItHub Actions failures, local steps work quite well.

Issue #2635, opened 2024-05-22
AWS SQS NACK doing nothing?


spc16670:
When nacking SQS message I would expect it to the immediately returned back to the broker i.e. NACKing should set the VisibilityTimeout to 0. Current implementation does not seem to do anything.

https://github.com/smallrye/smallrye-reactive-messaging/blob/4.21.0/smallrye-reactive-messaging-aws-sqs/src/main/java/io/smallrye/reactive/messaging/aws/sqs/SqsMessage.java#L120

Also is there a way to manually extend the message's VisibilityTimeout in case it is about to expire when the processing takes longer than expected?

prakashelango:
@ozangunalp am kind of thinking to contribute on this. Can you please give me some directions

ozangunalp:
@prakashelango that's great news! 

You can do like the ack handler already existing on the `SqsMessage`, to add a failure handler. A first implementation for that failure handler would do nothing (current strategy) and a second implementation would do a call using the consumer client to send a request to set message visibility to 0 using the message id. 

The  connector channel can receive an attribute to choose between two strategies. 

There are some other examples on other connectors. You can also check those.

adampoplawski:
@prakashelango  are you planning to contribute? If no maybe I can do it after we merge first one: https://github.com/quarkiverse/quarkus-amazon-services/issues/1414

prakashelango:
@adampoplawski am in midst, need to test it out.

prakashelango:
@adampoplawski  @ozangunalp Have raised a [pr](https://github.com/smallrye/smallrye-reactive-messaging/pull/2882), can you guys review this?

adampoplawski:
@ozangunalp will you be looking at PR? I can also contribute and implement changes after review.

PR #2882, opened 2024-12-30
Aws sqs nack

prakashelango:
When nacking SQS message I would expect it to the immediately returned back to the broker i.e. NACKing should set the VisibilityTimeout to 0. Current implementation does not seem to do anything.

refer below issue

https://github.com/smallrye/smallrye-reactive-messaging/issues/2635


ozangunalp:
I'll need to close this one in favour of #2903

PR #2915, opened 2025-02-04
Added dynamic priority support for jms similar to amqp

User-123321123:
Added priority to OutgoingJmsMessageMetadata, which allows setting the priority before sending the message. The ampq connector implementation provides a similar feature.

### DIFF
--- a/smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSink.java
+++ b/smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSink.java
@@ -137,6 +137,7 @@ class JmsSink {
                 Destination dest = metadata.getDestination();
                 int deliveryMode = metadata.getDeliveryMode();
                 String type = metadata.getType();
+                final Integer priority = metadata.getPriority();
                 JmsProperties properties = metadata.getProperties();
                 if (correlationId != null) {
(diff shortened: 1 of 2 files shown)


smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSink.java
```diff
@@ -156,6 +157,9 @@ private Uni<? extends Message<?>> send(JmsResourceHolder<JMSProducer> resourceHo
                 if (type != null) {
                     outgoing.setJMSType(type);
                 }
+                if (priority != null) {
```
ozangunalp, line 160: Shouldn't this set the priority on the outgoing message?

ozangunalp (changes requested):

smallrye-reactive-messaging-jms/src/main/java/io/smallrye/reactive/messaging/jms/JmsSink.java
User-123321123: That was my first thought too, but using **outgoing.setJMSPriority(priority);** is not working. Could this be a mapping problem with the **jakarta.jms.Message outgoing**?

Setting it directly on the sender did work for me (**resourceHolder.getClient().setPriority(priority)**)

cescoffier:
@ozangunalp @User-123321123 what's the status on this one?

PR #2962, opened 2025-03-09
feat: Add aws-sns support

holomekc:


holomekc:
Ok locally it looks still fine:
image
I will check if the test is flaky.

ozangunalp:
I had the same failure running locally:

```
SnsOutgoingTest.should_use_provided_client_to_send_string_messages:39->SnsTestBase.receiveAndDeleteMessages:120->SnsTestBase.receiveAndDeleteMessages:125->SnsTestBase.receiveAndDeleteMessages:165 » Runtime java.util.concurrent.ExecutionException: software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: Acquire operation took longer than the configured maximum time. This indicates that a request cannot get a connection from the pool within the specified maximum time. This can be due to high request rate.
2025-03-10T12:16:29.3181111Z Consider taking any of the following actions to mitigate the issue: increase max connections, increase acquire timeout, or slowing the request rate.
2025-03-10T12:16:29.3184850Z Increasing the max connections can increase client throughput (unless the network interface is already fully utilized), but can eventually start to hit operation system limitations on the number of file descriptors used by the process. If you already are fully utilizing your network interface or cannot further increase your connection count, increasing the acquire timeout gives extra time for requests to acquire a connection before timing out. If the connections doesn't free up, the subsequent requests will still timeout.
2025-03-10T12:16:29.3189171Z If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large traffic bursts cannot overload the client, being more efficient with the number of times you need to call AWS, or by increasing the number of hosts sending requests. (SDK Attempt Count: 1)

```

holomekc:
Hmm I will try to figure out why this happens.

Besides this though, I think the pool should be replaced anyway with:
```java
SnsAsyncClient.builder().asyncConfiguration(b -> b.advancedOption(SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR, Vertx.vertx().nettyEventLoopGroup()))
```

But I lack some vertx knowledge in this repo to be honest.

holomekc:
Hi @ozangunalp ,
I found it. This app from the test creates the issue in combination with other tests:
```java
    @ApplicationScoped
    public static class ProducerApp {

        private final AtomicInteger counter = new AtomicInteger(0);

        @Outgoing("data")
        public Uni<String> send() {
            if (counter.getAndIncrement() > 9) {
                return Uni.createFrom().nullItem();
            }
            return Uni.createFrom().item(String.format("hello-%d", counter.get()));
        }
    }
```

The issue seems to be the Uni usage. It looks like this floods(???) the connection pool of the AWS SDK. I copied the tests from the SQS connector. The error occurs in the test itself when it tries to request the data from the testing sqs queue. See error message:
```
2025-03-10T12:16:15.4441531Z Consider taking any of the following actions to mitigate the issue: increase max connections, increase acquire timeout, or slowing the request rate.
2025-03-10T12:16:15.4448973Z Increasing the max connections can increase client throughput (unless the network interface is already fully utilized), but can eventually start to hit operation system limitations on the number of file descriptors used by the process. If you already are fully utilizing your network interface or cannot further increase your connection count, increasing the acquire timeout gives extra time for requests to acquire a connection before timing out. If the connections doesn't free up, the subsequent requests will still timeout.
2025-03-10T12:16:15.4455974Z If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large traffic bursts cannot overload the client, being more efficient with the number of times you need to call AWS, or by increasing the number of hosts sending requests. (SDK Attempt Count: 1)
2025-03-10T12:16:15.4458871Z 	at io.smallrye.reactive.messaging.aws.sns.SnsTestBase.receiveAndDeleteMessages(SnsTestBase.java:165)
2025-03-10T12:16:15.4460465Z 	at io.smallrye.reactive.messaging.aws.sns.SnsTestBase.receiveAndDeleteMessages(SnsTestBase.java:125)
```

I adjusted the test similar to the other once and then it works (at least on my machine).
But this feels wrong. Still testing further.

ozangunalp:
You can change it to Multi and it should be fine. I think it still makes a send request even though the weld container is closed. 
It may be a problem with the connector closing sequence.

holomekc:
Ok then I have that.

cescoffier:
Just to be clear: Vertx().vertx() creates a new instance of Vertx. You should use the existing one if there is one.

holomekc:
Hi @cescoffier,
I saw that while testing with that. But to not make it too complicated I did not used it yet completely. I would create a dedicated pr to add this for SQS and SNS, but then via the ExecutionHolder.

cescoffier:
Vertx 5 is not going to expose Netty objects from its public API. We would have to workaround it. I know how to do it in Quarkus, but not sure here.

holomekc:
I guess how Vertx describes it in the migration guide. But it looks a little bit strange, because at the moment it would still show the deprecation. I guess this would disappear with Vertx 5 though.

cescoffier:
Yes, the deprecation will disappear with Vert.x 5.

holomekc:
@ozangunalp, @cescoffier,

is something else needed? Or do you need more time to check the PR?

ozangunalp:
Overall, it looks good. I have a question about how to get the SnsAsyncClient. In the target environment, there needs to be an implementation of the `SnsClientProvider` bean, right? 

Also, it would be cool if we can get some documentation. I can be in the form of https://smallrye.io/smallrye-reactive-messaging/latest/sqs/sqs/

holomekc:
Well it is optional. If none is defined the client is created by the configuration. If one is defined it uses the defined one. I am also not 100% sure about this approach, but this would allow me to use a client created by quarkiverse-amazon-services (qas). At least this was the idea. But maybe something is still missing there. Because qas allows you to define multiple clients. This might be necessary when you need to configure different clients with different credential providers. Ao maybe this approach is also not flexible enough. In the sqs approach I noticed it searches for one client bean and uses this one if the config is incomplete (e.g. by not specifying the region in the config). This falls apart with multiple clients due to ambigious bean definition. See:
 https://docs.quarkiverse.io/quarkus-amazon-services/dev/common-features.html

but I guess my approach might be more explicit, but also lacks some flexibility. 

The reason to use qas are the dev services and more powerful configuration options.

holomekc:
Hi @ozangunalp,
new approach to use provided clients. I used the Identifier and Named annotation to allow defining the client. In general I think this is very important. We for example use quarkiverse-amazon-services to create the client, which provides lots of configuration options.

Is this a better approach or is something still missing?

Br,
Chris

ozangunalp:
@holomekc thank you for your contribution! The client provider looks good too!

I've added a quick docs page, and removed an unused attribute.

If this is ok for you, we can get this in

holomekc:
Hi @ozangunalp,
yeah sounds awesome. There seems to be an issue with kafka at the moment, but looks good to me as well.

Br,
Chris

ozangunalp:
Yeah, I cannot seem to fix that correctly.

ozangunalp (approved):

PR #2963, opened 2025-03-09
fix: Inbound and outbound reference

holomekc:
When using a static variable it can lead to strange behavior in tests. E.g. two health check test based on the same direction would create melon tests.

I noticed that while working on the sns connector. I think static is not needed anyway the bean is applicationScoped


ozangunalp:
@holomekc thanks for this, it looks good to me!

ozangunalp:
@holomekc would it be possible to sign your commits? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

holomekc:
Sure. Is signed now

PR #2965, opened 2025-03-09
fix: sqs logging and health check in case batching is used

holomekc:
#2963 should be considered before this pr.


ozangunalp (approved):

holomekc:
Force-push due to signing the commit

ozangunalp (approved):

Issue #2945, opened 2025-02-26
Ability to create IncomingRabbitMQMetadata for testing purposes


djarnis73:
When unit testing using `InMemoryConnector` I could not figure out a way to receive messages containing `IncomingRabbitMQMetadata`, since it does not have any public constructors. See [my stack overflow post](https://stackoverflow.com/questions/79466934/testing-rabbitmq-messaging-containing-metadata-with-inmemoryconnector) for more details.

Would it be possible to allow for manually creation of `IncomingRabbitMQMetadata` or should the `InMemoryConnector` translate the `OutgoingRabbitMQMetadata` into `IncomingRabbitMQMetadata` when the message is processed?

I would love to contribute a PR on this, but would like some guidance on the appropriate way to implement this.

djarnis73:
So, I had a look at the code, and the `IncomingRabbitMQMetadata` is very tightly coupled with `io.vertx.rabbitmq.RabbitMQMessage` so you cannot create the metadata without having a message.

I did try to hack up a quick [PR](https://github.com/smallrye/smallrye-reactive-messaging/pull/2981) and would love to get some feedback.

djarnis73:
How can I gain traction on this? Do I just need to be patient, or are there other channels I can pursue?

ozangunalp:
Fixed with #2981

PR #2981, opened 2025-03-19
feat: allow creation of IncomingRabbitMQMetadata

djarnis73:
From OutgoingRabbitMQMetadata.

This is only a POC to see if this can get some traction, will probably need some polishing and unit tests, or perhaps a completely different approach.

Attempt at resolving #2945

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMetadata.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMetadata.java
@@ -10,6 +10,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import com.rabbitmq.client.BasicProperties;
+import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.LongString;
 
 import io.vertx.codegen.annotations.Nullable;
@@ -20,8 +22,9 @@ import io.vertx.rabbitmq.RabbitMQMessage;
  */
 public class IncomingRabbitMQMetadata {
 
-    private final RabbitMQMessage message;
     private final Map<String, Object> headers;
+    private final BasicProperties properties;
+    private final Envelope envelope;
 
     /**
      * Constructor.
@@ -29,10 +32,15 @@ public class IncomingRabbitMQMetadata {
      * @param message the underlying {@link RabbitMQMessage}
      */
     IncomingRabbitMQMetadata(RabbitMQMessage message) {
-        this.message = message;
+        this(message.properties(), message.envelope());
+    }
+
+    IncomingRabbitMQMetadata(BasicProperties properties, Envelope envelope) {
+        this.properties = properties;
+        this.envelope = envelope;
 
         // Ensure the message headers are cast appropriately
-        final Map<String, Object> incomingHeaders = message.properties().getHeaders();
+        final Map<String, Object> incomingHeaders = properties.getHeaders();
         headers = (incomingHeaders != null) ? incomingHeaders.entrySet().stream()
                 .collect(HashMap::new, (m, e) -> m.put(e.getKey(), mapValue(e.getValue())), HashMap::putAll)
                 : new HashMap<>();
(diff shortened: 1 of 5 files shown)


cescoffier (commented):
Looks good, but we would need some tests.

djarnis73:
Great, I will add some and update the PR.

djarnis73:
So I update the PR to include a unit test, but I also refactored `IncomingRabbitMQMessage` so it does not store the full message, but instead only the two fields it actually access `BasicProperties` and `Envelope`, this simplifies the test code that needs to generate `IncomingRabbitMQMessage`.

If you dislike it, I can undo the refactor and just include the unit test.

I also noticed that there exists two unit tests regarding message metadata `RabbitMQMetadataTest` and `IncomingRabbitMQMetadataTest` I think they could be merged into the former, but that is probably for a different PR.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMetadata.java
```diff
@@ -20,19 +22,25 @@
...
      *
      * @param message the underlying {@link RabbitMQMessage}
      */
     IncomingRabbitMQMetadata(RabbitMQMessage message) {
-        this.message = message;
+        this(message.properties(), message.envelope());
+    }
+
+    IncomingRabbitMQMetadata(BasicProperties properties, Envelope envelope) {
+        this.properties = properties;
```
cescoffier, line 39: Are we sure that this cannot be `null`? I see several accesses in the other methods.

Note that it's not new, it was already like this before.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/OutgoingRabbitMQMetadata.java
```diff
@@ -313,4 +318,34 @@ public OutgoingRabbitMQMetadata build() {
...
+                .correlationId(correlationId)
+                .deliveryMode(deliveryMode)
+                .expiration(expiration)
+                .priority(priority)
+                .messageId(messageId)
+                .replyTo(replyTo)
+                .timestamp(Date.from(timestamp.toInstant()))
+                .type(type)
+                .build();
+        Envelope envelope = new Envelope(1, isRedeliver, exchange, routingKey);
```
cescoffier, line 346: For my own understanding, why 1?

cescoffier (approved):
LGTM - I just added 2 comments.

Note that a word in the documentation would be nice!

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/OutgoingRabbitMQMetadata.java
djarnis73: I should have put a comment about it, but the field is never used in the in the `IncomingRabbitMQMessage` so it does not really matter, I think I may have copied the values from some other code I found but `0`might be a better value.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/IncomingRabbitMQMetadata.java
djarnis73: Did not think about that, do you have a standard for how you handle this? I can add `Objects.requireNonNull()` in constructor to ensure NPE on creation.

djarnis73:
> Note that a word in the documentation would be nice!

Sure, will try to add that as well.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/OutgoingRabbitMQMetadata.java
djarnis73: I have force pushed an update that used `0` instead of `1` with a comment on why it is hardcoded.

ozangunalp (dismissed):
This makes sense to me. @djarnis73 thank you for your contribution!

Could you sign your commits for us to merge the PR?

ozangunalp (approved):

Issue #2973, opened 2025-03-13
AMQP should use a retry failure attribute instead of reconnect attribute when message is rejected


soframel:
**Use case:** 
Sending a message in AMQP 1.0 to RabbitMQ 4, on a queue configured with a max-length and `x-overflow : reject-publish'.
This is similar to [2726](https://github.com/smallrye/smallrye-reactive-messaging/issues/2726) (solved by @gcuisinier in https://github.com/smallrye/smallrye-reactive-messaging/pull/2729) except that we are using the amqp module instead of the rabbitmq module. 

**Expected:** 
As the queue is full, the message is rejected and the error should get raised relatively quickly.

**Current Behaviour**
As the nack is considered a failure, sending is retried, but this uses the “reconnect attempts” and “reconnect interval” attributes, which are usually high, so the error is not sent to the producer quickly.  

https://github.com/smallrye/smallrye-reactive-messaging/blob/ee134d7dbb0e964fc26e333c8f83426b57aa1c24/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java#L242-L247

Note: it’s not clear for me yet if in AmqpCreditBasedSender the same option should be used:
https://github.com/smallrye/smallrye-reactive-messaging/blob/ee134d7dbb0e964fc26e333c8f83426b57aa1c24/smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java#L287-L290

**Proposed Solution**
Similarly to the rabbitmq module, I think it would be better to introduce “retry-on-fail-attempts” and “retry-on-fail-interval” attributes? do you think it would make sense?
should we propose a pull request?

soframel:
Hello, any news on this? I would like to propose a pull request: is it ok for you?

cescoffier:
Sure, contributions are very welcome!

PR #2984, opened 2025-03-26
add retry-on-fail-attemps and retry-on-fail-interval outgoing properties

soframel:
Closes https://github.com/smallrye/smallrye-reactive-messaging/issues/2973


ozangunalp (approved):

soframel:
sorry for the force push, I had to amend my commit for signature

ozangunalp:
@soframel I don't think the commit has a verified signature yet

soframel:
> @soframel I don't think the commit has a verified signature yet

sorry about that, should be ok now

soframel:
The build is failing but I don't think it's related to the proposed change? it's a Kafka test...

PR #3005, opened 2025-04-14
Properly close the worker executor

cescoffier:


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -54,7 +55,11 @@ public void terminate(
             @Observes(notifyObserver = Reception.IF_EXISTS) @Priority(100) @BeforeDestroyed(ApplicationScoped.class) Object event) {
         if (!workerExecutors.isEmpty()) {
             for (WorkerExecutor executor : workerExecutors.values()) {
-                executor.close();
+                if (Infrastructure.canCallerThreadBeBlocked()) {
```
ozangunalp, line 58: I didn't know that method

ozangunalp (approved):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
cescoffier: It's from Mutiny. When you integrate Mutiny into your framework/server, you need to provide the proper implementation.

cescoffier: By default, it's already true - but it will be different in Quarkus.

PR #3015, opened 2025-04-23
Worker pool graceful shutdown

ozangunalp:


smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -40,26 +42,61 @@
...
     private ExecutionHolder holder;
 
     public void terminate(
             @Observes(notifyObserver = Reception.IF_EXISTS) @Priority(100) @BeforeDestroyed(ApplicationScoped.class) Object event) {
         if (!workerExecutors.isEmpty()) {
-            for (WorkerExecutor executor : workerExecutors.values()) {
-                if (Infrastructure.canCallerThreadBeBlocked()) {
-                    executor.closeAndAwait();
-                } else {
+            // In most cases this is called from the main thread, so we can block.
```
cescoffier, line 63: We may need to add flag to reject any new execution.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -40,26 +42,61 @@
...
+                                executor.shutdownNow();
+                            } else {
+                                terminated = false; // Still waiting
+                            }
+                        }
+                    } catch (InterruptedException ignored) {
+                        terminated = false;
+                    }
+                }
             }
```
cescoffier, line 100: Should we close the worker executor when done?

cescoffier (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
```diff
@@ -74,6 +116,9 @@ public void init() {
     }
 
     public <T> Uni<T> executeWork(Context msgContext, Uni<T> uni, String workerName, boolean ordered) {
+        if (closed) {
+            throw new IllegalStateException("WorkerPoolRegistry is closed");
```
cescoffier, line 120: Should it be a rejected operation exception, to follow the executor semantic?

cescoffier (changes requested):

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/connectors/WorkerPoolRegistry.java
ozangunalp: I don't have a strong opinion on this. I've considered rejected operation exception, but it is typically thrown by `Executor`s. We also liberally use `UnsupportedOperationException`, while it is part of the Java Collections Framework.

cescoffier (approved):

PR #3018, opened 2025-04-25
Accept smallrye- as default connector name prefix for connector config

ozangunalp:
The channel attribute `connector` can omit the `smallrye-` prefix and still discover the right connector.


cescoffier (approved):

ozangunalp:
Recent flaky tests are starting to become annoying. I see flakiness in MQTT or AMQP tests (with RabbitMQ container).

PR #3050, opened 2025-05-26
Dark mode for docs

### LINKED ISSUE #2899: Add support for dark mode
It would be great to have a dark mode option for the documentation site. Many users prefer using dark themes for prolonged reading sessions, as it reduces eye strain and provides a more comfortable experience, especially in low-light environments.

ozangunalp:
Fixes #2899

### DIFF
--- a/documentation/mkdocs.yml
+++ b/documentation/mkdocs.yml
@@ -126,8 +126,31 @@ theme:
   logo: images/logo.png
   favicon: images/favicon.png
   palette:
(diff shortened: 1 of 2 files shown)


cescoffier:
We might need to change the pictures.

ozangunalp:
> We might need to change the pictures.

This bit is there for that: 
```
    figure {
        filter: invert(1) hue-rotate(180deg);
    }
```

PR #3060, opened 2025-06-06
Make sure the duplicated context duplication does not inherit locals from the owner duplicated context.

cescoffier:


cescoffier:
@ozangunalp what do you want to do with this one?

ozangunalp (approved):

ozangunalp:
At least we are sure now of the behaviour we'll get

PR #3074, opened 2025-06-24
Update for MP Telemetry 2.1

jasondlee:
Update MP Reactive Messaging for compliance/integration with MP Tel 2.1

Bump version numbers
Fix semantic convention issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,8 +89,9 @@
 
     <kafka.version>4.0.0</kafka.version>
 
-    <opentelemetry.instrumentation.version>2.5.0-alpha</opentelemetry.instrumentation.version>
-    <opentelemetry-semconv.version>1.25.0-alpha</opentelemetry-semconv.version>
+    <opentelemetry.instrumentation.version>2.14.0</opentelemetry.instrumentation.version>
+    <opentelemetry.version>1.48.0</opentelemetry.version>
+    <opentelemetry-semconv.version>1.32.0</opentelemetry-semconv.version>
 
     <smallrye-vertx-mutiny-clients.version>3.19.1</smallrye-vertx-mutiny-clients.version>
     <smallrye-reactive-converters.version>3.0.0</smallrye-reactive-converters.version>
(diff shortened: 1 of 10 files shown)


jasondlee:
FTR: after integrating this update into my local WF build -- and making some WF module changes to pick up the "incubating" semconv jars -- the failing tests in the MP 7.1 topic branch pass.

cescoffier:
@brunobat FYI - are we using the same version in Quarkus

@jasondlee what do you mean by "fix semantic conventions"? Did they change? Are we breaking dashboards?

cescoffier:
Hum, so we cannot really update - we would need 2 branches...

jasondlee:
> @brunobat FYI - are we using the same version in Quarkus
> 
> @jasondlee what do you mean by "fix semantic conventions"? Did they change? Are we breaking dashboards?

Two things: the codebase was reorganized in newer versions of `io.opentelemetry.semconv:opentelemetry-semconv`, so, in upgrading, there we compilation failures. Those are easy to fix and invisible to the end users. Additionally, there were deprecations for some attribute names, but those can be ignored relatively safely for now if we choose to do so. Some names, though, were changed (e.g., `messaging.client_id` -> `messaging.client.id`) or just completely removed. For those, dashboards could very well break. I'm not sure we have a good solution for that. It's an issue that we've had to face (and, I think, accept) in WF/EAP as our options seem to be breaking dashboards as we upgrade, or try to maintain some sort of compile-/run-time compatibility layer. The conventions have finally hit a certain level of stability, so breakage moving forward should be minimal, but, as you can see in the diff, some attribute names are from `io.opentelemetry.semconv:opentelemetry-semconv-incubating`, which means they very well may change in future releases.

cescoffier:
It would be great to avoid having to maintain two branches (one for quarkus and one for WildFly). 
If we update, we risk breaking Quarkus, so we need to coordinate to ensure a safe update. It would also require a new minor, as we would not backport this to Quarkus LTS.

cescoffier:
@brunobat, we should either consider it now (or by the beginning of July at the latest) or mid-August after the LTS feature freeze.

brunobat:
Back from PTO. Will know more about the timeline soon.

brunobat:
> > Back from PTO. Will know more about the timeline soon.
> 
> @brunobat I've updated my PR based on main. Any update on a timeline?

Not before the end of next week.

brunobat:
> > Do you think we should also include old otel-semver attributes in the spans?
> 
> @brunobat I'd like to have your opinion on this.

Not much we can do because they do this type of think frequently. We need to issue the changes in the release notes.

ozangunalp (approved):

brunobat:
Messaging semantic conventions are still incubating and we should move on as they break things, @ozangunalp. We need to document all the changes, though.

PR #3221, opened 2025-11-07
Add Otel tracing to SQS connector

### LINKED ISSUE #3188: No traces produces when using reactive-messaging with sqs connector
Hello everyone, I'm using reactive messaging with sqs connector to read messages from my queue. Is there any chance that I can instrument my setup with OTEL to see correlation between traces of sqs messages and traces from my code? I have currently added `quarkus-opentelemetry` as dependency and set following properties:
```
quarkus:
  otel:
      instrumentation:
        messaging:
          experimental:
            receive-telemetry:
              enabled: true
        aws-sdk:
          experimental-use-propagator-for-messaging: true
          experimental-record-individual-http-error: true
mp:
  messaging:
    incoming:
      queue:
        tracing-enabled: true
```
But I don't see any traces in my obervability tools, besides that logs from code executed by queue does not have any trace_id attribute. Have someone by any chance did such setup and maybe can help me to configure it properly?

ozangunalp:
Fixes #3188

### DIFF
--- a/smallrye-reactive-messaging-aws-sqs/pom.xml
+++ b/smallrye-reactive-messaging-aws-sqs/pom.xml
@@ -35,6 +35,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>io.smallrye.reactive</groupId>
+      <artifactId>smallrye-reactive-messaging-otel</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.microprofile.config</groupId>
       <artifactId>microprofile-config-api</artifactId>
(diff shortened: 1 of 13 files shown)


ozangunalp:
@majchrzw did you have time to give it a try ?

majchrzw:
I will try to do some tests this or next week.

ozangunalp:
@majchrzw Thanks for giving this a try. 

It is cool that most is working correctly. 

I think for your expectations on trace spans and exception context, you need to annotate the processing method with  `@WithSpan` so the processing itself has an associated span. The parent propagation with incoming span will be automatic.

cescoffier (approved):

PR #3242, opened 2025-11-27
Add MessageConverters for IncomingRabbitMQMessage

### LINKED ISSUE #3220: RabbitMQ- Allow Deserialization to complex objects
Similiar to #465, it would be great if for the RabbitMQ connector we could automatically **deserialize a JSON Payload to a given complex object type**.

Something like:
```java
   @Incoming("prices")
    public CompletionStage<Void> consume(Message<Price> price) {
             //...
    }
   
   class Price {
          double amount;
    }

``` 
Currently we can only [declare the payload of incoming RabbitMQ Message](https://quarkus.io/guides/rabbitmq-reference#deserialization) as: `byte[]`, `String`, `JsonObject`, `JsonArray`

Crustack:
Closes #3220 

I looked at how e.g. MQTT messages are converted via e.g. [StringMessageConverter](https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-mqtt/src/main/java/io/smallrye/reactive/messaging/mqtt/converter/StringMessageConverter.java) and came up with the following changes:
- Removed the String/generic JSON deserialization from `IncomingRabbitMQMessage.convertPayload`
- Added `MessageConverter` implementations for deserialization to `String`, `JsonObject`, `JsonArray` to comply with current behaviour
- Added `MessageConverter` implemenation for deserialization to any type if a `JsonMapping` implementation is on the classpath
- Added `fromJson(String str, Type type)` to `JsonMapping` interface with implementations for jackson + jsonb (Vert.x does not seem to support mapping for a `java.lang.reflect.Type` ?)


ozangunalp:
@Crustack This is my next on this list. I'd like to include your contribution, but I am guessing it'll need some adjusting after #3262.

Crustack:
> @Crustack Again, thanks for your PR! I rebased and cleaned up the code a bit.
> 
> If I try to explain some of the changes I've made:
> 
> * We are trying to rollback from using `Message` subtypes in application-level code. This is why the metadata type (`IncomingRabbitMQMetadata`) acts more like a contract for the connector than the message.
> * So I pulled the Vert.x RabbitMQ message to the metadata to allow accessing the buffer in converters.
> * The message no longer does any payload conversion; I kept it a parameterised class, but it holds the buffer as payload.
> * I've added a ByteArrayMessageConverter to keep the ability to consume `byte[]`
> 
> If you are ok with this, I'd be happy to merge.

Looks good to me 👍🏼

PR #3244, opened 2025-12-01
Fix: healthcheck down when initially paused

### LINKED ISSUE #3238: Kafka: health-check is down when a channel is "initially-paused"
The default strategy for the health-check tests that there is at least one connection to the broker (metric `connection-count` of the kafka-client). This works well when the channel is running or paused after running for a while. But if the channel is configured as `pausable=true` and `initially-paused=true` the connection count is always zero and thus the health check is down.

This is an issue if you have a started probe in kubernetes with an initially paused channel.

Somehow the problem is not present if the channel is not initially paused but you programmatically pause it later. My guess is that "initially paused" doesn't even start a connection to the kafka broker.

I was able to reproduce this problem in unit test, ~I will upload a reproducer~ reproducer: #3239.

I think the health check should check if the channel is paused, and return "UP" if it's the case. I'm happy to propose a patch if you think this is the right direction to go?

gaetancollaud:
Closes #3238

In addition to using `connection-count` for the health probe, we now check how many `connection-creation-total` were attempted.

The initial issue was that when a connection is `initially-paused` it won't really start the kafka consumer and the `connection-count` would be at zero => health check DOWN.

Now when the connection count is zero, we check how many connections were attempted. If zero => UP since the connector is just paused since the beginning.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/KafkaSinkHealth.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/health/KafkaSinkHealth.java
@@ -11,13 +11,13 @@ import org.apache.kafka.clients.admin.DescribeTopicsOptions;
 import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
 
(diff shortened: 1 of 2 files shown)


cescoffier (approved):
LGTM!

It's a nice approach!

gaetancollaud:
> Ok, I only pushed a test to verify the authentication failure case. It looks good to me.
> 
> @gaetancollaud Thanks for your contribution!

Thanks @ozangunalp  for the review and doing the additional test yourself, you were too fast for me.

Thanks @cescoffier

PR #3266, opened 2025-12-24
RabbitMQ: shared connections with context safety and config mismatch guard

### LINKED ISSUE #3189: [smallrye-rabbitmq] Proposal/Question: Allow Multiple Channels to Share a Single RabbitMQ Connection
## Summary
Today each `mp.messaging` channel that uses the `smallrye-rabbitmq` connector opens its own AMQP connection. In practice this means one emitting channel plus one consuming channel already establish two independent TCP connections, even if they target the same broker/exchange pair. The RabbitMQ documentation recommends reusing a single connection and opening multiple AMQP channels on top of it, so the current behaviour makes it easy to exhaust broker connection quotas in larger deployments.

## Current Behaviour
- `IncomingRabbitMQChannel` creates a fresh `RabbitMQClient` per channel and memoizes its connection.
- `OutgoingRabbitMQChannel` follows the same pattern for producers.

Because the connector instantiates a dedicated client per channel, there is no supported way to share a single TCP connection across multiple channels.

## Why This Is Problematic
- Hosted RabbitMQ clusters typically enforce connection quotas; with the current design, the number of connections grows linearly with the number of channels, making it easy to hit those limits.
- RabbitMQ best practices suggest opening one connection per app instance and multiplexing via channels instead of creating many short-lived connections.
- Users migrating from “bare” clients (e.g. Vert.x, Spring AMQP) are surprised that SmallRye cannot follow the same pattern.

## Questions for Maintainers
1. Was there a historical reason to tie one `RabbitMQClient` (connection) to each messaging channel? E.g. compatibility, back-pressure, or resilience constraints?
2. Would the project consider a feature enhancement that allows multiple channels to share a single connection (for example by introducing a connection pool keyed by host/virtual-host + credentials)?
3. If the concern is isolation/failure propagation, could the connector offer a toggle so users can opt into shared connections while keeping the current default?

## Desired Outcome
- Clarify the design rationale so we can document it internally.
- If there is no blocker, collaborate on an enhancement (or contribute one) that lets applications reuse existing connections, reducing broker load.

Note: If I misunderstood and am using the tool incorrectly, please explain to me the best way to use it.

CassioGangrel:
### Summary

* Add shared-connection-name support to reuse a single RabbitMQ connection across channels.
* Preserve incoming event-loop context even when a shared connection is created by outgoing first.
* Add a fingerprint check so channels using the same shared-connection-name must have identical connection settings.
* Add/adjust tests for shared connections and context.

### Configuration

Example:

```
mp.messaging.incoming.data.shared-connection-name=shared-connection
mp.messaging.outgoing.sink.shared-connection-name=shared-connection
```

If two channels share the same name but have different connection settings, startup fails with:
`Shared connection '<name>' has mismatched configuration...`

### Breaking changes / Documentation

* No breaking changes (feature is opt-in).
* Suggest documenting shared-connection-name and the config mismatch guard.
* Testing  `./mvnw -pl smallrye-reactive-messaging-rabbitmq -DskipITs test`

### Simple flowchart:

```mermaid
flowchart TD
  IN[Incoming channel] -->|shared-connection-name| CONN[RabbitMQConnector]
  OUT[Outgoing channel] -->|shared-connection-name| CONN

  CONN -->|getClientHolder| POOL[sharedClients by name]
  POOL -->|compute fingerprint| FP[compare or create]

  FP -->|ok| CLIENT[(RabbitMQClient shared)]
  FP -->|mismatch| ERR[fail startup]

  CLIENT --> INCTX[Incoming event-loop context]
  CLIENT --> OUTCTX[Outgoing caller context]
  CLIENT --> MET[metrics/connection name using shared label]
```

Fixes #3189

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
     <testcontainers.version>2.0.4</testcontainers.version>
 
     <sonar.coverage.jacoco.xmlReportPaths>${project.basedir}/target/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
-    <mockito.version>5.22.0</mockito.version>
+    <mockito.version>5.23.0</mockito.version>
     <junit-pioneer.version>2.3.0</junit-pioneer.version>
     <junit-platform-commons.version>1.9.3</junit-platform-commons.version>
     <system-stubs-jupiter.version>2.1.8</system-stubs-jupiter.version>


smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
```diff
@@ -89,6 +63,17 @@ public Context getContext() {
         }
     }
 
+    public void ensureContext(Context context) {
```
ozangunalp: Won't the shared client switch context with each channel with shared connection ?

CassioGangrel: No. In a shared connection, the `ClientHolder` keeps a single connection context set once via `ensureContext (compare-and-set)` and it does not switch per channel. Each `IncomingRabbitMQChannel` creates its own context for message processing and passes it to `IncomingRabbitMQMessage`, so `ack/nack` runs on the channel context, not a per-channel connection context.

Issue #3253, opened 2025-12-11
[smallrye-jms] Producer stops sending messages after the internal connection is broken (autorecoverable producer)


MRLimcon:
Good morning, afternoon or evening!

We're using smallrye-jms and we're seeing some problems with the producers (when the internal connection on the JMSContext gets broken and doesn't reconnect), it seems the consumers (JmsSource) are recovered every retry:

```java
                .onFailure(t -> {
                    log.terminalErrorOnChannel(channel);
                    this.resourceHolder.close();
                    return retry;
                })
                .retry()
```

but the producers (JmsSink) aren't:

```java
    sink = MultiUtils.via(m -> m.onItem().transformToUniAndConcatenate(message -> send(resourceHolder, message)
                .onFailure(t -> retry)
                .retry()
```

Trying to fix this issue, and dealing with the jakarta jms spec, we found that creating a producer doesn't talk to the broker, so forcing to rebuild the context (creating a new one from the connection factory) works.

So I want to ask if the team would accept a PR for putting a rebuild of the context on any failure to send to the producer (I think the only sync method on the producer) or if you would want a wrapper library that handles all of this internally, but it would override the reuse-jms-context, as creating child contexts seem to not talk to the broker too (on our broker).

ozangunalp:
Resolved by #3269

PR #3269, opened 2026-01-05
[smallrye-jms] Autorecoverable producer

MRLimcon:
Adding a recovery path for failures to send to the producer, closing the context, for the JmsResourceHolder to create a new context when called on the next try.

Related to the issue #3253.


ozangunalp:
rebased and added tests

PR #3319, opened 2026-02-20
Support for Kafka Queues (Share consumer)

ozangunalp:
Provides a new Kafka Source to support Kafka Queues (Share Consumer client).

Kafka companion: `shareConsume*` methods 

Kafka Connector: 
- `share-group` flag to create Kafka Share Consumer client. 
- `ReactiveKafkaShareConsumer` with polling stream without pause/resume. -> Keeps polling the consumer and retries on `IllegalStateException` thrown by the consumer. 
- `ReactiveKafkaAdminClient` methods for share Consumer description
- Fixed commit and failure strategies to handle per-message acknowledgement. -> Tracks received messages and listens to committed acks. Applies a timeout for pending records that are never acknowledged. Sends RENEW acknowledgements for messages still in-processing 
- Graceful shutdown waits for pending messages, trying to commit at each iteration.
- Topic-based Health check to check partition assignments to the current client.
- `ShareGroupAcknowledgment` metadata enables processing code to decide between different per-message ack type. Same nack metadata. 
- Batch consuming mode with per-message acking.
- Deserialization failure handler flags to choose which ack type to apply on deserialization failure.

### DIFF
--- a/documentation/src/main/docs/kafka/receiving-kafka-records.md
+++ b/documentation/src/main/docs/kafka/receiving-kafka-records.md
@@ -806,6 +806,144 @@ The checkpoint commit strategy calls the state store in following events:
 - `close` on channel shutdown.
 
 
+## Receiving Kafka Records Using Kafka Queues (Share Groups)
+
+!!!warning "Experimental"
+    Kafka Queues (Share Groups) support is experimental.
+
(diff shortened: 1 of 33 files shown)


cescoffier (approved):
Super interesting!

PR #3331, opened 2026-03-06
Switch to Vert.x 5 and New RabbitMQ connector

cescoffier:
And remove the MQTT connector for now.


ozangunalp:
AMQP 1.0 tests are failing on the CI, because the embedded Artemis AMQP broker used for tests depend on netty 4.1 : 
https://mvnrepository.com/artifact/org.apache.activemq/artemis-jakarta-server/2.44.0/dependencies

Going forward, I think it'll be better to migrate these tests to use a containerised broker.

ozangunalp:
Superceded by #3492

PR #3353, opened 2026-03-26
RabbitMQ: shared connections with context safety and config mismatch guard

ozangunalp:
Clone of #3266


CassioGangrel:
Excellent work @ozangunalp , and sorry for my absence. I was away from GitHub for a while. Thank you for completing the work.

Issue #1983, opened 2022-11-13
Add support for request/reply


cescoffier:
Using a specific emitter type it should be possible to implement the request/reply interaction scheme for connectors supporting it (RabbitMQ, AMQP 1.0).

econte66:
Hi there!
@cescoffier : Are there an entry feature in mp.messaging.incoming or mp.messaging.outgoing to use the RPC RabbitMQ in Quarkus? The solution I found is to use the "client" rabbitmq library and AMQP 1.0.
Could you help me find the correct solution?
Thanks
Enrico

ozangunalp:
Resolved by #3301, #3357 and #3382

PR #3357, opened 2026-03-29
Add RabbitMQRequestReply implementation

Crustack:
Adds Request/Reply implementation for RabbitMQ as proposed in #1983 

#### Summary 
- Adds `RabbitMQRequestReply` Emitter type to support [RPC over RabbitMQ](https://www.rabbitmq.com/tutorials/tutorial-six-java)
- Heavily based on [Kafka Request Reply implementation](https://github.com/smallrye/smallrye-reactive-messaging/pull/2409) by @ozangunalp 
-  Leverages [RabbitMQ Direct Reply-To](https://www.rabbitmq.com/docs/direct-reply-to) by opening an `IncomingRabbitMQChannel` on special `amq.rabbitmq.reply-to` "queue", emitting a request message to configured exchange + routing key and waiting for response message with correct `correlationId`
_Note: Since the requestor needs to publish the request and consume the response with the same connection, the `shared-connection-name` is set to achieve this_
- A Replier/RPC-Server would simply need to publish a message on the default exchange with the correct `routingKey` (generated `amq.rabbitmq.reply-to...`) and `correlationId` to send a response back (see e.g. [Replier.java](https://github.com/Crustack/smallrye-reactive-messaging/blob/feat/rabbitmq-request-reply/examples/rabbitmq-quickstart/src/main/java/acme/Replier.java))
- Same customization options as for `KafkaRequestReplyImpl`, including `correlationIdHandler`, `replyFailureHandler`, `replyTimeout`, `replyConverter`

### DIFF
--- a/documentation/src/main/docs/rabbitmq/request-reply.md
+++ b/documentation/src/main/docs/rabbitmq/request-reply.md
@@ -5,7 +5,7 @@
 
 The RabbitMQ [Request-Reply](https://www.enterpriseintegrationpatterns.com/patterns/messaging/RequestReply.html) pattern allows you to publish a message to a RabbitMQ address and then await for a reply message that responds to the initial request, which can be used to implement [RPC over RabbitMQ](https://www.rabbitmq.com/tutorials/tutorial-six-java).
 
-Given the following request/reply outgoing configuration:
+Given the following request/reply outgoing configuration example:
 
 ```properties
 mp.messaging.outgoing.my-request.connector=smallrye-rabbitmq
@@ -19,6 +19,10 @@ The `RabbitMQRequestReply` emitter implements the requestor (or the client) of t
 ``` java
 {{ insert('rabbitmq/outbound/RabbitMQRequestReplyEmitter.java') }}
 ```
+
+!!! note
+    While in this example we use a direct exchange, it is also possible to use a topic or fanout exchange as well as [Local Random Exchanges](https://www.rabbitmq.com/docs/local-random-exchange) for request/reply.
+
 
 The `request` method publishes the request message to the configured target address of the outgoing channel leveraging RabbitMQ [Direct Reply-To](https://www.rabbitmq.com/docs/direct-reply-to),
 and waits for a reply.


ozangunalp:
@Crustack Thanks for this, this looks super promising. I haven't yet checked everything, but I think we can automatically set the `shared-connection-name` on the internal reply consumer channel.

Crustack:
> I found that out yesterday. I thought of if we reused the connection name as the metrics name.

I pushed those changes 👍🏼


About simplifying the Replier,
Would it suffice to do this in the [RabbitMqMessageConverter.convert](https://github.com/smallrye/smallrye-reactive-messaging/blob/3a39eabd3c9cfe5acf2c96dfc1544429dc959cfa/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageConverter.java#L64)?
```java
      final String routingKey = getRoutingKey(message)
                .or(() -> message.getMetadata(IncomingRabbitMQMetadata.class).flatMap(IncomingRabbitMQMetadata::getReplyTo))
                .orElse(defaultRoutingKey);
                //...
        if (rabbitMQMessage.isPresent()) {
                //...
        } else {
           //...
            final OutgoingRabbitMQMetadata metadata = message.getMetadata(OutgoingRabbitMQMetadata.class)
                    .orElseGet(() -> {
                        var builder = new OutgoingRabbitMQMetadata.Builder()
                                .withContentType(defaultContentType)
                                .withExpiration(defaultTtl.map(String::valueOf).orElse(null));
                        message.getMetadata(IncomingRabbitMQMetadata.class).ifPresent(incomingMetadata -> {
                            builder.withRoutingKey(incomingMetadata.getReplyTo().orElse(null))
                                    .withCorrelationId(incomingMetadata.getCorrelationId().orElse(null));
                        });
                        return builder.build();
                    });
        }
```
It works as long as the Replier only returns the outgoing payload, but I'm not sure if this would break other use-cases where `@Incoming` and `@Outgoing` is used on a method in a non-direct-replyTo scenario?

ozangunalp:
> About simplifying the Replier, Would it suffice to do this in the [RabbitMqMessageConverter.convert](https://github.com/smallrye/smallrye-reactive-messaging/blob/3a39eabd3c9cfe5acf2c96dfc1544429dc959cfa/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageConverter.java#L64)?
> 
> ```java
>       final String routingKey = getRoutingKey(message)
>                 .or(() -> message.getMetadata(IncomingRabbitMQMetadata.class).flatMap(IncomingRabbitMQMetadata::getReplyTo))
>                 .orElse(defaultRoutingKey);
>                 //...
>         if (rabbitMQMessage.isPresent()) {
>                 //...
>         } else {
>            //...
>             final OutgoingRabbitMQMetadata metadata = message.getMetadata(OutgoingRabbitMQMetadata.class)
>                     .orElseGet(() -> {
>                         var builder = new OutgoingRabbitMQMetadata.Builder()
>                                 .withContentType(defaultContentType)
>                                 .withExpiration(defaultTtl.map(String::valueOf).orElse(null));
>                         message.getMetadata(IncomingRabbitMQMetadata.class).ifPresent(incomingMetadata -> {
>                             builder.withRoutingKey(incomingMetadata.getReplyTo().orElse(null))
>                                     .withCorrelationId(incomingMetadata.getCorrelationId().orElse(null));
>                         });
>                         return builder.build();
>                     });
>         }
> ```
> 

This is what I had in mind, too. I believe we can further clean up the code later.


> It works as long as the Replier only returns the outgoing payload, but I'm not sure if this would break other use-cases where `@Incoming` and `@Outgoing` is used on a method in a non-direct-replyTo scenario?

It is a good question. I think it is if the incoming message has a replyTo, and it is not already handled by the application code (if the routing key hasn't been provided), we can "safely" set it automatically.

Crustack:
> It is a good question. I think it is if the incoming message has a replyTo, and it is not already handled by the application code (if the routing key hasn't been provided), we can "safely" set it automatically.

Sounds good.
One thing I do find a bit annoying with my proposed changes in `RabbitMQMessageConvert.convert` is that as soon as the `@Outgoing` returns a `Message<?>` it wont pick up the `replyTo` and `correlationId` at all.
E.g. for the [ReplyServerWithFailure](https://github.com/Crustack/smallrye-reactive-messaging/blob/a7cdfe6d49f4a8c9b26fe52bb47130f8c597f998/smallrye-reactive-messaging-rabbitmq/src/test/java/io/smallrye/reactive/messaging/rabbitmq/reply/RabbitMQRequestReplyTest.java#L390) that means we have to build the `OutgoingRabbitMQMetadata` in its entirety manually, just to add some error as a header:
```java
@Incoming("req")
@Outgoing("rep")
public Message<String> replier(Message<String> message) {
    IncomingRabbitMQMetadata metadata = message.getMetadata(IncomingRabbitMQMetadata.class).get();
    String payload = message.getPayload();
    Builder outgoing = OutgoingRabbitMQMetadata.builder()
            .withCorrelationId(metadata.getCorrelationId().get())
            .withRoutingKey(metadata.getReplyTo().get());
    if (Integer.parseInt(payload) % 3 == 0) {
        outgoing.withHeader("REPLY_ERROR", "Cannot reply to " + payload);
    }
    return Message.of(payload).addMetadata(outgoing.build());
}
```
It would be much more convenient if we could:
```java
@Incoming("req")
@Outgoing("rep")
public Message<String> replier(String payload) {
    if (Integer.parseInt(payload) % 3 == 0) {
        return Message.of(payload).addMetadata(OutgoingRabbitMQMetadata.builder()
                .withHeader("REPLY_ERROR", "Cannot reply to " + payload));
    }
    return Message.of(payload);
}
```
And then the provided `OutgoingRabbitMQMetadata` would be merged with the `replyTo` and `correlationId` of the `IncomingRabbitMQMetadata`.
But I guess that would break things in other scenarios?
If yes, maybe we could at least provide a shorter way to create the `OutgoingRabbitMQMetadata.Builder` from a given message, like:
```java
@Incoming("req")
@Outgoing("rep")
public Message<String> replier(Message<String> message) {
    String payload = message.getPayload();
    Builder outgoing = OutgoingRabbitMQMetadata.builderForDirectReply(message);
    if (Integer.parseInt(payload) % 3 == 0) {
        outgoing.withHeader("REPLY_ERROR", "Cannot reply to " + payload);
    }
    return Message.of(payload).addMetadata(outgoing.build());
}
```
```java        
public static Builder builderForDirectReply(Message<String> incomingMessage) {
    return incomingMessage.getMetadata(IncomingRabbitMQMetadata.class)
            .map(incomingMetadata -> OutgoingRabbitMQMetadata.builder()
                    .withCorrelationId(incomingMetadata.getCorrelationId().orElse(null))
                    .withRoutingKey(incomingMetadata.getReplyTo().orElse(null)))
            .orElse(OutgoingRabbitMQMetadata.builder());
}
```

ozangunalp:
@Crustack, I've pushed a change to extract the correlation ID and routing key from replyTo. If an incoming message contains replyTo, I think it would be safe to honour it. 

Also, I've been thinking about why request-reply would be restricted to the direct exchange type. 
There is at least the https://www.rabbitmq.com/docs/local-random-exchange, but also topic and fanout could be used. 

I've done some tests with it. I'll also push that for you to check. 

Thanks!

Crustack:
> @Crustack, I've pushed a change to extract the correlation ID and routing key from replyTo. If an incoming message contains replyTo, I think it would be safe to honour it.
> 
> Also, I've been thinking about why request-reply would be restricted to the direct exchange type. There is at least the https://www.rabbitmq.com/docs/local-random-exchange, but also topic and fanout could be used.
> 
> I've done some tests with it. I'll also push that for you to check.
> 
> Thanks!

Looks great, and definitely makes it more intuitive to work with.
I added a minor note to the documentation to hint that the other exchange types can be used as well

From my part this is good to go 🙂

ozangunalp (approved):

ozangunalp:
@Crustack thanks again for this contribution!

PR #3371, opened 2026-04-15
Split CI into per-connector parallel test jobs

ozangunalp:


ozangunalp:
@cescoffier take a look at the CI when you've some time. By separating connector tests in different jobs we can go from 50min down to 11min.

For PR gating, I have a CI status job that checks test results.

cescoffier (approved):

PR #3378, opened 2026-04-20
Stop / Resume reading from Broker

dometec:
This is a Draft PR since use vertx-mqtt compiled with https://github.com/vert-x3/vertx-mqtt/pull/267 and https://github.com/vert-x3/vertx-mqtt/pull/268 not released yet.

The first commit add a test to show the error where the application message buffer is full.

In the second commit I've added the logic to stop and restart the socket read when the buffer is going to be almost full.
Added also some documentation in receiving-mqtt-messages.md, I hope that clear the problem and the parameters usage.

The attirbutes "receive-buffer-size", "recv-bytebuf-allocator-size" have direction = INCOMING_AND_OUTGOING becouse are used by the mqtt client that can be common between incoming and outgoing channel.


dometec:
I tested this feature with some private application module and it works well. I'll switch this PR in ready on the first release of verts-mqtt.

PR #3382, opened 2026-04-24
AMQP 1.0 RequestReply emitter

ozangunalp:


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
```diff
@@ -0,0 +1,236 @@
...
+                    if (replyFailureHandler != null) {
+                        Throwable failure = replyFailureHandler.handleReply(
+                                (AmqpMessage<?>) m);
+                        if (failure != null) {
+                            return Uni.createFrom().failure(failure);
+                        }
+                    }
+                    return Uni.createFrom().item(m);
+                })
+                .onTermination().invoke(() -> pendingReplies.remove(correlationId))
```
cescoffier, line 157: I can't remember if this is also called on cancellation - if not it could leak.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/ConnectionHolder.java
```diff
@@ -112,61 +106,73 @@ public synchronized void onFailure(Consumer<Throwable> callback) {
...
+
+    private Uni<AmqpConnection> establishConnection() {
+        CurrentConnection current = holder.get();
+        if (current != null && current.connection != null
+                && !current.connection.isDisconnected()) {
+            return Uni.createFrom().item(current.connection);
+        }
+
+        CompletableFuture<AmqpConnection> existing = ongoingConnection.get();
+        if (existing != null) {
```
cescoffier: Another thread could complete the existing future and set `ongoingConnection` back to null. This creates a window where multiple threads could start concurrent connection attempts. I'm pretty sure we have this already.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Override
+    public Map<CorrelationId, PendingReply> getPendingReplies() {
+        return new HashMap<>(pendingReplies);
+    }
+
+    @Override
+    public void onSubscribe(Flow.Subscription subscription) {
+        if (Subscriptions.setIfEmpty(this.subscription, subscription)) {
```
cescoffier, line 174: just in case, complete is not called, a pre-destroy could be useful to cancel the subscription.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
```diff
@@ -0,0 +1,236 @@
...
+                                pendingReplies.remove(correlationId);
+                                emitter.fail(failure);
+                            });
+        })
+                .ifNoItem().after(replyTimeout)
+                .failWith(() -> new AmqpRequestReplyTimeoutException(correlationId))
+                .onItem().transformToUniAndConcatenate(m -> {
+                    if (replyFailureHandler != null) {
+                        Throwable failure = replyFailureHandler.handleReply(
+                                (AmqpMessage<?>) m);
```
cescoffier, line 150: Should we check for type first?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java
```diff
@@ -181,11 +152,43 @@ static AmqpClientOptions getOptionsFromChannel(AmqpConnectorCommonConfiguration
...
+        if (options.getContainerId() == null) {
+            options.setContainerId("vert.x-" + UUID.randomUUID());
+        } else {
+            options.setMetricsName("amqp|" + options.getContainerId());
+        }
+        return ConfigUtils.customize(config.config(), connector.getConfigCustomizers(), options);
+    }
+
+    public static String computeConnectionFingerprint(AmqpClientOptions options) {
+        return sha256(options.toJson().encode());
```
cescoffier, line 177: It's a bit fragile:

- JSON object field ordering may not be stable across JVM versions or Vert.x versions
- Floating-point representation in JSON can vary
- It depends on Vert.x's internal JSON serialization (which can change, use databind or not...)

cescoffier: It might be better to compute something like:

return sha256(
        options.getHost() + "|" + 
        options.getPort() + "|" + 
        options.getUsername() + "|" +
        // ... other relevant fields
    );

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java
```diff
@@ -369,6 +370,8 @@ private String getActualAddress(Message<?> message, io.vertx.mutiny.amqp.AmqpMes
                     }
                     return Optional.ofNullable(addressFromMessage);
                 })
+                .or(() -> message.getMetadata(IncomingAmqpMetadata.class)
```
cescoffier, line 373: `getReplyTo()` could return null, which would be passed to `orElse(configuredAddress)`.  Would that work?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessageConverter.java
```diff
@@ -69,7 +69,13 @@ static io.vertx.mutiny.amqp.AmqpMessage convertToAmqpMessage(Message<?> message,
         output.setAddress(metadata.getAddress());
         output.setSubject(metadata.getSubject());
         output.setReplyTo(metadata.getReplyTo());
-        output.setCorrelationId(metadata.getCorrelationId());
+        if (metadata.getCorrelationId() != null) {
+            output.setCorrelationId(metadata.getCorrelationId());
+        } else {
+            message.getMetadata(IncomingAmqpMetadata.class)
+                    .filter(incoming -> incoming.getReplyTo() != null)
```
cescoffier, line 76: This auto-copies the incoming message ID as correlation ID when replying, which is correct for request-reply but is a behavior change for regular send operations. This logic should only apply when the message is part of a reply flow.

Does the filter validate this?

smallrye-reactive-messaging-amqp/src/test/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyEmitterTest.java
```diff
@@ -0,0 +1,492 @@
...
+
+    private void startSlowReplyServer() {
+        usage.client.createReceiver("requests")
+                .onItem().transformToMulti(AmqpReceiver::toMulti)
+                .subscribe().with(request -> {
+                    String replyTo = request.replyTo();
+                    String correlationId = request.id();
+                    if (replyTo != null && correlationId != null) {
+                        try {
+                            Thread.sleep(3000);
```
cescoffier: Aren't we on the event loop here?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
```diff
@@ -0,0 +1,236 @@
...
+            } else {
+                log.requestReplyMessageIgnored(channel, correlationId.toString());
+            }
+        }
+        subscription.get().request(1);
+    }
+
+    @Override
+    public void onFailure(Throwable failure) {
+        log.requestReplyConsumerFailure(channel, failure);
```
cescoffier, line 199: Consumer failures are logged but not propagated to pending requests. Pending requests will timeout instead of failing fast with the actual error.

You probably want something like:

log.requestReplyConsumerFailure(channel, failure);
    for (Map.Entry<CorrelationId, PendingReplyImpl<Rep>> entry : pendingReplies.entrySet()) {
        entry.getValue().emitter.fail(failure);
    }
    pendingReplies.clear();

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHolder.java
```diff
@@ -0,0 +1,38 @@
...
+    public AmqpClient client() {
+        return client;
+    }
+
+    public AmqpClientHolder retain(String channel) {
+        channels.add(channel);
+        return this;
+    }
+
+    public synchronized ConnectionHolder getOrCreateConnectionHolder(
```
cescoffier: We need to describe how the thread safety works. 
What is thread safe, what is not.

cescoffier: Why `getOrCreateConnectionHolder` is synchronized vs using CAS on an AtomicReference?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/i18n/AMQPExceptions.java
```diff
@@ -32,4 +32,6 @@ public interface AMQPExceptions {
     @Message(id = 16005, value = "Only one subscriber allowed")
     IllegalStateException illegalStateOnlyOneSubscriberAllowed();
 
+    @Message(id = 16006, value = "Container ID '%s' does not match the configuration")
```
cescoffier: What about "Container ID '%s' is already in use with different connection settings. Channels sharing a container-id must have identical connection configuration (host, port, credentials, etc.)"

cescoffier (changes requested):
I think it would also be good:

- to expose pending reply count as a metric
- validation for configuration at startup 

   ```
Config connectorConfig = Configs.outgoing(config.get(), AmqpConnector.CONNECTOR_NAME, configuration.name());
    String replyAddress = connectorConfig.getOptionalValue(REPLY_ADDRESS_KEY, String.class).orElse(null);
    if (replyAddress == null) {
        log.warn("No reply address configured for request-reply channel: " + configuration.name());
    }
```

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java (hunk 1 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java:157]: It is called both on completion and cancellation.

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java (hunk 9 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java:199]: Makes sense

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java (hunk 5 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpClientHelper.java:177]: At the same time, that fingerprint stays in memory, and all we care about is that it is consistent inside a runtime.

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessageConverter.java (hunk 7 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpMessageConverter.java:76]: This is a relatively safe assumption we do, when we add request-reply to the connector:
When sending a message as part of a processing (incoming metadata is present), if the correlation ID is not already set, and replyTo is present in the incoming message, we set it automatically.

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java (hunk 3 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java:174]: This is not a bean, the complete will be called when the outgoing channel is being closed.

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java (hunk 6 above)
Comment (ozangunalp) [reply smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java:373]: Yes, if the replyTo address in the incoming message is null, we fall back to the configured address.

File: smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java [hunk 12]
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Override
+    public Multi<Message<Rep>> requestMulti(Message<Req> request) {
+        var builder = request.getMetadata(OutgoingAmqpMetadata.class)
+                .map(OutgoingAmqpMetadata::from)
+                .orElseGet(OutgoingAmqpMetadata::builder);
+        CorrelationId correlationId = correlationIdHandler.generate(request);
+        // if using direct-reply-to reply-to address is assigned by broker when the receiver attaches
+        String replyTo = DIRECT_REPLY_TO_ADDRESS.equals(replyToAddress) ? connector.getIncomingAddress(channel)
```
cescoffier, line 127: Should we throw something if we don't get a reply-to address?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/BytesCorrelationIdHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import org.eclipse.microprofile.reactive.messaging.Message;
+
+import io.smallrye.common.annotation.Identifier;
+
+@ApplicationScoped
+@Identifier("bytes")
+public class BytesCorrelationIdHandler implements CorrelationIdHandler {
+
+    @Inject
+    @ConfigProperty(name = "smallrye.amqp.request-reply.correlation-id.bytes.length", defaultValue = "12")
```
cescoffier, line 18: Maybe mention this is an example.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
```diff
@@ -0,0 +1,239 @@
...
+        log.requestReplyConsumerFailure(channel, failure);
+        for (Map.Entry<CorrelationId, PendingReplyImpl<Rep>> replyEntry : pendingReplies.entrySet()) {
+            replyEntry.getValue().emitter.fail(failure);
+        }
+        pendingReplies.clear();
+    }
+
+    @Override
+    public void onCompletion() {
+
```
cescoffier: Should we clear any pending state? Just is case...

cescoffier (dismissed):

cescoffier:
Make this feature experimental, until we have tracing/metrics and so on.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
ozangunalp: Yes we should

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpRequestReplyImpl.java
ozangunalp: I realize this was there to support `amq.direct-reply.to` addresses, which is supported on RabbitMQ with link-pairing. We've reverted that until vertx-amqp-client has the basis to support that: https://github.com/vert-x3/vertx-amqp-client/pull/91.

I suggest we keep it here. I've added a note on documentation on limitations.

ozangunalp:
> Make this feature experimental, until we have tracing/metrics and so on.

Yes, it is marked as experimental in the docs

ozangunalp:
@cescoffier Thanks for the review! I think we can get this in.

cescoffier (approved):

PR #3383, opened 2026-04-27
Fix RabbitMQ incoming channel race condition during startup

gastaldi:
The incoming channel's connection and infrastructure setup (exchange, queue, bindings, consumer) was fully asynchronous, completing after the container reported "started". 

This caused race conditions where producers could publish to exchanges before they existed.

Two fixes:
- `ClientHolder`: move `hasBeenConnected/log` from `onSubscription()` to `onItem()` so they fire after the connection is actually established
- `IncomingRabbitMQChannel`: eagerly await the connection and consumer creation during construction, ensuring all infrastructure is set up before `getPublisher()` returns

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
@@ -30,7 +30,7 @@ public class ClientHolder {
     public ClientHolder(RabbitMQClient client) {
         this.client = client;
         this.connection = Uni.createFrom().deferred(() -> client.start()
-                .onSubscription().invoke(() -> {
+                .onItem().invoke(ignored -> {
                     hasBeenConnected.set(true);
                     log.connectionEstablished(String.join(", ", channels));
                 })
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
@@ -30,7 +30,7 @@ public class ClientHolder {
     public ClientHolder(RabbitMQClient client) {
         this.client = client;
         this.connection = Uni.createFrom().deferred(() -> client.start()
-                .onItem().invoke(ignored -> {
+                .onSubscription().invoke(() -> {
                     hasBeenConnected.set(true);
                     log.connectionEstablished(String.join(", ", channels));
                 })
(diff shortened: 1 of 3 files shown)


gastaldi:
This fixes the quarkus-quickstart failures: https://github.com/quarkusio/quarkus-quickstarts/actions/runs/24974509581

ozangunalp:
@gastaldi thanks for looking at this!

I am not sure if we need to memoize the created consumer. When there is a disconnection I think the created consumer will be stuck on the previous connection.

ozangunalp:
This is probably due to #3354

ozangunalp:
Hence, RabbitMQReconnectionTest is failing. I'll check the failing quickstart to reproduce the problem

ozangunalp:
@gastaldi also did you see this one : https://github.com/quarkusio/quarkus-quickstarts/pull/1638 ?

gastaldi:
@ozangunalp yeah, that fixes the race problem for that particular quickstart, but doesn't solve the issue in general, which is what this PR is trying to fix. 

Anyway, feel free to close this if it doesn't make sense :)

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
```diff
@@ -65,11 +66,12 @@ public IncomingRabbitMQChannel(RabbitMQConnector connector,
         final RabbitMQFailureHandler onNack = createFailureHandler(connector.failureHandlerFactories(), ic);
         final RabbitMQAckHandler onAck = createAckHandler(ic);
 
-        Multi<? extends Message<?>> multi = createConsumer(connector, ic)
-                .invoke(tuple -> client = tuple.getItem1().client())
-                // Translate all consumers into a merged stream of messages
-                .onItem().transformToMulti(
-                        tuple -> getStreamOfMessages(tuple.getItem2(), tuple.getItem1(), incomingContext, ic, onNack, onAck));
+        Tuple2<ClientHolder, RabbitMQConsumer> consumer = createConsumer(connector, ic)
```
cescoffier: I don't believe this is correct.
It resolves the consumer exactly once during construction. But, after a disconnect, the consumer is stale and pinned to the old connection.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
```diff
@@ -65,11 +66,12 @@ public IncomingRabbitMQChannel(RabbitMQConnector connector,
...
+                .await().atMost(Duration.ofMillis(ic.getConnectionTimeout()));
```
cescoffier: This should be async.

cescoffier: In addition, with this change, shared connections with multiple incoming channels are now serialized.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolder.java
```diff
@@ -30,7 +30,7 @@ public class ClientHolder {
     public ClientHolder(RabbitMQClient client) {
         this.client = client;
         this.connection = Uni.createFrom().deferred(() -> client.start()
-                .onSubscription().invoke(() -> {
+                .onItem().invoke(ignored -> {
```
cescoffier: That introduces a race condition. If the Uni<Void> from client.start() completes asynchronously, there's a window where subscription has happened but  hasBeenConnected is still false. This could cause health checks to briefly report unhealthy even though connection is in progress.

cescoffier (changes requested):

ozangunalp:
Closing this in favour of #3393

PR #3393, opened 2026-05-13
RabbitMQ connection flow cleanup and lazy-client=false option

ozangunalp:
`lazy-client=false` for eager client/queue-exchange creation

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
@@ -18,7 +18,6 @@ import com.rabbitmq.client.AMQP;
 import io.opentelemetry.api.OpenTelemetry;
 import io.smallrye.mutiny.Multi;
 import io.smallrye.mutiny.Uni;
-import io.smallrye.mutiny.tuples.Tuple2;
 import io.smallrye.reactive.messaging.health.HealthReport;
 import io.smallrye.reactive.messaging.providers.helpers.CDIUtils;
 import io.smallrye.reactive.messaging.providers.helpers.VertxContext;
@@ -65,11 +64,22 @@ public class IncomingRabbitMQChannel {
         final RabbitMQFailureHandler onNack = createFailureHandler(connector.failureHandlerFactories(), ic);
         final RabbitMQAckHandler onAck = createAckHandler(ic);
 
-        Multi<? extends Message<?>> multi = createConsumer(connector, ic)
-                .invoke(tuple -> client = tuple.getItem1().client())
-                // Translate all consumers into a merged stream of messages
+        ClientHolder holder = connector.getClientHolder(ic);
+        this.client = holder.client();
+        registerInfrastructureCallback(holder.client(), ic, connector);
+
+        if (!ic.getLazyClient()) {
+            Uni<Void> connection = holder.getOrEstablishConnection().replaceWithVoid();
+            if (ic.getSharedConnectionName().isPresent()) {
+                connection = connection.call(() -> declareInfrastructure(holder.client(), ic, connector));
+            }
+            connection.await().indefinitely();
+        }
+
+        Multi<? extends Message<?>> multi = holder.getOrEstablishConnection()
+                .flatMap(connection -> createConsumer(ic, connection))
                 .onItem().transformToMulti(
-                        tuple -> getStreamOfMessages(tuple.getItem2(), tuple.getItem1(), incomingContext, ic, onNack, onAck));
+                        consumer -> getStreamOfMessages(consumer, holder, incomingContext, ic, onNack, onAck));
 
         if (ic.getBroadcast()) {
             multi = multi.broadcast().toAllSubscribers();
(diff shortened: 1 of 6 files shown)


smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
```diff
@@ -65,11 +64,22 @@ public IncomingRabbitMQChannel(RabbitMQConnector connector,
...
+        ClientHolder holder = connector.getClientHolder(ic);
+        this.client = holder.client();
+        registerInfrastructureCallback(holder.client(), ic, connector);
+
+        if (!ic.getLazyClient()) {
+            Uni<Void> connection = holder.getOrEstablishConnection().replaceWithVoid();
+            if (ic.getSharedConnectionName().isPresent()) {
+                connection = connection.call(() -> declareInfrastructure(holder.client(), ic, connector));
+            }
+            connection.await().indefinitely();
```
cescoffier: Aren't we on the event loop? Even if we are on a worker thread, I would add a timeout here.

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/OutgoingRabbitMQChannel.java
```diff
@@ -34,13 +34,17 @@ public OutgoingRabbitMQChannel(RabbitMQConnector connector, RabbitMQConnectorOut
...
-        client.getDelegate().addConnectionEstablishedCallback(promise -> {
-            // Ensure we create the exchange to which messages are to be sent
-            RabbitMQClientHelper.declareExchangeIfNeeded(client, oc, connector.configMaps())
-                    .subscribe().with((ignored) -> promise.complete(), promise::fail);
-        });
+        registerExchangeCallback(client, oc, connector);
+
+        if (!oc.getLazyClient()) {
+            Uni<Void> connection = holder.getOrEstablishConnection().replaceWithVoid();
+            if (oc.getSharedConnectionName().isPresent()) {
```
cescoffier: Is this conditioning the lazy-client?

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/OutgoingRabbitMQChannel.java
```diff
@@ -34,13 +34,17 @@ public OutgoingRabbitMQChannel(RabbitMQConnector connector, RabbitMQConnectorOut
...
+                connection = connection
+                        .call(() -> RabbitMQClientHelper.declareExchangeIfNeeded(client, oc, connector.configMaps()));
+            }
+            connection.await().indefinitely();
```
cescoffier: Same here - we need to make sure we can block, and even if we can, we should add a timeout.

smallrye-reactive-messaging-rabbitmq/src/test/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQTest.java
```diff
@@ -199,6 +199,36 @@ void testIncomingDeclarations() throws Exception {
         assertThat(binding2.getString("routing_key")).isEqualTo("urgent");
     }
 
+    /**
+     * Verifies that with eager client (default), incoming channel infrastructure
```
cescoffier: Is it the default? lazy-client is set to true by default.

smallrye-reactive-messaging-rabbitmq/src/test/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolderTest.java
```diff
@@ -0,0 +1,146 @@
...
+        io.vertx.mutiny.rabbitmq.RabbitMQClient reconnected = holder.getOrEstablishConnection()
+                .await().atMost(Duration.ofSeconds(10));
+        assertThat(reconnected).isSameAs(first);
+        assertThat(client.isConnected()).isTrue();
+    }
+
+    @Test
+    public void testConcurrentGetOrEstablishConnectionAfterDisconnect() throws InterruptedException {
+        Vertx vertx = executionHolder.vertx();
+        io.vertx.mutiny.rabbitmq.RabbitMQClient client = io.vertx.mutiny.rabbitmq.RabbitMQClient
```
cescoffier: Any reason to not import it?

cescoffier (changes requested):

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
ozangunalp: No, channels are created at wiring time. But yes, let's put a timeout as connection timeout

smallrye-reactive-messaging-rabbitmq/src/test/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQTest.java
ozangunalp: Yes, we keep the previous lazy behaviour of connecting at subscription time, the comment was wrong

smallrye-reactive-messaging-rabbitmq/src/test/java/io/smallrye/reactive/messaging/rabbitmq/ClientHolderTest.java
ozangunalp: I forgot to clean it up

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/OutgoingRabbitMQChannel.java
ozangunalp: Same as the incoming

smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/OutgoingRabbitMQChannel.java
ozangunalp: That's a nice catch. We should be able to get that condition out of the eager client

cescoffier (approved):

PR #3400, opened 2026-05-27
Fix ThrowingEmitter race condition between hasRequests and emit

ozangunalp:


cescoffier (commented):
Classic TOCTOU race - nice! 

Don't you think we can test this?
1. Create a `ThrowingEmitter` with `bufferSize=1`
2. Subscribe with a request of 1
3. Assert `requested()` returns 1 (it would have returned the delegate's value before your fix)
4. Call `emit()` once
5.  Assert `requested()` returns 0

ozangunalp:
Indeed, let me check to add that test

ozangunalp:
Failing GCP test could be a red flag.

cescoffier (approved):

cescoffier:
Arf yes, the CI is unhappy.

PR #3402, opened 2026-05-28
fix ack/nack completing on failure

bojanv55:
Hi @ozangunalp. Can you please check the code and below description. This seems to be legit problem for nacking case.

- IncomingKafkaRecord.nack(...) delegates to the configured KafkaFailureHandler.                                                                                                                                                                                    
  - KafkaFailStop.handle(...) reports the failure, then returns:                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                      
  return Uni.createFrom().<Void> failure(reason)                                                                                                                                                                                                                      
          .emitOn(record::runOnMessageContext);                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                      
  So a nacked record using failure-strategy=fail will produce a failed completion stage. With the old thenRun(postProcessing), that means ordered release was skipped.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedIncomingKafkaRecord.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedIncomingKafkaRecord.java
@@ -1,5 +1,7 @@
 package io.smallrye.reactive.messaging.kafka.impl;
 
+import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
+
 import java.util.concurrent.CompletionStage;
 
 import org.eclipse.microprofile.reactive.messaging.Metadata;
@@ -21,12 +23,20 @@ public class OrderedIncomingKafkaRecord<K, T> extends IncomingKafkaRecord<K, T> 
 
     @Override
     public CompletionStage<Void> ack(Metadata metadata) {
-        return super.ack(metadata).thenRun(postProcessing);
+        return super.ack(metadata).whenComplete((ignored, failure) -> runPostProcessing());
     }
 
     @Override
     public CompletionStage<Void> nack(Throwable reason, Metadata metadata) {
-        return super.nack(reason, metadata).thenRun(postProcessing);
+        return super.nack(reason, metadata).whenComplete((ignored, failure) -> runPostProcessing());
+    }
+
+    private void runPostProcessing() {
+        try {
+            postProcessing.run();
+        } catch (Exception e) {
+            log.debugf(e, "An exception was thrown while completing ordered Kafka record post-processing");
+        }
     }
 
 }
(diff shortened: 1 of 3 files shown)


ozangunalp:
@bojanv55 Thanks for this. Looks like a bug indeed. 

Can we have another test that checks the fail without release case on a kafka broker ?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/impl/OrderedIncomingKafkaRecordTest.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        };
+        return new IncomingKafkaRecord<>(
+                new ConsumerRecord<>("ordered-topic", 0, 0, "key", "value"),
+                "channel",
+                0,
+                commitHandler,
+                failureHandler,
+                false);
+    }
```
cescoffier, line 130: I don't know how to test this, but it would be great to test a `ack` failure, for example, if the consumer is closed, or a rebalance in progress.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedIncomingKafkaRecord.java
```diff
@@ -21,12 +21,12 @@ public OrderedIncomingKafkaRecord(IncomingKafkaRecord<K, T> delegate, Runnable p
...
     @Override
     public CompletionStage<Void> ack(Metadata metadata) {
-        return super.ack(metadata).thenRun(postProcessing);
+        return super.ack(metadata).whenComplete((ignored, failure) -> postProcessing.run());
     }
 
     @Override
     public CompletionStage<Void> nack(Throwable reason, Metadata metadata) {
-        return super.nack(reason, metadata).thenRun(postProcessing);
+        return super.nack(reason, metadata).whenComplete((ignored, failure) -> postProcessing.run());
```
cescoffier: If postProcessing throws an exception, this exception will _suppress_ `failure` (and I'm not sure we should add it to the list of suppressed exceptions). So, I believe we are losing the original reason for the failure.

I guess something like this would be more correct:
```
return super.nack(reason, metadata).whenComplete((ignored, failure) -> {
    try {
        postProcessing.run();
    } catch (Exception e) {
        // log but don't suppress the original failure, or add it to the suppress list
    }
});
```

cescoffier (changes requested):

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/impl/OrderedIncomingKafkaRecordTest.java
bojanv55: I guess `shouldRunPostProcessingWhenAckFails` should be enough for this part?

cescoffier: I only see `shouldRunPostProcessingWhen**N**ackFails`.

bojanv55: ack line 21 and nack line 36 ?

cescoffier (approved):
Please squash your commits.

ozangunalp (approved):

PR #3403, opened 2026-05-28
race that turns a partition rebalance into permanent merge-slot leaks

bojanv55:
@ozangunalp One more bug for you to check. Should also be legit case.

Test is doing following:                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                      
  1. Start ordered-by-key processing with ordered.max-concurrency=1.                                                                                                                                                                                                  
  2. Emit one record for key-0, so the first ordered group occupies the only merge slot.                                                                                                                                                                              
  3. Emit a second record for key-1; its group is created internally but not yet subscribed because max concurrency is full.                                                                                                                                          
  4. Revoke the partition before that buffered key-1 group is subscribed.                                                                                                                                                                                             
  5. Complete the source.                                                                                                                                                                                                                                             
  6. Assert the whole stream completes instead of wedging forever.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
@@ -1,6 +1,8 @@
 package io.smallrye.reactive.messaging.kafka.impl;
 
 import static io.smallrye.mutiny.helpers.Subscriptions.CANCELLED;
+import static io.smallrye.mutiny.helpers.Subscriptions.complete;
+import static io.smallrye.reactive.messaging.kafka.i18n.KafkaLogging.log;
 
 import java.util.Collection;
 import java.util.Map;
@@ -55,7 +57,7 @@ public class OrderedStreamHandler extends ContextHolder implements KafkaCommitHa
     private final int maxPollRecords;
     private final int maxQueueSizeFactor;
     private final int maxConcurrency;
-    private Map<TopicPartitionKey, OrderedGroup> orderedByGroups;
+    Map<TopicPartitionKey, OrderedGroup> orderedByGroups;
 
     public OrderedStreamHandler(KafkaConnectorIncomingConfiguration configuration,
             Vertx vertx,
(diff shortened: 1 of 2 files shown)


ozangunalp:
Yeah, looking at the code it makes very much sense.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
```diff
@@ -235,7 +236,10 @@ class SelfPurgingPauserGroupMultiProcessor
 
             @Override
             public void onSubscribe(Flow.Subscription subscription) {
-                if (compareAndSetUpstreamSubscription(null, subscription) && orderedGroup != null) {
+                if (orderedGroup == null) {
+                    subscription.cancel();
```
cescoffier, line 241: Maybe add:
    log.debugf("Group for key %s already revoked before subscription, completing downstream", key);

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandlerTest.java
```diff
@@ -0,0 +1,128 @@
...
+                0,
+                new PassThroughCommitHandler(),
+                new IgnoredFailureHandler(),
+                false);
+    }
+
+    @SuppressWarnings("unchecked")
+    private Map<TopicPartitionKey, OrderedStreamHandler.OrderedGroup> orderedGroups(OrderedStreamHandler handler)
+            throws Exception {
+        Field field = OrderedStreamHandler.class.getDeclaredField("orderedByGroups");
```
cescoffier: Would it be better to make it package private?

smallrye-reactive-messaging-kafka/src/test/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandlerTest.java
```diff
@@ -0,0 +1,128 @@
...
+                    .containsKey(TopicPartitionKey.ofKey(TOPIC, 0, "key-1")));
+
+            handler.partitionsRevoked(List.of(new TopicPartition(TOPIC, 0)));
+            emitterRef.get().complete();
+
+            subscriber.awaitCompletion(Duration.ofSeconds(1));
+            assertThat(subscriber.getItems())
+                    .extracting(IncomingKafkaRecord::getKey)
+                    .containsExactly("key-0");
+        } finally {
```
cescoffier, line 68: should we also "terminate" the handler?

cescoffier (dismissed):
Just some small comments.

ozangunalp (approved):

ozangunalp:
failing pulsar and GCP tests are handled in #3404 and #3394

PR #3415, opened 2026-06-08
AMQP 1.0 Claim-based Security Implementation

ozangunalp:
Implements AMQP 1.0 CBS : https://docs.oasis-open.org/amqp/amqp-cbs/v1.0/amqp-cbs-v1.0.html

Both as it is described in the 1.0 version of the spec, using the`set-token` action, built on send message with confirm,
But also as described on an older version of the spec, using the `put-token` action, built on request-reply.
This second version is the auth mechanism supported by Azure Service Bus :
https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-protocol-guide

The connector provides implementations for the token manager, token exchange, both set and put, and token refresh. It requires an implementation of `CbsTokenProvider` as CDI bean to provide a `CbsToken` when authenticating.


smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsPutTokenExchange.java
```diff
@@ -0,0 +1,136 @@
...
+    private final AmqpDirectRequestReply requestReply;
+
+    @ApplicationScoped
+    @Identifier("put-token")
+    public static class Factory implements CbsExchange.Factory {
+
+        @Override
+        public CbsExchange create(AmqpConnection connection, AmqpConnectorCommonConfiguration config) {
+            String entity = config.getAddress().orElseGet(config::getChannel);
+            String audience = config.config().getOptionalValue("cbs-audience", String.class)
```
cescoffier: you use a - here, while the connector use a .

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/RefreshingCbsTokenManager.java
```diff
@@ -0,0 +1,71 @@
...
+    public Uni<CbsToken> authorize(CbsExchange exchange) {
+        if (timerId != 0) {
+            vertx.cancelTimer(timerId);
+        }
+        return authorizeWithConnection(exchange);
+    }
+
+    Uni<CbsToken> authorizeWithConnection(CbsExchange exchange) {
+        return tokenManager.authorize(exchange)
+                .invoke(t -> refreshToken(exchange, MILLIS.between(OffsetDateTime.now(), t.refreshAt())));
```
cescoffier: Be careful here, it can be negative if refreshAt is in the past.  I don't know what will happen with setTimer is you pass a negative time. I would say that if negative, refresh immediately.

cescoffier: Or add a default delay (like 1s)

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsSetTokenExchange.java
```diff
@@ -0,0 +1,51 @@
...
+                        .applicationProperties(JsonObject.of("token-type", token.type()))
+                        .withBody(token.token())
+                        .build())
+                        .replaceWith(token)));
+    }
+
+    @Override
+    public void close() {
+        closed = true;
+        senderUni.subscribe().with(AmqpSender::closeAndForget);
```
cescoffier: If the sender was never created (because close is called too early, of if closed is called twice.)

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsToken.java
```diff
@@ -0,0 +1,31 @@
...
+public interface CbsToken {
+
+    /**
+     * The token value
+     */
+    String token();
+
+    /**
+     * The type of the token
+     * "servicebus.windows.net:sastoken" or "jwt
```
cescoffier:
```suggestion
     * "servicebus.windows.net:sastoken" or "jwt"
```

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java
```diff
@@ -294,11 +294,14 @@ private Uni<Tuple3<Message<?>, AmqpSender, Long>> send(Message<?> msg, boolean d
                 // We are on Vert.x context that created the client, we can access the remaining credits and update it.
                 .replaceWith(() -> Tuple3.<Message<?>, AmqpSender, Long> of(msg, s, s.remainingCredits())))
                 .onFailure().retry().withBackOff(ofSeconds(1), ofSeconds(retryInterval)).atMost(retryAttempts)
-                .onItemOrFailure().call((s, failure) -> {
+                .onItemOrFailure().transformToUni((s, failure) -> {
```
cescoffier, line 297: This is a breaking change as now we are returning `null` on failure. Did you add a test for this?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpDirectRequestReply.java
```diff
@@ -0,0 +1,130 @@
...
+                    String correlationId = message.correlationId();
+                    if (correlationId != null) {
+                        UniEmitter<AmqpMessage> emitter = pendingRequests.remove(correlationId);
+                        if (emitter != null) {
+                            emitter.complete(message);
+                        } else {
+                            log.requestReplyMessageIgnored("direct-reply", correlationId);
+                        }
+                    }
+                }, this::failAllPending);
```
cescoffier: should we also fail on cancellation?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Optional;
+
+import org.apache.qpid.proton.amqp.Symbol;
+
+import io.vertx.mutiny.amqp.AmqpConnection;
+import io.vertx.proton.ProtonConnection;
+
+public class CbsUtils {
+
+    // TODO handle default node name in a more flexible way, possibly through configuration
```
cescoffier, line 12: Should it be handled?

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpConnector.java
```diff
@@ -61,12 +63,19 @@
...
 @ConnectorAttribute(name = "selector", direction = INCOMING, description = "Sets a message selector. This attribute is used to define an `apache.org:selector-filter:string` filter on the source terminus, using SQL-based syntax to request the server filters which messages are delivered to the receiver (if supported by the server in question). Precise functionality supported and syntax needed can vary depending on the server.", type = "string")
 @ConnectorAttribute(name = "qos", direction = INCOMING, description = "Sets the local QOS config, values can be {@code null}, {@code AT_MOST_ONCE} or {@code AT_LEAST_ONCE}.", type = "string")
+@ConnectorAttribute(name = "initial-credits", direction = INCOMING, description = "The number of credits initially requested for incoming messages. This controls how many messages the receiver can buffer.", type = "int", defaultValue = "1000")
```
cescoffier, line 78: Is this related?

cescoffier (changes requested):

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/RefreshingCbsTokenManager.java
ozangunalp: It makes sense to go with the default delay

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/AmqpCreditBasedSender.java
ozangunalp: It is handled by the subscriber to this Uni, and handled by the same error handling as serialization failure. So send nack'ed by the broker doesn't fail the connector but nacks the message.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsUtils.java
ozangunalp: Maybe in the future. `$cbs` is in the spec anyway.

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/reply/AmqpDirectRequestReply.java
ozangunalp: Done

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsSetTokenExchange.java
ozangunalp: Done. The exchange is created per connection so we handle the lazy init via the uni and the closing via the actual reference

smallrye-reactive-messaging-amqp/src/main/java/io/smallrye/reactive/messaging/amqp/cbs/CbsPutTokenExchange.java
ozangunalp: Changed to `cbs.audience`.

cescoffier (approved):

PR #3439, opened 2026-06-11
Upgrade OTel SDK to 1.62.0

brunobat:
This is required by https://github.com/quarkusio/quarkus/issues/54670


brunobat:
Not sure if the test error is related @ozangunalp

ozangunalp:
Don't think so. I'll rerun and check it.

ozangunalp (approved):

PR #3440, opened 2026-06-11
Support GenericPayload in Emitters

ozangunalp:


smallrye-reactive-messaging-provider/src/test/java/io/smallrye/reactive/messaging/GenericPayloadTest.java
```diff
@@ -188,10 +188,9 @@ public Object metadata() {
     }
 
     @Test
-    public void testEmitterGenericPayloadNotSupported() {
+    public void testEmitterGenericPayload() {
```
cescoffier, line 191: Should we also test: `emitter.send(Message.of(GenericPayload.of("test", Metadata.of(myMeta))))`

documentation/src/main/docs/concepts/generic-payloads.md
```diff
@@ -26,9 +26,15 @@ You can combine generic payloads with [metadata injection](incoming-metadata-inj
 
 Note that the metadata provided with the outgoing generic payload is merged with the incoming message metadata.
 
+`GenericPayload` can also be used with emitters to attach metadata without dealing with `Message` directly:
+
+``` java
+{{ insert('genericpayload/GenericPayloadExample.java', 'emitter') }}
+```
+
+The inner payload is extracted and the metadata from the `GenericPayload` is merged into the outgoing message.
```
cescoffier, line 35: It might be worth clarifying that metadata from the `GenericPayload` does not overwrite existing metadata entries.

cescoffier (dismissed):
I made 2 small comments.

PR #3453, opened 2026-06-22
Kafka handle position and commit offset on seek operations

ozangunalp:
- Latest commit strategy: clear offsets on partition revocation
- Added clearBuffer to the PausableChannel interface
- Added `onPartitionsSeeked` callback to `KafkaConsumerRebalanceListener` and `partitionsSeeked` to KafkaCommitHandler interface. On seek operations, the Kafka consumer clears the poller buffer and calls `partitionsSeeked` on commit handlers.

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
@@ -56,6 +56,7 @@ import io.vertx.mutiny.core.Vertx;
 public class KafkaThrottledLatestProcessedCommit extends ContextHolder implements KafkaCommitHandler {
 
     private final Map<TopicPartition, OffsetStore> offsetStores = new HashMap<>();
+    private final Set<TopicPartition> seekedPartitions = new HashSet<>();
 
     private final String groupId;
     private final KafkaConsumer<?, ?> consumer;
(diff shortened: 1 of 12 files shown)


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
```diff
@@ -154,6 +154,11 @@ public void partitionsRevoked(Collection<TopicPartition> partitions) {
         delegate.partitionsRevoked(partitions);
     }
 
+    @Override
+    public void partitionsSeeked(Collection<TopicPartition> partitions) {
+        delegate.partitionsSeeked(partitions);
```
cescoffier, line 159: After a seek, the ordered groups for the affected partitions are holding a stale state (pending counts, buffered items). This could cause records from the new position to be incorrectly ordered or dropped if the group's `pending` counter is non-zero. 

I think we should consider whether the ordered groups for sought partitions should be canceled and removed, similar to what `partitionsRevoked` does.

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
```diff
@@ -56,6 +56,7 @@
 public class KafkaThrottledLatestProcessedCommit extends ContextHolder implements KafkaCommitHandler {
 
     private final Map<TopicPartition, OffsetStore> offsetStores = new HashMap<>();
+    private final Set<TopicPartition> seekedPartitions = new HashSet<>();
```
cescoffier, line 59: Use a copy on write variant - HashSet is not thread-safe.

cescoffier (changes requested):
Should KafkaCheckpointCommit also implement partitionsSeeked?

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java
ozangunalp: Like the offset stores, it is only accessed on the channel context

cescoffier (approved):

PR #3456, opened 2026-06-22
Reset commit handler offset state on seek operations

### LINKED ISSUE #3455: kafka consumer seekToBeginning gap in committed offsets
**Problem description**

When performing a `seekToBeginning` operation on a consumer, there is a gap in committed offsets until the previously largest offset is reached.

This is caused by the internal bookkeeping of the offset (see line 73 of KafkaLatestCommit): https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaLatestCommit.java#L73

Example for illustration:
1. Consumer commits offset 100 (stored internally).
2. `seekToBeginning` is triggered. Consumer resumes reading from offset 1.
3. Consumer stops committing offsets entirely due to the internally stored offset.
4. Only after reaching offset 100 again does the consumer resume committing.

This occurs when the `latest` or `throttled` `commit-strategy` is configured.

gastaldi:
## Summary

Fixes #3455

- When `seekToBeginning`/`seekToEnd`/`seek` is called on a consumer, the `latest` and `throttled` commit handlers retained stale offset state, causing commits to be silently skipped until the old high watermark was re-reached.
- Adds a `partitionsSeeked` callback to `KafkaCommitHandler` (default no-op for backward compatibility) and notifies it from `ReactiveKafkaConsumer` after seek operations, so cached offset tracking is reset.

## Test plan

- [x] New test `testLatestCommitStrategyAfterSeekToBeginning` — verifies commits resume after seekToBeginning with `latest` strategy
- [x] New test `testThrottledStrategyAfterSeekToBeginning` — verifies commits resume after seekToBeginning with `throttled` strategy
- [x] Full `CommitStrategiesTest` suite passes (9 tests, 0 failures)

### DIFF
--- a/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaCommitHandler.java
+++ b/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaCommitHandler.java
@@ -95,6 +95,17 @@ public interface KafkaCommitHandler {
     }
 
     /**
+     * Called when the consumer position for partitions has been changed by a seek operation.
+     * Implementations should reset any cached offset state for the given partitions so that
+     * commits resume correctly from the new position.
+     *
+     * @param partitions the partitions for which the consumer position was changed
+     */
+    default void partitionsSeeked(Collection<TopicPartition> partitions) {
+        // Do nothing by default.
+    }
+
+    /**
      * Handle message acknowledgment
      *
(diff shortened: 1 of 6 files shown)


ozangunalp:
Wait is this an effective duplicate of the PR #3453 ?

gastaldi:
> Wait is this an effective duplicate of the PR #3453 ?

Yeah, looks like it. I closed to avoid confusion

PR #3462, opened 2026-06-28
Fix using tracing.attribute-headers in RabbitMQTraceAttributesExtractor

### LINKED ISSUE #3367: RabbitMQ tracing.attribute-headers not set in Otel Spans
Maybe I am mistaken, but it seems the `mp.messaging.[incoming/outgoing].*.tracing.attribute-headers` does not work anymore?
I tried setting it in an example application, the custom header is also present in the internally created `RabbitMQTrace` but its not set in the OpenTelemetry span.

In the codebase I also can't find any usage of `RabbitMQConnectorCommonConfiguration.getTracingAttributeHeaders()` anywhere.
It was used in [First commit of rabbitmq connector (b1e6ad8)](https://github.com/smallrye/smallrye-reactive-messaging/blob/b1e6ad84cc8672d895c2721e721123a2a0d205fc/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/RabbitMQMessageSender.java#L250) but all usages have later been removed by [Use OpenTelemetry Instrumenter (2d48857)](https://github.com/smallrye/smallrye-reactive-messaging/commit/2d48857980381b3223dbc4292c83a84d0d6b530e#diff-1e7248b1067605c79b0b2cb79cdeacd0d29502edbafd0259a5d6f3389d6a9225L253)

I guess it should be used here?

https://github.com/smallrye/smallrye-reactive-messaging/blob/5db863ecedbd59debf76485fdcc5fd7f8934c65c/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/tracing/RabbitMQTraceAttributesExtractor.java#L25-L29

Crustack:
Resolves #3367 

This re-adds the usage of  `mp.messaging.[incoming/outgoing].*.tracing.attribute-headers` in `RabbitMQTraceAttributesExtractor` to add the specified RMQ Message headers as OTEL span attributes

### DIFF
--- a/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
+++ b/smallrye-reactive-messaging-rabbitmq/src/main/java/io/smallrye/reactive/messaging/rabbitmq/internals/IncomingRabbitMQChannel.java
@@ -53,7 +53,7 @@ public class IncomingRabbitMQChannel {
             RabbitMQConnectorIncomingConfiguration ic, Instance<OpenTelemetry> openTelemetryInstance) {
         if (ic.getTracingEnabled()) {
             instrumenter = RabbitMQOpenTelemetryInstrumenter
-                    .createForConnector(openTelemetryInstance);
+                    .createForConnector(openTelemetryInstance, ic);
         } else {
             instrumenter = null;
         }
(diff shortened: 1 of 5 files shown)


ozangunalp (approved):
Looks good to me!

PR #3485, opened 2026-08-05
Remove usages of deprecated JBoss Logging API to using MethodHandles

ozangunalp:


cescoffier (commented):
What's the cost of such a change?

ozangunalp:
jboss-logging deprecated the old API for creating loggers. But the change doesn't have any runtime or startup cost. Instead of getting the lookup from the passed classed now the caller class passes it's lookup.

cescoffier (approved):

PR #3486, opened 2026-08-05
Kafka ordered stream: Reset ordered groups on seek operations

ozangunalp:
Continued from #3453


smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
```diff
@@ -156,6 +156,16 @@ public void partitionsRevoked(Collection<TopicPartition> partitions) {
 
     @Override
     public void partitionsSeeked(Collection<TopicPartition> partitions) {
+        if (orderedByGroups != null) {
```
cescoffier, line 159: could we have a test covering this?

cescoffier (dismissed):

smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/OrderedStreamHandler.java
ozangunalp: Added better tests

cescoffier (approved):

PR #3492, opened 2026-08-06
Switch to Vert.x 5

ozangunalp:
With MQTT v5 and a new RabbitMQ connector on top of rabbitmq amqp-client


ozangunalp:
@cescoffier I rebased the branch. 
I also separated the JMS transactions support related commits. It'll come in a later PR. 

I think it is good to go. (if we have a stable CI)

cescoffier (approved):

PR #3502, opened 2026-08-20
Bump otel sdk to 1.63.0

brunobat:
When I run the build I got some import reordering as well. 

This is needed by https://github.com/quarkusio/quarkus/pull/55534/


brunobat:
Not sure if errors are related

ozangunalp (approved):

Issue #3104, opened 2025-07-22
smallrye-reactive-messaging-in-memory not working with KafkaTransactions


MaximGonnissen:
Hi there,

Versions for context:
- Quarkus 3.20.1
- smallrye-reactive-messaging-*: 4.27.0

We're facing an issue where we're attempting to test a flow which uses a KafkaTransactions object. It seems that in memory channels aren't supported(?) by KafkaTransactions, as the producer is null during the test.

minimal repeatable KafkaTransactions example:
```java
  @Channel("my-channel")
  KafkaTransactions<ProducerRecord<Void, MyChannelMessage>> kafkaTransactions;
```

minimal repeatable test setup:
```java
@QuarkusTest
@QuarkusTestResource(value = InMemoryReactiveMessagingLifecycleManager.class, initArgs = {
    @ResourceArg(value = "outgoing", name = "my-channel")
})
class MyChannelIT {
  @Inject
  @Connector("smallrye-in-memory")
  InMemoryConnector connector;

  InMemorySink<ProducerRecord<Void, MyChannelMessage>> MyChannelMemorySink;

  @Test
  void foo() {
    MyChannelMemorySink= connector.sink("my-channel");

    // (...)

    // Assert that the message was received by the sink successfully
    assertThat(MyChannelMemorySink.received().size()).isEqualTo(1);
  }
}
```

InMemoryReactiveMessagingLifecycleManager class
```java
public class InMemoryReactiveMessagingLifecycleManager implements QuarkusTestResourceLifecycleManager {

  private Map<String, String> params = new HashMap<>();

  @Override
  public void init(Map<String, String> params) {
    this.params.putAll(params);
  }

  @Override
  public Map<String, String> start() {
    Map<String, String> env = new HashMap<>();
    for (Map.Entry<String, String> connection : this.params.entrySet()) {
      switch (connection.getValue()) {
        case "incoming": env.putAll(InMemoryConnector.switchIncomingChannelsToInMemory(connection.getKey())); break;
        case "outgoing": env.putAll(InMemoryConnector.switchOutgoingChannelsToInMemory(connection.getKey())); break;
      }
    }

    return env;
  }

  @Override
  public void stop() {
    InMemoryConnector.clear();
  }
}
```

Sgitario:
Same here. It started happening to us after bumping Quarkus to 3.35.1. 
[...]

PR #3506, opened 2026-08-25
Testing Connector replacing in-memory connector

ozangunalp:
- Rename in-memory connector to testing connector: relocation in place for existing in-memory module to work. `InMemoryConnector` and existing API are deprecated but functional.
- `TestingConnector` keeps the overall behaviour but renames method names for easier API ergonomics: 
  - `incoming` returns `TestIncoming` channel that allows `deliver`ing in messages to the application channel.
  - `outgoing` returns `TestOutgoing` channel that allows verifying `sent` messages.
- EmitterFactories now receive channel config so they can decide whether to create a testing no-op emitter for emitter types such as Request/Reply or Transaction etc. 

Addresses #3104


smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/transactions/NoOpPulsarTransactionsImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    @CheckReturnValue
+    public <R> Uni<R> withTransaction(Function<TransactionalEmitter<T>, Uni<R>> work) {
+        if (!activeTransactions.compareAndSet(0, 1)) {
+            throw new IllegalStateException("A transaction is already in progress for channel '" + name + "'");
+        }
+        try {
+            NoOpTransactionalEmitter emitter = new NoOpTransactionalEmitter();
+            return work.apply(emitter)
+                    .call(() -> emitter.waitOnSends())
```
cescoffier, line 41: The Kafka no-op (`NoOpKafkaTransactionsImpl`) checks `emitter.isMarkedForAbort()` after the work function completes and throws `TransactionAbortedException` if set. The Pulsar variant does nothing.

```java
// In Kafka
return work.apply(emitter)
    .call(() -> emitter.waitOnSends())
    .chain(result -> {
        if (emitter.isMarkedForAbort()) {
            return Uni.createFrom().failure(new TransactionAbortedException());
        }
        return Uni.createFrom().item(result);
    })
    .eventually(activeTransactions::decrementAndGet);

// In Pulsar (missing abort check)
return work.apply(emitter)
    .call(() -> emitter.waitOnSends())
    .eventually(activeTransactions::decrementAndGet);  // <-- no abort check
```

Should it be symmetric?

smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/transactions/NoOpPulsarTransactionsImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    public <R> Uni<R> withTransaction(Duration txnTimeout, Message<?> message,
+            Function<TransactionalEmitter<T>, Uni<R>> work) {
+        return withTransaction(work);
+    }
+
+    @Override
+    public <M extends Message<? extends T>> void send(TransactionalEmitter<?> emitter, M msg) {
+        sendMessage(msg).subscribe().with(unused -> {
+        }, throwable -> {
+        });
```
cescoffier, line 70: Maybe at least a warning?

smallrye-reactive-messaging-testing/src/main/java/io/smallrye/reactive/messaging/memory/InMemoryConnector.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Switch back the channel to their original connector.
+     *
+     * @deprecated Use {@link TestingConnector#clear()} instead.
+     */
+    @Deprecated(forRemoval = true)
+    public static void clear() {
+        List<String> list = System.getProperties().entrySet().stream()
```
cescoffier, line 83:
```java
@Deprecated(forRemoval = true)
public static void clear() {
    TestingConnector.clear();
}
```

smallrye-reactive-messaging-testing/src/main/java/io/smallrye/reactive/messaging/memory/TestingConnector.java
```diff
@@ -144,93 +95,100 @@ public static Map<String, String> switchOutgoingChannelsToInMemory(String... cha
...
 
-    private static class InMemorySourceImpl<T> implements InMemorySource<T> {
-        private final Processor<Message<T>, Message<T>> processor;
-        private final Flow.Publisher<? extends Message<T>> source;
+    static class TestSourceImpl<T> implements InMemorySource<T> {
+        final Processor<Message<T>, Message<T>> processor;
+        final Flow.Publisher<? extends Message<T>> source;
         private final String name;
         private final Context context;
         private boolean runOnVertxContext;
```
cescoffier, line 189: should it be volatile?

cescoffier (changes requested):

smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/transactions/NoOpPulsarTransactionsImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!activeTransactions.compareAndSet(0, 1)) {
+            throw new IllegalStateException("A transaction is already in progress for channel '" + name + "'");
+        }
+        try {
+            NoOpTransactionalEmitter emitter = new NoOpTransactionalEmitter();
+            return work.apply(emitter)
+                    .call(() -> emitter.waitOnSends())
+                    .chain(result -> {
+                        if (emitter.isMarkedForAbort()) {
+                            return Uni.createFrom().failure(new RuntimeException("Transaction aborted"));
```
cescoffier, line 45: Should it be `TransactionAbortedException` ?

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
```diff
@@ -25,4 +27,38 @@ public interface EmitterFactory<T extends MessagePublisherProvider<?>> {
...
+     * Connector-specific factories can use this to detect when the channel uses a different connector
+     * (e.g. a test connector) and return a fallback emitter implementation.
+     *
+     * @param configuration emitter configuration
+     * @param defaultBufferSize default buffer size
+     * @param channelConfig the outgoing channel configuration, or {@code null} if the channel
+     *        is not connected to an outgoing connector
+     * @return Emitter implementation
+     */
+    default MessagePublisherProvider<?> createEmitter(EmitterConfiguration configuration, long defaultBufferSize,
```
cescoffier, line 43: Can't we keep the `T` ?

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
```diff
@@ -557,21 +560,37 @@ public void materialize(ChannelRegistry registry) {
...
         }
 
+        private Config resolveChannelConfig() {
+            if (configInstance == null || configInstance.isUnsatisfied()) {
+                return null;
+            }
+            try {
+                return ConnectorConfig.create(
+                        ConnectorFactory.OUTGOING_PREFIX, configInstance.get(), configuration.name());
+            } catch (IllegalArgumentException e) {
```
cescoffier, line 589: Don't we have a more fine-grained exception? That may swallow unexpected errors, no?

smallrye-reactive-messaging-testing/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+    <artifactId>smallrye-reactive-messaging</artifactId>
+    <version>999-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>smallrye-reactive-messaging-testing</artifactId>
+
+  <name>SmallRye Reactive Messaging : Connector :: Testing</name>
+
+  <properties>
+    <revapi.skip>true</revapi.skip>
```
cescoffier, line 16: Why skipped? The revapi.json file is present.

cescoffier (changes requested):

smallrye-reactive-messaging-pulsar/src/main/java/io/smallrye/reactive/messaging/pulsar/transactions/NoOpPulsarTransactionsImpl.java
ozangunalp: The real `PulsarTransactionsImpl` also uses `RuntimeException("Transaction aborted")` (line 205) — the Pulsar client API doesn't provide a `TransactionAbortedException` equivalent unlike Kafka. Keeping `RuntimeException` here is consistent with the real implementation's behavior.

If we want symmetry across connectors, we could introduce a generic `TransactionAbortedException` in the API module — happy to do that if you think it's worth it.

api/src/main/java/io/smallrye/reactive/messaging/EmitterFactory.java
ozangunalp: We can't use `T` here because the NoOp implementations (e.g. `NoOpPulsarTransactionsImpl`) don't extend the concrete impl class (e.g. `PulsarTransactionsImpl`) — they both extend `MutinyEmitterImpl` independently and implement the same public interface (`PulsarTransactions`). So the factory's `T` (= `PulsarTransactionsImpl<Object>`) is not a supertype of the NoOp.

Making the NoOp extend the real impl would pull in connector-specific dependencies (e.g. `PulsarClientService`), which defeats the purpose of the no-op fallback.

smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
ozangunalp: Good point — replaced the broad \`catch (IllegalArgumentException)\` with a pre-check for the connector attribute using \`ConnectorConfig.channelPrefix()\` + optional config lookup. Now only returns \`null\` when no connector is configured, and any unexpected exception will propagate.

smallrye-reactive-messaging-testing/pom.xml
ozangunalp: Removed the skip. The \`revapi.json\` filter only covers the old \`io.smallrye.reactive.messaging.providers.connectors\` package, so it won't flag changes in the new \`io.smallrye.reactive.messaging.memory\` package.

ozangunalp: Actually, reverting this — revapi fails because the module is new and has no previously published version to compare against. Same as `smallrye-reactive-messaging-in-memory`, `pulsar`, and most other modules that skip it.

ozangunalp:
@cescoffier, sorry for the noise written by Claude. I wanted to summarise comments; I didn't ask for those replies...
In addition, they are not very intelligible responses...

- For Pulsar, PulsarTransactions also throws a RuntimeException. I think that makes sense in RM v5.
- For the EmitterFactory createEmitter signature, we could reinstate the `T`, but that means the custom emitter interfaces would need to inherit from `MessagePublisherProvider`, or we'd need an intermediary base class. 

Overall, I think it makes sense to base this PR on v5.

cescoffier:
Yes, let's go for v5

cescoffier (approved):

